PR #66, opened 2026-05-18
fix(dashboard): improve responsive design and dark mode contrast

ashiwanikumar:
## Summary

- **Webhooks page**: Redesigned from cramped table layout to responsive card-based layout. Each webhook displays URL, session, status, and events clearly with proper overflow handling on all screen sizes.
- **Webhooks dark mode**: Added contrast improvements for event tags, status badges, labels, and URL text using `[data-theme=dark]` and `@media (prefers-color-scheme: dark)` selectors.
- **Webhooks status badge**: Changed to `inline-flex` with `width: fit-content` to prevent full-width stretching on mobile.
- **Message Tester**: Toggle groups (recipient type, message type) now wrap into individual pill buttons on mobile instead of overflowing off-screen.
- **Plugins**: Show all engine features instead of truncating at 8. Feature names formatted with spaces for readability.

## Test Plan

- [ ] Open `/webhooks` on desktop — cards display with URL, session, status, events
- [ ] Open `/webhooks` on mobile (< 768px) — cards stack, URL truncates with ellipsis, event tags wrap
- [ ] Toggle dark mode — event tags, labels, status badges have good contrast
- [ ] Open `/message-tester` on mobile — toggle buttons wrap into pills
- [ ] Open `/plugins` — all engine features visible with readable names
- [ ] Build passes: `npm run build` in dashboard directory

### DIFF
--- a/dashboard/src/index.css
+++ b/dashboard/src/index.css
@@ -359,3 +359,56 @@ code {
   left: 1rem;
 }
 
+/* Mobile Responsive - Global */
+@media (max-width: 768px) {
+  h1 {
+    font-size: 1.5rem;
+  }
+
+  .modal {
+    width: 95%;
+    max-width: none;
+    margin: 1rem;
+    border-radius: 12px;
+  }
+
+  .modal-header {
+    padding: 1.25rem 1.25rem 0.875rem;
+  }
+
+  .modal-header h2 {
+    font-size: 1.125rem;
+  }
+
+  .modal-body {
+    padding: 1.25rem;
+  }
+
+  .modal-footer {
+    padding: 0.875rem 1.25rem 1.25rem;
+  }
+
+  select {
+    padding: 0.625rem 2.25rem 0.625rem 0.875rem;
+    font-size: 0.875rem;
+  }
+}
+
+@media (max-width: 480px) {
+  h1 {
+    font-size: 1.25rem;
+  }
+
+  .modal-overlay {
+    align-items: flex-end;
+  }
+
+  .modal {
+    width: 100%;
+    margin: 0;
+    border-radius: 16px 16px 0 0;
+    max-height: 90vh;
+    overflow-y: auto;
+  }
+}
+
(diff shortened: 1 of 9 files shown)


rmyndharis (commented):
Thank you for this — the dark-mode contrast work and the mobile card layouts are a genuine, well-executed improvement, and the refactor is careful (handlers, state references, and event keys are preserved in the Webhooks table-to-card change).

A scoping note and a few items before this lands:

**Scope: this is a v0.2.0 enhancement, not a v0.1.8 patch.**
The change goes beyond a bugfix — it adds new responsive breakpoints across several pages, redesigns the Webhooks page from a table to cards, and changes Plugins display behavior (below). You already assigned the **v0.2.0** milestone, which is exactly right. We're keeping the upcoming **v0.1.8** strictly to bugfixes, so this will ride with v0.2.0 rather than the patch.

**Please rebase onto current `main`.**
This branch predates the 0.1.7 release, so CI is currently red against a stale base — it's mergeable (no conflicts), just behind. A rebase picks up the lint check-mode and `npm audit` fixes from 0.1.7 and should turn CI green.

**One product decision — the `Plugins.tsx` feature cap.**
Dropping the `.slice(0, 8)` cap and the "+N more" indicator is a deliberate UX change: engines like the whatsapp-web.js adapter expose ~15 features, which will now all render. `.features-list` uses `flex-wrap`, so it won't overflow horizontally, but the engine card can get tall. Could you either keep the cap behind a "show more" toggle, or confirm you intend the full list to render?

**Minor:**
- `Webhooks.css` `.events-reference h3` re-declares the global H3 typography (`font-size`/`font-weight`/`color`/`text-transform`/`letter-spacing`); the rest of the dashboard uses the `/* Inherits from H3 */` pattern — worth matching for consistency.
- The `sessionName` line is 128 chars (over the 120 limit); `npm run format` will reflow it.
- Quick sanity-check on the new `.sessions-table .table-header` rule in the mobile card block — it appears to have an empty body. Please confirm the header is intentionally hidden/handled in card mode.

Thanks again — this is good work; let's land it in v0.2.0 once rebased.

rmyndharis:

PR #67, opened 2026-05-18
feat(security): harden NestJS API for production deployment

ijaznaeem:
Add body size limits, production CORS policy, trust proxy, and Swagger gating. Wire API_MASTER_KEY for bootstrap auth, enforce webhook HMAC secrets on create/dispatch, and register SecurityMiddleware globally. Update .env.example with new security variables.

## Description
Brief description of changes

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] Lint passes
- [ ] Self-reviewed

## Related Issues
Closes #

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -19,8 +19,14 @@ DASHBOARD_PORT=2886
 # BASE_URL=https://api.yourdomain.com
 # DASHBOARD_URL=https://dashboard.yourdomain.com
 
-# CORS Configuration (comma-separated origins, or * for all)
-CORS_ORIGINS=*
+# CORS Configuration (comma-separated origins; avoid * in production)
+CORS_ORIGINS=https://dashboard.yourdomain.com
+
+# Trust reverse proxy (Nginx/Traefik) for X-Forwarded-* headers
+TRUST_PROXY=true
+
+# Max JSON body size (Express limit)
+BODY_SIZE_LIMIT=5mb
 
 # SSL/TLS (enable for production with real domain)
 # TRAEFIK_ACME_EMAIL=admin@yourdomain.com
@@ -94,6 +100,8 @@ S3_SECRET_KEY=minioadmin
 WEBHOOK_TIMEOUT=10000               # Timeout in milliseconds
 WEBHOOK_MAX_RETRIES=3               # Number of retry attempts
 WEBHOOK_RETRY_DELAY=5000            # Delay between retries in ms
+WEBHOOK_SECRET=                     # Default HMAC secret for new webhooks (min 32 chars)
+WEBHOOK_REQUIRE_SECRET=true         # Require secret on every webhook in production
 
 # =============================================================================
 # RATE LIMITING
@@ -111,7 +119,7 @@ PLUGINS_DIR=./data/plugins          # Plugin directory
 # SECURITY
 # =============================================================================
 # Master API key (leave empty to disable, or set to secure value)
-API_MASTER_KEY=
+API_MASTER_KEY=                     # Master key for REST API (min 32 chars recommended)
 
 # =============================================================================
 # DEVELOPER SETTINGS
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -1,4 +1,4 @@
-import { Module, DynamicModule, Type } from '@nestjs/common';
+import { Module, DynamicModule, Type, NestModule, MiddlewareConsumer } from '@nestjs/common';
 import { ConfigModule, ConfigService } from '@nestjs/config';
 import { TypeOrmModule } from '@nestjs/typeorm';
 import { ThrottlerModule } from '@nestjs/throttler';
@@ -26,6 +26,7 @@ import { CatalogModule } from './modules/catalog/catalog.module';
 import { HooksModule } from './core/hooks';
 import { PluginsModule } from './core/plugins';
 import { PluginsApiModule } from './modules/plugins/plugins.module';
+import { SecurityMiddleware } from './common/security/security.middleware';
 
 // Only import QueueModule if explicitly enabled to avoid Redis connection errors
 const queueModules: Array<Type | DynamicModule> = [];
@@ -161,4 +162,8 @@ if (process.env.QUEUE_ENABLED === 'true') {
     PluginsApiModule, // Phase 5: Plugins API
   ],
 })
-export class AppModule {}
+export class AppModule implements NestModule {
+  configure(consumer: MiddlewareConsumer): void {
+    consumer.apply(SecurityMiddleware).forRoutes('*');
+  }
+}
--- a/src/common/security/secure-compare.util.ts
+++ b/src/common/security/secure-compare.util.ts
@@ -1,0 +1,19 @@
+import * as crypto from 'crypto';
+
+/**
+ * Constant-time string comparison to prevent timing attacks on secrets.
+ */
+export function secureCompare(a: string, b: string): boolean {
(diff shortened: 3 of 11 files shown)


rmyndharis:
Thanks @ijaznaeem — solid production-hardening direction. A fair amount has since landed or conflicts, so this needs a rebase + split:

- **Already on main:** `API_MASTER_KEY` seed (#133), generic `Error` → HTTP exceptions (#102), global Swagger X-API-Key scheme (#109).
- **Worth keeping** (please extract into a focused PR): `ENABLE_SWAGGER` gating (activates the documented-but-dead flag), body-size limits, `@SkipThrottle()` on `/health`, and wiring the existing `SecurityMiddleware` (de-duped against helmet).
- **Please drop / rework separately:** the webhook HMAC-required enforcement (silently skips delivery for existing null-secret webhooks and makes `secret` mandatory — breaking), the `validateApiKey` master-key bypass (bypasses IP/session/expiry and conflicts with #133's revocable model), and `trust proxy` default-on (re-introduces X-Forwarded-For spoofing — see #211's `TRUSTED_PROXIES` approach).

The PR is currently conflicting against `main`. Moving to v0.2.0. Thanks!

rmyndharis (changes requested):
## Code review

Thanks for the substantial work on production hardening. The direction is right, but several of the changes introduce regressions that partially work against the stated hardening goal. Recommending the following items be addressed before merge.

### Blocking

#### 1. `TRUST_PROXY` defaults to `true` when the env var is unset

In `src/config/configuration.ts` and `src/main.ts`, `trustProxy` is computed as `process.env.TRUST_PROXY !== 'false'`, which evaluates to `true` whenever the variable is absent. Trusting proxy headers (`X-Forwarded-For`, `X-Forwarded-Proto`) when there is no proxy in front allows any client to spoof their source IP, defeating IP allowlists and any IP-based rate limiting.

For a hardening PR, the safer default is opt-in (`process.env.TRUST_PROXY === 'true'`), or at minimum require `true` only when `NODE_ENV=production` AND the operator has explicitly set the variable.

https://github.com/rmyndharis/OpenWA/blob/0af1eff181897969bdacc5667806439a06deabcb/src/config/configuration.ts#L67-L74

https://github.com/rmyndharis/OpenWA/blob/0af1eff181897969bdacc5667806439a06deabcb/src/main.ts#L74-L79

#### 2. Swagger is enabled by default in production

`enableSwagger` is `process.env.ENABLE_SWAGGER !== 'false'`, meaning Swagger UI is exposed in production unless the operator explicitly opts out. This exposes the full API surface, auth header names, and route schema. For a hardening PR the safer default is opt-in (`=== 'true'`), particularly when `NODE_ENV=production`.

https://github.com/rmyndharis/OpenWA/blob/0af1eff181897969bdacc5667806439a06deabcb/src/config/configuration.ts#L67-L74

#### 3. `resolveWebhookSecret` throws unconditionally and makes `WEBHOOK_REQUIRE_SECRET` dead code

`resolveWebhookSecret` evaluates `requireSecret`, but the second `if (!secret)` block throws regardless of the flag — so the `requireSecret=false` branch is unreachable. As a result, any `POST /api/webhooks` or `PATCH /api/webhooks/:id` request that omits `secret` (with no `WEBHOOK_SECRET` default configured) is rejected with a 400 regardless of environment. Existing webhook clients that relied on the original opt-in behaviour will break.

```ts
if (!secret && requireSecret) {
  throw new BadRequestException(...);
}

if (!secret) {
  throw new BadRequestException('Webhook secret is required for signed deliveries.');
}
```

Suggested fix: drop the second throw, or gate it behind the same `requireSecret` check.

https://github.com/rmyndharis/OpenWA/blob/0af1eff181897969bdacc5667806439a06deabcb/src/modules/webhook/webhook.service.ts#L367-L386

### Recommended

#### 4. Retry path throws a raw `Error`, bypassing the global HTTP exception filter

`deliverWebhook` (the fallback retry path) calls `getEffectiveSecret(webhook)` without a `try/catch`. If a legacy webhook has no secret stored and `WEBHOOK_SECRET` is unset, `getEffectiveSecret` throws `new Error('Webhook secret is not configured')` — a plain JS error rather than a NestJS `HttpException`. The dispatch loop (`dispatchToWebhooks`) already guards this call; the retry path needs the same treatment so the global exception filter can format the response consistently rather than producing an unstructured 500.

https://github.com/rmyndharis/OpenWA/blob/0af1eff181897969bdacc5667806439a06deabcb/src/modules/webhook/webhook.service.ts#L321-L328

#### 5. `secureCompare` leaks length via early exit

`secureCompare` returns `false` early when the two buffers differ in length, which lets an attacker distinguish "wrong length" from "right length, wrong bytes" through timing. For HMAC verification (fixed-length digests) this is harmless, but if `secureCompare` is ever used for variable-length comparisons (master key, raw API key) the length leak is meaningful.

A common timing-safe alternative is to HMAC both inputs to a fixed digest before comparing.

https://github.com/rmyndharis/OpenWA/blob/0af1eff181897969bdacc5667806439a06deabcb/src/common/security/secure-compare.util.ts#L6-L19

#### 6. `console.warn` instead of the project logger in `main.ts`

The new CORS warning uses `console.warn`. The rest of the codebase uses `LoggerService` for production logging. Bootstrap context makes structured logging slightly awkward, but the warning is security-relevant and should land in the same log stream operators monitor.

https://github.com/rmyndharis/OpenWA/blob/0af1eff181897969bdacc5667806439a06deabcb/src/main.ts#L121-L125

#### 7. Master key usage tracking is silently broken

`validateApiKey` short-circuits with `buildMasterKeyPrincipal()`, which returns a synthetic in-memory `ApiKey` with `usageCount: 0` and never persists `lastUsedAt`. Any dashboard or audit tooling reading these fields for the master key will show it as "never used" even when it is actively serving requests. If this is intentional (e.g. to keep the master key off the audit trail), please document it.

https://github.com/rmyndharis/OpenWA/blob/0af1eff181897969bdacc5667806439a06deabcb/src/modules/auth/auth.service.ts#L165-L171

https://github.com/rmyndharis/OpenWA/blob/0af1eff181897969bdacc5667806439a06deabcb/src/modules/auth/auth.service.ts#L218-L235

#### 8. New logic added to a `@deprecated` method

`deliverWebhook` carries the JSDoc `@deprecated Use job queue dispatch instead. This is kept for fallback.` The PR layers new secret-resolution logic into it instead of moving the new code into the queue processor. Either lift the deprecation, or move the secret resolution to the active dispatch path.

https://github.com/rmyndharis/OpenWA/blob/0af1eff181897969bdacc5667806439a06deabcb/src/modules/webhook/webhook.service.ts#L309-L328

rmyndharis:
Quick follow-up to the review above: this branch now conflicts with `main`. A fair amount of related hardening landed in v0.1.7 since you opened this — e.g. the X-Forwarded-For / trust-proxy handling (#211), path-traversal containment + ADMIN-only infra endpoints (#207), and the global Swagger `X-API-Key` scheme (#109).

Could you rebase onto current `main` and then address the three **Blocking** items? Some of the overlap may resolve or simplify after the rebase, so it's worth doing that first. Once it's rebased and CI is green I'll re-review promptly. Thanks again for the substantial hardening work — the direction is right, this is just about landing it safely in focused pieces.

rmyndharis:

PR #94, opened 2026-05-19
Fix Swagger API key docs and postinstall

jaydeeprusia:
## Description
Align Swagger auth docs with API key authentication, make postinstall cross-platform, and guard the sqlite migration.

## Type of Change
- [x] Bug fix
- [x] Documentation update

## Testing
Verified locally in WSL; Swagger auth docs work as expected.

## Checklist
- [x] Documentation updated
- [x] Self-reviewed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "openwa",
-  "version": "0.1.5",
+  "version": "0.1.6",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "openwa",
-      "version": "0.1.5",
+      "version": "0.1.6",
       "hasInstallScript": true,
       "license": "MIT",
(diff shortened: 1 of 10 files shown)


rmyndharis:
Thanks @jaydeeprusia! The cross-platform postinstall idea here is great and I have landed it as a focused one-line change in #209 (credited to you). Closing this PR because it also bundles three unrelated changes that we cannot take as-is: (1) the Swagger @ApiSecurity/@ApiBearerAuth changes are already superseded by the global X-API-Key scheme work in #109 (commit 49328bc), which also adds a test; (2) the SQLite migration guard only protects the `sessions` CREATE TABLE while upSqlite creates several tables in sequence, so it risks a partial schema; and (3) the regenerated package-lock.json is noise. The postinstall fix lives on in #209 — appreciate the contribution!

PR #105, opened 2026-05-19
fix(security): harden default API key, IP whitelist, and QR access

Pakzone1:
## Description

This PR tightens a few security gaps around API keys and session pairing.

On first startup, OpenWA used to create a fixed `dev-admin-key` whenever `NODE_ENV` was not set to production. Many Docker and staging setups never set that variable, so new installs could end up with a known admin key. Now the default is always a random key. The old predictable key is only used when you set `ALLOW_DEV_API_KEY=true`, which is meant for local development only. The dev Docker compose file turns that on so local workflows stay convenient.

API keys with an IP whitelist used to skip the check when the server could not see a client IP. That meant the restriction could be bypassed. If a whitelist is configured and the client IP is unknown, the request is now rejected.

The QR code endpoint for linking WhatsApp (`GET /sessions/:id/qr`) did not require operator permissions. A viewer key could fetch the QR and pair another device to someone else's session. That route now requires the OPERATOR role.

Docs were updated in `.env.example` and `docs/README.md` to explain `ALLOW_DEV_API_KEY`.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

After `npm install`, run:

- `npm test -- src/modules/auth/auth.service.spec.ts`
- `npm run lint`

Manual checks (optional):

- Fresh install without `ALLOW_DEV_API_KEY`: startup log and `data/.api-key` should show a random `owa_k1_...` key.
- With `ALLOW_DEV_API_KEY=true`: first boot uses `dev-admin-key` (dev only).
- A VIEWER key calling `GET /sessions/:id/qr` should get 401.

## Checklist
- [x] Tests added/updated
- [x] Documentation updated
- [ ] Lint passes (run locally before merge)
- [x] Self-reviewed

N/A

## Related Issues

No linked issue. This addresses insecure default API key behavior and related auth hardening.

### DIFF
--- a/src/modules/auth/auth.service.spec.ts
+++ b/src/modules/auth/auth.service.spec.ts
@@ -234,6 +234,16 @@ describe('AuthService', () => {
       expect(result.id).toBe(key.id);
     });
 
+    it('should reject when IP whitelist is set but client IP is unknown', async () => {
+      const key = createMockApiKey({
+        allowedIps: ['10.0.0.1'],
+        keyHash: hashKey('ip-no-client'),
+      });
+      (repository.findOne as jest.Mock).mockResolvedValue(key);
+
+      await expect(service.validateApiKey('ip-no-client')).rejects.toThrow('Client IP could not be determined');
+    });
+
     it('should throw UnauthorizedException when session not in allowedSessions', async () => {
       const key = createMockApiKey({
         allowedSessions: ['session-A'],
--- a/.env.example
+++ b/.env.example
@@ -113,6 +113,9 @@ PLUGINS_DIR=./data/plugins          # Plugin directory
 # Master API key (leave empty to disable, or set to secure value)
 API_MASTER_KEY=
 
+# First-boot default admin key: always random unless this is set (local dev only)
+# ALLOW_DEV_API_KEY=true
+
 # =============================================================================
 # DEVELOPER SETTINGS
 # =============================================================================
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -12,6 +12,7 @@ services:
       - '127.0.0.1:2785:2785'
     environment:
       - NODE_ENV=development
+      - ALLOW_DEV_API_KEY=true
       - PORT=2785
       - DATABASE_TYPE=sqlite
       - DATABASE_NAME=/app/data/openwa.sqlite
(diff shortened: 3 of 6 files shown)


src/modules/auth/auth.service.spec.ts
```diff
@@ -234,6 +234,16 @@ describe('AuthService', () => {
       expect(result.id).toBe(key.id);
     });
 
+    it('should reject when IP whitelist is set but client IP is unknown', async () => {
+      const key = createMockApiKey({
+        allowedIps: ['10.0.0.1'],
```
guerralindao, line 239: It would be great if we could configure it via the .env file (allowedIps). Have you already thought about this, or is the configuration done via the web interface ?

Pakzone1: thanks for the comment!
the `allowedIps: ['10.0.0.1']` in the test is just mock data for the unit test, it's not pulled from `.env` or anything like that.
in openwa right now.. IP whitelisting is configured per API key, not globally in the env file. you set it when creating or updating a key via the API (`POST/PATCH /api/auth/api-keys`) or through the dashboard, and those values live in the database on the `api_keys` table. so..this PR doesn't add any new config. it just fixes how existing per-key whitelists are enforced. specifically, if a key has `allowedIps` set and the server can't figure out the client IP, the request gets rejected now instead of silently skipping the check.
a global IP allowlist via `.env` (something like `API_ALLOWED_IPS`) could definitely be handy, but that's a bit out of scope for this security fix. happy to open a follow-up issue or PR for that if the maintainers are interested.

rmyndharis (changes requested):
Thanks @Pakzone1 — great catches. I've extracted the two clean, non-breaking fixes into #213 (credited to you): the **fail-closed IP whitelist** (+ regression test) and the **QR `OPERATOR` role gate**.

The remaining change — making the first-boot default key random unless `ALLOW_DEV_API_KEY=true` — is valuable but (a) behavior-breaking for existing dev/Docker setups that rely on `dev-admin-key`, and (b) conflicts with the `API_MASTER_KEY` precedence already merged in #133. Could you rebase that part on current `main`, preserve `API_MASTER_KEY || (ALLOW_DEV_API_KEY ? 'dev-admin-key' : random)`, reconcile the README wording (changed by #130), and flag it as a breaking change with an upgrade note? Moving to the v0.2.0 milestone for that part. Thanks!

rmyndharis:

PR #109, opened 2026-05-20
fix(swagger): apply X-API-Key security scheme globally

### LINKED ISSUE #104: Error: Unauthorized
Send and manage messages
/api/sessions/{sessionId}/messages/send-bulk

When I try to send a mass message to two numbers, I get the "Unauthorized" error, even though I've already given the necessary authorization.

From the panel I can send individual test messages normally

Image

Image

Image

Image

Image

rmyndharis:
## Problem

Issue #104: `POST /sessions/:id/messages/send-bulk` (and every other protected endpoint) returns `401 Unauthorized` when called from Swagger UI, even after clicking Authorize and entering a valid API key.

## Root cause

`main.ts` defined the `X-API-Key` security scheme with `addApiKey()` but never applied it. In OpenAPI, defining a scheme is not the same as applying it: without a `security` requirement on operations, Swagger UI never attaches the credential. So Authorize was cosmetic, and every request reached the global `ApiKeyGuard` with no key, returning 401.

It looked like a `send-bulk`-only bug because the dashboard (which sends the key itself) has no bulk feature, so bulk was the first Swagger-only operation a user would reach. Every protected endpoint was affected.

A separate inconsistency: 5 controllers (`auth`, `plugins`, `catalog`, `status`, `stats`) used `@ApiBearerAuth()`, which references a `bearer` scheme that was never defined.

## Fix

- Extract the Swagger config into `src/config/swagger.config.ts` as `createSwaggerConfig()`, making it unit-testable.
- Apply the scheme globally with `addSecurityRequirements`, mirroring the global `ApiKeyGuard`.
- Remove the 5 stray `@ApiBearerAuth()` decorators. Operation-level security overrides the global requirement, so leaving them would keep those controllers broken.
- Add a regression test (`swagger.config.spec.ts`), written test-first.

## Verification

- Unit test: RED then GREEN.
- `npm run build`, `npm test` (111/111), `npm run lint`: all pass.
- Runtime: booted the backend and fetched `/api/docs-json`. The document now carries a global `security` `[{"X-API-Key":[]}]`, `send-bulk` inherits it, and 0 operation-level overrides across all 91 paths.

Fixes #104


onepay-ye:
jus add :
.addSecurityRequirements('X-API-Key')
in config main.ts

PR #112, opened 2026-05-20
fix: use node 20 for whatsapp-web.js stability

DISTinTheHouse:
## Description

Fixes WhatsApp session instability caused by Node.js 22 in the production Docker image.

The application successfully generated QR codes and authenticated sessions, but crashed immediately after login with:

```bash
Execution context was destroyed, most likely because of a navigation.
```

The issue was caused by incompatibility between:

* Puppeteer
* whatsapp-web.js
* Node.js 22

### Changes

Replaced:

```dockerfile
FROM node:22-slim AS production
```

With:

```dockerfile
FROM node:20-bookworm-slim AS production
```

## Type of Change

* [x] Bug fix

## Checklist

* [x] Self-reviewed
* [x] Tested locally
* [x] Sessions authenticate correctly
* [x] QR generation works
* [x] Message sending works
* [x] Connection remains stable after authentication

## Validation Performed

Tested successfully with:

* Docker Desktop
* Windows
* Session creation
* QR authentication
* Persistent WhatsApp connection
* Sending WhatsApp messages

## Notes

After this change:

* sessions no longer disconnect after authentication
* Puppeteer remains stable
* WhatsApp sessions stay connected properly

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ COPY . .
 RUN npm run build
 
 # ===== Stage 2: Production =====
-FROM node:22-slim AS production
+FROM node:20-bookworm-slim AS production
 
 # Install Chrome/Chromium and required dependencies
 RUN apt-get update && apt-get install -y \


rmyndharis:
Thanks @DISTinTheHouse! Closing this one: pinning Node 20 would now conflict with Vite 8 (requires Node ^20.19 || >=22.12) and archiver v8 (ESM, needs Node >=22.12), both project dependencies. The official image is node:22-slim. If you are seeing a specific whatsapp-web.js stability problem on Node 22, please open a focused issue with details and we will track it separately.

PR #121, opened 2026-05-21
fix(auth): prevent X-Forwarded-For IP spoofing in ApiKeyGuard

itsprason:
## Description
Fix IP spoofing vulnerability by ensuring X-Forwarded-For is only trusted
when the request originates from configured trusted proxy IPs. Otherwise,
the server uses the direct connection IP.

## Issue POC
image

## Testing
image

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -112,6 +112,17 @@ PLUGINS_DIR=./data/plugins          # Plugin directory
 # =============================================================================
 # Master API key (leave empty to disable, or set to secure value)
 API_MASTER_KEY=
+# Trusted reverse-proxy IPs / CIDR ranges (comma-separated).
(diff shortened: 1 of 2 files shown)


rmyndharis:
Thanks @itsprason — this is a real and well-spotted auth-bypass, and your trusted-proxy approach is the right model. I have landed a complete fix in #211 that builds on your direction and credits you. Closing this one because it was not mergeable as-is: it added a required ConfigService constructor dependency to ApiKeyGuard without updating api-key.guard.spec.ts (so the existing test suite no longer compiled/passed), and it missed IPv4-mapped IPv6 normalization (e.g. ::ffff:172.18.0.1), which would prevent Traefik/Docker dual-stack proxies from matching a v4 CIDR. #211 closes those gaps with full test coverage. Really appreciate the report and the fix direction!

PR #122, opened 2026-05-21
Fix Location constructor error

rafaelmerlotto:
## Description
Fixes the error "Location is not a constructor" occurring when using whatsapp-web.js.

The issue was caused by incorrect import and usage of the Location class. The code now correctly imports Location as LocationMessage and uses it accordingly.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

## Related Issues
No related issue


rmyndharis:
Thanks @rafaelmerlotto! Closing in favor of #186 (merged) — your safe-access version, which handles the ESM/CJS interop more defensively (module.Location || module.default?.Location). Appreciate both takes!

PR #129, opened 2026-05-22
fix(security): replace direct Docker socket with socket proxy and add…

A831ARD0:
… Bull Board auth

Three critical security fixes:

1. Docker socket proxy (tecnativa/docker-socket-proxy)
   - Add docker-proxy service that is the only container with access to /var/run/docker.sock
   - openwa-api connects via DOCKER_HOST=tcp://docker-proxy:2375 (TCP)
   - Proxy exposes only the required Docker API operations (CONTAINERS, IMAGES, VOLUMES, INFO, PING, POST, DELETE)
   - Eliminates the Docker daemon escape vector from openwa-api

2. Non-root container execution
   - Install gosu for safe privilege dropping
   - Add docker-entrypoint.sh: fixes named-volume ownership as root, then exec's to openwa user via gosu before starting Node.js
   - Process no longer runs as root inside the container

3. Bull Board dashboard authentication
   - Add BullBoardAuthMiddleware that validates X-API-Key / Bearer token before allowing access to /admin/queues
   - Requires ADMIN role; returns 401/403 otherwise
   - Applied globally via AppModule.configure()

Also fix dashboard nginx upstream name: openwa -> openwa-api to match the Docker Compose service name (fixes dashboard crash-loop on startup).

## Description
Brief description of changes

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] Lint passes
- [ ] Self-reviewed

## Related Issues
Closes #

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,6 +49,7 @@ RUN apt-get update && apt-get install -y \
     libxrandr2 \
     xdg-utils \
     dumb-init \
+    gosu \
     && rm -rf /var/lib/apt/lists/*
 
 # Set Chrome executable path for Puppeteer
@@ -73,9 +74,9 @@ COPY --from=builder /app/dist ./dist
 RUN mkdir -p ./data/sessions ./data/media && \
     chown -R openwa:openwa /app
 
-# Note: Running as root to allow Docker socket access for orchestration
-# For production with stricter security, consider using a Docker socket proxy
-# USER openwa
+# Copy entrypoint script (runs as root to fix volume permissions, then drops to openwa)
+COPY docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
+RUN chmod +x /usr/local/bin/docker-entrypoint.sh
 
 # Expose port
 EXPOSE 2785
@@ -84,6 +85,6 @@ EXPOSE 2785
 HEALTHCHECK --interval=30s --timeout=10s --start-period=30s --retries=3 \
     CMD node -e "require('http').get('http://localhost:2785/api/health', (r) => process.exit(r.statusCode === 200 ? 0 : 1))"
 
-# Start with dumb-init to handle signals properly
-ENTRYPOINT ["dumb-init", "--"]
+# Entrypoint: fixes volume permissions as root, then drops to openwa user via gosu
+ENTRYPOINT ["/usr/local/bin/docker-entrypoint.sh"]
 CMD ["node", "dist/main"]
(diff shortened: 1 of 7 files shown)


rmyndharis (changes requested):
Thanks @A831ARD0 — the `/admin/queues` auth gap is real and important; I've reworked the Bull Board auth middleware into #214 (verified at runtime), credited to you.

⚠️ Heads-up: the `configure()/forRoutes('/admin/queues')` wiring here doesn't actually protect the route at runtime — Bull Board's router runs first and serves the UI to everyone. #214 registers via `app.use()` **before** `app.listen()` instead.

The other two changes — **docker-socket-proxy** and **non-root entrypoint (gosu)** — are good defense-in-depth but carry deployment/runtime risk and have no tests, so they're best as their own focused PRs (the proxy becomes a required sidecar; the entrypoint needs an image smoke test, and note the dropped `--` after `dumb-init`). The nginx upstream rename is already handled for compose by #131. Could you split those into separate PRs rebased on current `main`? Moving to v0.2.0. Thanks!

A831ARD0:
Thanks @rmyndharis! Glad #214 landed with the app.use() fix — that's the correct approach.

I've split the remaining changes into two focused PRs rebased on current main as requested:

[#227](https://github.com/A831ARD0/OpenWA/issues/227) — docker-socket-proxy

Adds docker-proxy as the sole container with /var/run/docker.sock access; openwa-api connects via DOCKER_HOST=tcp://docker-proxy:2375
DockerService updated to read DOCKER_HOST instead of hardcoding the socket path (so no code changes are needed when the proxy is in place)
Includes unit tests for the connection options logic and a post-deploy smoke test script
[#228](https://github.com/A831ARD0/OpenWA/issues/228) — non-root entrypoint

Adds gosu + docker-entrypoint.sh to drop privileges before Node starts
Fixed the dropped --: ENTRYPOINT ["dumb-init", "--", "/usr/local/bin/docker-entrypoint.sh"] so dumb-init is guaranteed PID 1 and receives signals cleanly
Includes an image smoke test (docker run --rm <image> id) asserting uid != 0 and PID 1 is dumb-init
This PR can be closed — all three concerns are now covered by [#214](https://github.com/A831ARD0/OpenWA/issues/214), [#227](https://github.com/A831ARD0/OpenWA/issues/227), and [#228](https://github.com/A831ARD0/OpenWA/issues/228).

rmyndharis:
Superseded by your own cleaner, better-tested splits #227 and #228, both included in v0.2.0 (#229). Thank you! Closing in favor of those.

PR #177, opened 2026-06-02
updated vite configs for fixing ui not launching +fixing the logical issue that the /api/auth/validate endpoint is being blocked by the globally applied ApiKeyGuard. The validate endpoint needs to be public (no authentication required) since it's used to validate the API key itself - a chicken-and-egg problem.

MOREQASSE:


rmyndharis:
Closing: this mixes stray junk files (e.g. `openwa`, `response.json())`, netlify.toml) with one useful line (making /api/auth/validate public). If the public-validate fix is still needed, please open a clean, focused PR for just that change and we'll review it promptly.

Issue #168, opened 2026-06-01
Webhook not firing for self-messages (fromMe: true) on same session number


cavemanharris:
# Webhook not firing for self-messages (fromMe: true) on same session number

## Summary

When a WhatsApp user sends a message **to themselves** (i.e. opens their own chat and sends a message on the same number that OpenWA is running on), the webhook is **not fired**. The message does not appear in the webhook receiver at all — not even as a raw payload.

## Environment

- OpenWA version: (latest version)
- Node.js version: (latest version)
- OS: Linux (hosted)
- Webhook transport: HTTP POST

## Steps to Reproduce

1. Register a WhatsApp session in OpenWA on number A
2. Configure a webhook with `message.sent`, `message.received` and `*` events
3. From the same device running number A, open the self-chat and send a message
4. Observe that the webhook endpoint receives nothing

## Expected Behaviour

The webhook should fire with `fromMe: true` when the session owner sends a message to themselves, just as it fires for messages sent to other contacts.

## Actual Behaviour

No webhook event is fired. The endpoint receives nothing. This is confirmed by raw payload logging at the very top of the webhook receiver, before any processing logic — the file is simply not written to.

## What Does Work

- Webhook fires correctly for `message.received` from other numbers ✅
- Webhook fires correctly for `session.status` events ✅
- `message.sent` is registered in the webhook events array ✅
- The `*` wildcard event is also registered ✅

None of these result in a webhook firing for self-messages.

## Use Case

This is needed for owner self-service scenarios — for example, a consultant running an AI booking agent on their business WhatsApp number needs to be able to message their own number to query their schedule, check bookings, or interact with their own AI agent. Without this, the only workaround is to use a second phone number, which is impractical.

## Workaround Attempted

Registering `message.sent` and `*` as webhook events — neither results in the webhook firing for self-messages on the same session number.

## Request

Please either:
1. Fire the webhook for self-messages with `fromMe: true` so the receiving application can handle them, or
2. Provide an alternative API endpoint or event that captures self-messages so they can be processed by the webhook consumer

Otherwise, thank you. This is a magnificent product.

nalmeida84:
should be nice to have also for manual reply in a automated chat... I mean, if I send a message from my phone I'd like to have it in my app conversation... this message isn't being captured.

rmyndharis:
[...]

Issue #146, opened 2026-05-26
字段缺失


ghost:
message.received 的 webhook payload 缺少 contact 对象和 author 字段

rmyndharis:
Fixed in #223 — the `message.received` payload now includes **`author`** (the group participant WID who actually sent the message; `from` is the group JID) and a **`contact`** object with `name` + `pushName`. Both are read from the engine (push name synchronously, saved name best-effort via `getContact()`). The change is additive, so existing webhook consumers are unaffected. Thanks for the report!

PR #178, opened 2026-06-02
fix: add message ack handling and sent message webhooks

### LINKED ISSUE #93: [BUG] Message Sent Webhook Not Working
I tried the webhook feature and have activated all the webhook configurations, but the "message.sent" event does not appear at all in the webhook event.

Image

### LINKED ISSUE #155: Callback for message.act is not initialize
Session service is initialize engine without registering onMessageAct callback that will fired message.act event.

Singh1106:
- Update onMessageAck engine callback to accept full IncomingMessage instead of just message ID and ack
- Implement onMessageAck handler in SessionService to log acks, update session activity, and dispatch webhook events
- Add webhook dispatch for all outgoing messages in MessageService
- Refine onMessage handler in SessionService to properly distinguish sent vs received messages
- Add WebhookModule dependency to MessageModule
- Clean up peer dependency metadata in package-lock.json

## Description
Fixes the missing `message.sent` webhook event reported in #93. Even with all webhook configurations enabled and `message.sent` subscribed, no webhook was being fired for outgoing messages. The root cause was that `MessageService` never dispatched a webhook after sending, and `SessionService`'s `onMessage` handler wasn't distinguishing sent vs received messages. This PR adds the webhook dispatch for outgoing messages, updates the `onMessageAck` callback to carry the full message payload, and wires `WebhookModule` into `MessageModule`.

## Type of Change
- [x] Bug fix

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] Lint passes
- [x] Self-reviewed

## Related Issues
Closes #93
Closes #155
Refs #168, #195 (partial — phone-originated `fromMe` sends need a `message_create` listener in the adapter; tracked separately and to be de-duplicated with this PR's API-send dispatch using the #179 idempotency keys)


rmyndharis (changes requested):
Thanks for tackling this, @Singh1106 — registering `onMessageAck`, dispatching `message.ack`, and firing `message.sent` for API sends are all the right ideas, and they do address the root of #93 and #155 conceptually. But a line-by-line trace against current `main` turned up several blockers that need fixing before this can merge. The most important one is that the branch is based on an old `main` and would silently revert recent fixes.

### Blocking

**1. The rewritten adapter `message` handler drops payload enrichment (reverts #146 / #202).**
This branch replaces the adapter's `onMessage` handler with a stripped synchronous object literal that omits `media`, `quotedMessage`, the new `author`, and the `contact` object. Since this PR was opened, `main` gained sender `author` + `contact` enrichment (#146) and location handling (#202). Merging as-is would **regress** `message.received` payloads back to the bare shape. Please **rebase onto current `main`** and build on the existing enriched handler rather than replacing it.

**2. Triple dispatch of `message.sent` per API-sent message.**
For an API send, `message.sent` currently fires from three independent sites: `MessageService` (right after `engine.sendTextMessage`), the adapter/`onMessage` `fromMe` branch (wwebjs fires `message` for `fromMe:true` too), and `onMessageAck`. A consumer would receive 3+ webhooks for one send. Pick a single source of truth — the cleanest is to dispatch `message.sent` once from the `MessageService` send path, and reserve `onMessageAck` for `message.ack` only.

**3. `message.received` regression for own-sent messages.**
Routing `fromMe:true` from `onMessage` to `message.sent` means subscribers who relied on `message.received` (e.g. wildcard `*`) silently stop seeing own-sent messages there. Phone-originated sends are tracked separately in #168 and need the `message_create` event — please don't fold that into this PR's `onMessage` rewrite.

**4. `onMessageAck` dispatches `message.sent` on every ACK level (0–4).**
It should dispatch `message.ack` (with the numeric ack), once, not `message.sent` per ack transition.

**5. No assertions on the new dispatch behavior.**
The added `WebhookService` mock only prevents a DI failure. Please add tests asserting: `message.sent` dispatched exactly once per send method; not double-dispatched for an API send; `onMessageAck` dispatches `message.ack` with the right ack code.

### Note on de-duplication

Once rebased, coordinate the `message.sent` key with the idempotency keys already on `main` (`idempotency.util` now uses `data.id ?? data.messageId`, from #179) so a future `message_create` listener for #168 doesn't double-emit.

The direction is good — this just needs a rebase onto current `main` (so it doesn't revert the enrichment) and a single, well-tested dispatch path. Happy to re-review promptly once that's in.

rmyndharis:
Thanks for surfacing message-ack handling and the sent-message webhook. v0.2.0 (#229) implements this via a fresh `message_create` + delivery-state design (#93 / #155 / #168 / #220) that avoids double-dispatching `message.sent`. Closing as superseded by the consolidated release.

PR #179, opened 2026-06-02
fix(webhook-received): adjust idempotency key generation

### LINKED ISSUE #168: Webhook not firing for self-messages (fromMe: true) on same session number

Singh1106:
Flip the message idempotency-key resolution to prefer `data.id` over `data.messageId` (the dispatched payload is an `IncomingMessage` carrying `id`), applied to `message.received` / `message.sent` / `message.ack` / `message.revoked`, with regression tests for the real `id`-shape payload.

## Description

`generateIdempotencyKey` resolved the message id with `toStr(data.messageId) || toStr(data.id)`. Because `toStr()` returns the truthy string `'unknown'` for a missing field, the `||` never fell through to `data.id` — so every dispatched message webhook was keyed `msg_unknown`, collapsing all messages into one bucket for consumers that deduplicate on the `X-OpenWA-Idempotency-Key` header. Resolved with nullish coalescing (`id ?? messageId`) before stringifying.

## Type of Change
- [x] Bug fix

## Checklist
- [x] Tests added/updated
- [ ] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

## Related Issues

Refs #168. **Note:** this fixes the idempotency-key collapse only. It does **not**
resolve #168 on its own — self-messages (`fromMe: true`) not firing a webhook is a
separate root cause (the engine listens to the `message` event, not `message_create`)
and remains tracked under #168 for v0.2.0. This PR is intentionally **not** auto-closing #168.

### DIFF
--- a/src/modules/webhook/utils/idempotency.util.spec.ts
+++ b/src/modules/webhook/utils/idempotency.util.spec.ts
@@ -12,6 +12,22 @@ describe('Idempotency Utils', () => {
       expect(key).toBe('ack_ABC123_3');
     });
 
+    it('should use the IncomingMessage `id` field for message.received (the real dispatch shape)', () => {
+      // session.service dispatches the IncomingMessage object, which carries `id`, not `messageId`.
+      const key = generateIdempotencyKey('message.received', { id: 'ABC123' });
+      expect(key).toBe('msg_ABC123');
+    });
+
+    it('should prefer `id` over a legacy `messageId` when both are present for message.received', () => {
+      const key = generateIdempotencyKey('message.received', { id: 'REAL', messageId: 'LEGACY' });
+      expect(key).toBe('msg_REAL');
+    });
+
+    it('should use the `id` field for message.ack (the real dispatch shape)', () => {
+      const key = generateIdempotencyKey('message.ack', { id: 'ABC123', ack: 3 });
+      expect(key).toBe('ack_ABC123_3');
+    });
+
     it('should generate key for session.status', () => {
       const key = generateIdempotencyKey('session.status', {
         sessionId: 'sess_1',
--- a/src/modules/webhook/utils/idempotency.util.spec.ts
+++ b/src/modules/webhook/utils/idempotency.util.spec.ts
@@ -28,6 +28,11 @@ describe('Idempotency Utils', () => {
       expect(key).toBe('ack_ABC123_3');
     });
 
+    it('should use the `id` field for message.revoked (the real dispatch shape)', () => {
+      const key = generateIdempotencyKey('message.revoked', { id: 'ABC123' });
+      expect(key).toBe('rev_ABC123');
+    });
+
     it('should generate key for session.status', () => {
       const key = generateIdempotencyKey('session.status', {
         sessionId: 'sess_1',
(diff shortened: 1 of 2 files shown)


rmyndharis (dismissed):
Thanks @Singh1106 — the intent (align the idempotency key with IncomingMessage.id) is right, but as-is this regresses dedup: it reads data.id only, so events that carry data.messageId now produce `<type>_unknown` and the existing idempotency.util.spec.ts fails (e.g. msg_ABC123 -> msg_unknown). Two small changes and it's a clean merge: (1) use a fallback like `data.id || data.messageId`; (2) update idempotency.util.spec.ts to cover both shapes. Thanks!

rmyndharis:
Thanks for finding this, @Singh1106 — your root-cause analysis is exactly right: the dispatched `message.received` payload is an `IncomingMessage` that carries `id`, not `messageId`.

I pushed a small follow-up to your branch to land it safely. The order-flip on its own (`toStr(data.id) || toStr(data.messageId)`) still short-circuits, because `toStr()` returns the truthy string `'unknown'` for a missing field — so `message.ack` (which carries `messageId`) would have produced `ack_unknown_N`, which is why the existing tests stayed red. The fix resolves the value **before** stringifying:

```ts
// message.received / message.sent
return `msg_${toStr(data.id ?? data.messageId)}`;
// message.ack
return `ack_${toStr(data.id ?? data.messageId)}_${toStr(data.ack, '0')}`;
```

I also added tests for the real `id`-shape payload (and `id`-over-`messageId` precedence) so this stays regression-protected. Verified against current `main`: lint + build + 154/154 tests green.

This is queued for the upcoming **v0.1.8** patch. Thanks again for the contribution!

Issue #192, opened 2026-06-08
send-bulk endpoint always returns 400 Bad Request for text messages


carlosjcuello:
## Bug Report

The `/api/sessions/:sessionId/messages/send-bulk` endpoint returns 400 Bad Request for all requests due to missing `@IsOptional()` decorators in `BulkMessageContentDto`.

## Root Cause

In `src/modules/message/dto/bulk-message.dto.ts`, the fields `image`, `video`, `audio`, `document`, and `caption` have `@ApiPropertyOptional` but no class-validator decorator:

```typescript
image?: { url?: string; base64?: string; mimetype?: string };
video?: { url?: string; base64?: string; mimetype?: string };
audio?: { url?: string; base64?: string; mimetype?: string };
document?: { url?: string; base64?: string; mimetype?: string; filename?: string };
caption?: string;
```

When `class-transformer` instantiates `BulkMessageContentDto` via `@Type()`, it adds these as `undefined` properties on the object. Since the global `ValidationPipe` is configured with `forbidNonWhitelisted: true`, any property without a class-validator decorator is rejected — including these undefined ones.

## Steps to Reproduce

```bash
POST /api/sessions/:sessionId/messages/send-bulk
{
  "messages": [{
    "chatId": "1234567890@c.us",
    "type": "text",
    "content": { "text": "Hello" }
  }]
}
```

## Response

```json
{
  "message": [
    "messages.0.content.property image should not exist",
    "messages.0.content.property video should not exist",
    "messages.0.content.property audio should not exist",
    "messages.0.content.property document should not exist"
  ],
  "error": "Bad Request",
  "statusCode": 400
}
```

## Fix

Add `@IsOptional()` to each affected field in `BulkMessageContentDto`.

## Environment

- OpenWA v0.1.6
- Docker deployment

rmyndharis:
Fixed by #193 (add @IsOptional() to the bulk message media fields), merged to main.

PR #193, opened 2026-06-08
fix: add @IsOptional() to BulkMessageContentDto media fields

carlosjcuello:
## Problem

The `send-bulk` endpoint always returns 400 Bad Request for text messages, even with a valid request body.

When `class-transformer` instantiates `BulkMessageContentDto` via `@Type()`, it adds `image`, `video`, `audio`, and `document` as `undefined` properties on the object. Since the global `ValidationPipe` is configured with `forbidNonWhitelisted: true`, these properties are rejected because they lack class-validator decorators.

## Fix

Added `@IsOptional()` to `image`, `video`, `audio`, and `document` fields in `BulkMessageContentDto` so they are recognized as whitelisted optional properties.

## Testing

Verified that `POST /api/sessions/:sessionId/messages/send-bulk` with a text message payload now returns 202 and the message is delivered successfully.

### DIFF
--- a/src/modules/message/dto/bulk-message.dto.ts
+++ b/src/modules/message/dto/bulk-message.dto.ts
@@ -19,15 +19,19 @@ class BulkMessageContentDto {
   text?: string;
 
   @ApiPropertyOptional({ description: 'Image URL or base64' })
+  @IsOptional()
   image?: { url?: string; base64?: string; mimetype?: string };
 
   @ApiPropertyOptional({ description: 'Video URL or base64' })
+  @IsOptional()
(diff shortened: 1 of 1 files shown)


rmyndharis (approved):
Verified: the four media fields (image/video/audio/document) gain @IsOptional() — IsOptional is already imported in this file — so a text-only bulk payload is no longer rejected by the global whitelist/forbidNonWhitelisted ValidationPipe. Minimal, correct, zero regression risk (fields are optional and read via optional chaining in the service). Thanks @carlosjcuello, and for filing #192!

PR #200, opened 2026-06-11
feat(groups): expose linkedParentJID for community sub-groups

ferhatte10:
## Summary

  - Adds a `linkedParentJID` field to the group response objects returned by:
    - `GET /api/sessions/{sessionId}/groups`
    - `GET /api/sessions/{sessionId}/groups/{groupId}`
  - For groups that belong to a WhatsApp Community, this field contains the
    parent community's JID (e.g. `120363416871560789@g.us`). For standalone
    groups it is `null`.
  - This mirrors the `LinkedParentJID` field already exposed by WAHA, allowing
    clients to filter community sub-groups with
    `groups.filter(g => g.linkedParentJID === communityJID)`.

  ## Implementation details

  - `Group` and `GroupInfo` (in `src/engine/interfaces/whatsapp-engine.interface.ts`)
    now declare `linkedParentJID?: string | null`.
  - `src/engine/types/whatsapp-web-js.types.ts` adds `SerializedWid` and
    `GroupMetadataRaw` types, and exposes `groupMetadata` on `GroupChat`.
  - `src/engine/adapters/whatsapp-web-js.adapter.ts` adds a small
    `extractLinkedParentJID()` helper that reads the parent-community JID from
    the raw `groupMetadata` object. Since the exact field name has varied across
    whatsapp-web.js / WA Web versions, it checks several known candidates
    (`parentGroup`, `linkedParentGroup`, `linkedParent`) and falls back to `null`
    if none are present.
  - Both `getGroups()` and `getGroupInfo()` now populate `linkedParentJID`.

  ## Other fix included

  - `dashboard/Dockerfile`: `npm ci` → `npm ci --legacy-peer-deps`. The dashboard
    build currently fails out of the box because `@vitejs/plugin-react@5.1.4`
    declares a peer dependency on `vite@^4 || ^5 || ^6 || ^7`, while
    `package.json` pins `vite@^8.0.13`, causing `npm ci` to fail with
    `ERESOLVE`. This flag unblocks the Docker build without changing the
    dependency versions.

  ## Testing

  - `npm run build` (NestJS) compiles successfully with the new types.
  - Dashboard Docker image builds successfully with the `--legacy-peer-deps` fix.

  ## Behavior

  | Group type            | `linkedParentJID`                  |
  |------------------------|-------------------------------------|
  | Community sub-group    | `"120363416871560789@g.us"`         |
  | Standalone group        | `null`                              |

### DIFF
--- a/dashboard/Dockerfile
+++ b/dashboard/Dockerfile
@@ -6,7 +6,7 @@ WORKDIR /app
 COPY package*.json ./
 
 # Install dependencies
-RUN npm ci
+RUN npm ci --legacy-peer-deps
 
 # Copy source code
 COPY . .
--- a/src/engine/adapters/whatsapp-web-js.adapter.ts
+++ b/src/engine/adapters/whatsapp-web-js.adapter.ts
@@ -30,6 +30,7 @@ import {
 import { createLogger } from '../../common/services/logger.service';
 import {
   GroupChat,
+  GroupMetadataRaw,
   MessageWithReactions,
   BusinessClient,
   WwjsChannelData,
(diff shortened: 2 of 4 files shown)


ferhatte10:
Closing this PR in favor of a follow-up that includes an additional fix.
  
  While testing the linkedParentJID changes against a live deployment, I found that GET /api/sessions/:id/groups is actually served by SessionController/SessionService (not
  GroupController), which maps the engine's group list down to { id, name } before returning it — so the new linkedParentJID field never reached the response on the list
  endpoint, even though it worked correctly on GET /api/sessions/:id/groups/:groupId.

  I'll open a new PR that includes the original linkedParentJID changes plus this mapping fix, so the field is consistently available on both endpoints.

PR #201, opened 2026-06-11
feat(groups): expose linkedParentJID for community sub-groups

ferhatte10:
## Summary

  Adds a `linkedParentJID` field to the group endpoints so API consumers can tell whether a group is a sub-group of a WhatsApp Community, and which community it belongs to.
  
  - `GET /api/sessions/{sessionId}/groups`
  - `GET /api/sessions/{sessionId}/groups/{groupId}`
  
  Both endpoints now return:
  
```json
  {
    "id": "126663421933678702@g.us",
    "name": "Groupe Name",
    "linkedParentJID": "120363416871160700@g.us"
  }
```

  For standalone groups (not part of a community), `linkedParentJID` is `null`.
  
  ## Changes

  - **`src/engine/types/whatsapp-web-js.types.ts`**: added `SerializedWid` and `GroupMetadataRaw` types, and a `groupMetadata` field on `GroupChat`. The underlying field name
  for the parent community link has varied across whatsapp-web.js / WA Web versions, so multiple known candidates (`parentGroup`, `linkedParentGroup`, `linkedParent`) are
  declared and checked.
  - **`src/engine/interfaces/whatsapp-engine.interface.ts`**: added optional `linkedParentJID?: string | null` to the `Group` and `GroupInfo` interfaces.
  - **`src/engine/adapters/whatsapp-web-js.adapter.ts`**:
      - Added `extractLinkedParentJID()` helper that reads the parent community JID from `groupMetadata`, handling the different field-name candidates and serialized WID shapes.
      - `getGroupInfo()` now populates `linkedParentJID` from `chat.groupMetadata`.
      - `getGroups()` now also populates `linkedParentJID`. Since `client.getChats()` doesn't always have `groupMetadata` populated for groups that haven't been loaded into the
  WhatsApp Web store yet, `getGroups()` falls back to `getChatById()` (which reliably loads full group metadata) when it's missing from the chat list.
  - **`src/modules/session/session.service.ts` / `session.controller.ts`**: `GET /api/sessions/{id}/groups` is actually served by `SessionController` (it takes priority over
  `GroupController` for this route), and `SessionService.getGroups()` was mapping the engine result down to `{ id, name }`, silently dropping any extra fields. Updated the
  mapping and return types to include `linkedParentJID` so the list endpoint is consistent with the single-group endpoint.
  - **`dashboard/Dockerfile`**: `npm ci` → `npm ci --legacy-peer-deps` to fix an `ERESOLVE` conflict between `@vitejs/plugin-react@5.1.4` and `vite@8.0.13` that currently breaks
  the dashboard Docker build.
  
  ## Testing

  - Built and deployed both the API and dashboard images via Docker.
  - Verified `GET /api/sessions/{id}/groups` returns `linkedParentJID` with the correct community JID for sub-groups and `null` for standalone groups, matching `GET
  /api/sessions/{id}/groups/{groupId}`.

### DIFF
--- a/dashboard/Dockerfile
+++ b/dashboard/Dockerfile
@@ -6,7 +6,7 @@ WORKDIR /app
 COPY package*.json ./
 
 # Install dependencies
-RUN npm ci
+RUN npm ci --legacy-peer-deps
 
 # Copy source code
 COPY . .
--- a/src/engine/adapters/whatsapp-web-js.adapter.ts
+++ b/src/engine/adapters/whatsapp-web-js.adapter.ts
@@ -388,18 +388,44 @@ export class WhatsAppWebJsAdapter extends EventEmitter implements IWhatsAppEngin
     // Filter only group chats
     const groups = chats.filter(chat => chat.isGroup);
 
-    return groups.map(g => {
-      const groupChat = g as unknown as GroupChat;
-      return {
-        id: g.id._serialized,
-        name: g.name,
-        participantsCount: groupChat.participants?.length,
-        isAdmin: groupChat.participants?.some(
-          p => p.isAdmin && p.id._serialized === this.client?.info?.wid?._serialized,
-        ),
-        linkedParentJID: extractLinkedParentJID(groupChat.groupMetadata),
-      };
-    });
+    return Promise.all(
+      groups.map(async g => {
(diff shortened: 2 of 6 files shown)


rmyndharis:

PR #203, opened 2026-06-12
Security hardening, dependency CVE fixes, and dev-experience fixes

rafaelfiguereod-stack:
## Summary

End-to-end review of OpenWA (clone → install → run → audit → fix → verify). Three independent, reviewable commits:

1. **Security hardening** (`d2c2729`)
2. **Dev-experience / functional fixes** (`c7810ee`)
3. **Dependency CVE fixes** (`4234bd4`)

All 118 backend tests pass; backend (`nest build`) and dashboard (`tsc` + `vite build`) compile and lint clean; the core flows were verified on a live instance, including a real-browser check of login → create/start session → live QR. Full findings and reproduction details are in [`SECURITY_AUDIT.md`](SECURITY_AUDIT.md).

## 🔒 Security hardening

> **For maintainers:** the default-admin-credential item below is severe enough that you may want to publish a GitHub Security Advisory alongside merging this.

- **Predictable default admin API key in non-production mode.** `AuthService` seeded the default ADMIN key as the hard-coded constant `dev-admin-key` whenever `NODE_ENV !== 'production'` — which includes the README's primary quick-start (`docker-compose.dev.yml` sets `NODE_ENV=development`) and `npm run dev`. Any reachable dev-mode instance accepted a globally-known admin key. Now a cryptographically random key is always generated (an explicit `API_MASTER_KEY` is honored if set).
- **Path traversal (tar-slip) in storage import.** Archive entry names / caller-supplied paths in the local storage backend are now contained within the storage root, so a crafted `.tar.gz` can no longer write outside `data/media`.
- **Broken access control.** The `infra` control plane (Docker orchestration, DB & storage import/export, env/config writes) and the mutating plugin routes had no role requirement and were callable by any key. They now require the `ADMIN` role.
- **IP-allowlist bypass.** A per-key `allowedIps` restriction was defeatable by spoofing `X-Forwarded-For`; that header is now only trusted when `TRUST_PROXY=true`.
- Wildcard CORS no longer enables credentials; WebSocket subscriptions enforce a key's `allowedSessions`; opt-in webhook SSRF guard (`WEBHOOK_SSRF_PROTECT`, default off); `ENABLE_SWAGGER` is honored; added a `HOST` bind knob.

## 🛠 Dev-experience / functional fixes

- **Dashboard real-time events were inert.** `useWebSocket` never sent a `subscribe` message (the gateway only emits to joined rooms) and listened on `session:status|qr|message` while the server emits on the `message` channel with a `{type:'event', payload:{event,sessionId,data}}` envelope. It now subscribes on connect and translates the envelope into the existing callbacks. Verified live: the session status badge updates over WS with no reload, and the QR renders.
- **Vite dev server only proxied `/api`, not `/socket.io`**, so live events never reached the dashboard under `npm run dev`. Added a ws-enabled `/socket.io` proxy (matching `nginx.conf`).
- **Engine launch was fragile.** Chrome cold-start could exceed Puppeteer's default 30s WS-endpoint wait and surface as an opaque 500 / `status=failed`. Added an env-tunable `PUPPETEER_LAUNCH_TIMEOUT` (default 60s) and `PUPPETEER_EXECUTABLE_PATH` passthrough.
- Corrected the dashboard `ApiKey.role` type from the stale `admin|user|readonly` to the backend's `admin|operator|viewer`.

## 📦 Dependency CVE fixes

`npm audit`: **21 → 8 vulnerabilities (2 critical → 0)**.

- `typeorm` `^0.3.29 → ^0.3.30` — fixes the SQL-injection advisory in `repository.save`/`update` (a real runtime path here).
- `concurrently` `^9.2.1 → ^10.0.3` (devDep) — clears the two **critical** `shell-quote` advisories.
- `npm audit fix` also resolved the `ws` advisory transitively.
- `sqlite3` intentionally kept at `^5.1.7`: `typeorm@0.3.30` declares `peerOptional sqlite3@^5.0.3`, so `sqlite3@6` is unsupported by the ORM. The remaining 8 advisories are entirely in `sqlite3`'s install-time `node-gyp`/`tar`/`cacache` toolchain, which never runs at application runtime.

## ✅ Validation

- 118/118 backend tests pass; backend + dashboard build and lint clean.
- Booted the stack and confirmed ORM read/write to SQLite, login, session create → start → QR, and live WebSocket status updates in a real (Puppeteer-driven) browser session.

### DIFF
--- a/SECURITY_AUDIT.md
+++ b/SECURITY_AUDIT.md
@@ -1,0 +1,174 @@
+# OpenWA — Security Audit
+
+**Target:** `rmyndharis/OpenWA` (Open Source WhatsApp API Gateway), v0.1.6
+**Commit audited:** `0fbee7f` (branch `main`)
+**Stack:** NestJS 11 / TypeScript 5 / TypeORM (SQLite default, PostgreSQL optional) / whatsapp-web.js (Puppeteer) / Socket.IO / dockerode / React dashboard
+**Scope:** Backend `src/**` (~13.8K LOC, 121 files). Dashboard reviewed at a high level only.
+**Method:** Manual code review of the full backend, plus a live run (`node dist/main`) to confirm exploitability. Dependency posture via `npm audit`.
+**Date:** 2026-06-11
+
+---
+
+## 0. Executive summary
+
+OpenWA has a **reasonable baseline**: a global `ApiKeyGuard` (`APP_GUARD`) protects every route unless explicitly `@Public()`, inputs are validated with `class-validator` + a strict `ValidationPipe` (`whitelist`, `forbidNonWhitelisted`), Helmet sets security headers, throttling is on, webhooks support HMAC signing, and DB access is parameterized (no SQL injection found). The WebSocket gateway authenticates the handshake.
+
+The dominant problem is the **default credential and the access model around the privileged "infra" control plane**:
+
+| # | Severity | Title | Status |
+|---|----------|-------|--------|
+| C-1 | **Critical** | Hard‑coded, globally‑known default admin key `dev-admin-key` in non‑production mode (the README's #1 Quick Start) | **Fixed** |
+| H-1 | High | Tar‑slip / path traversal in storage import + arbitrary local file read (`/infra/storage/import`) | **Fixed** |
+| H-2 | High | Broken access control — destructive infra & plugin endpoints not gated to ADMIN | **Fixed** |
+| H-3 | High | API‑key IP allowlist bypass via spoofable `X-Forwarded-For` | **Fixed** |
+| M-1 | Medium | CORS reflects any origin **with credentials** when `CORS_ORIGINS=*` (the default) | **Fixed** |
+| M-2 | Medium | WebSocket subscriptions ignore the key's `allowedSessions` (cross‑session message leak) | **Fixed** |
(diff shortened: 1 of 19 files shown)


rmyndharis (changes requested):
Thanks @rafaelfiguereod-stack — genuinely useful audit. It's a 19-file / 3-concern bundle though, so it needs splitting and rebasing on current `main`:

- **Overlaps in-flight PRs:** the X-Forwarded-For work duplicates #211, and the tar-slip containment + infra RBAC duplicate #207 — please let those own those areas rather than double-applying.
- **High-value & still-novel** (worth a focused PR): removing the hard-coded `dev-admin-key` fallback (even #133 kept it), the WS `allowedSessions` enforcement (cross-tenant leak — but review it together with the dashboard `subscribe('*')` so scoped keys don't lose events), the opt-in SSRF guard, CORS-credentials-with-wildcard, `ENABLE_SWAGGER`, and the `typeorm@0.3.30` advisory bump.
- **Please also** drop `SECURITY_AUDIT.md` and the AI-attribution line from the PR body, and hold `concurrently@10` until Node ≥22 is confirmed in CI.

Moving to v0.2.0. Thanks for the thorough work!

rmyndharis:

Issue #204, opened 2026-06-12
PostgreSQL SSL configuration is ignored in runtime TypeORM connection


farrasyakila:
### Description
When running OpenWA with PostgreSQL behind PgBouncer (e.g. Percona Everest, CrunchyData, Cloud providers, or any PostgreSQL deployment that enforces TLS), the application fails to connect even when DATABASE_SSL=true is configured.

The application repeatedly retries database connections and eventually exits with:

Error: SSL required

### Environment : 

OpenWA version: v0.1.6
Database: PostgreSQL
Connection target: PgBouncer / PostgreSQL requiring SSL
DATABASE_TYPE=postgres
DATABASE_SSL=true

### Current Behavior

The application fails to establish a database connection and logs:

error: SSL required

### Investigation

The configuration layer correctly exposes:

dataDatabase.ssl
dataDatabase.sslRejectUnauthorized

However, the runtime TypeORM configuration in AppModule does not pass SSL settings into the PostgreSQL connection options.

### Current implementation:

```
return {
...baseConfig,
type: 'postgres',
host: configService.get<string>('dataDatabase.host'),
port: configService.get<number>('dataDatabase.port'),
username: configService.get<string>('dataDatabase.username'),
password: configService.get<string>('dataDatabase.password'),
database: 'openwa',
synchronize: configService.get<boolean>('dataDatabase.synchronize', false),
migrationsRun: true,
retryAttempts: 10,
retryDelay: 3000,
extra: {
max: configService.get<number>('dataDatabase.poolSize', 10),
},
};
```

The SSL configuration exists in configuration.ts but is never applied to the TypeORM runtime connection.

### Expected Behavior

When DATABASE_SSL=true is configured, the TypeORM PostgreSQL connection should enable TLS and pass the SSL options to the underlying pg driver.

### Impact

This prevents OpenWA from connecting to PostgreSQL installations that require SSL, including:

- Percona Everest PostgreSQL
- PgBouncer deployments configured for TLS
- Managed PostgreSQL services
- Any PostgreSQL cluster enforcing encrypted connections

### Suggested Fix

Pass the SSL configuration from ConfigService into the TypeORM PostgreSQL options.

PR #205, opened 2026-06-12
fix(database): add SSL support for runtime PostgreSQL TypeORM connection

farrasyakila:
## Description
fix for https://github.com/rmyndharis/OpenWA/issues/204
This change adds SSL configuration support to the runtime PostgreSQL TypeORM connection in AppModule.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] Lint passes
- [x] Self-reviewed

### DIFF
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -86,16 +86,11 @@ if (process.env.QUEUE_ENABLED === 'true') {
             password: configService.get<string>('dataDatabase.password'),
             database: 'openwa',
 
-            ssl:
-              configService.get<boolean>('dataDatabase.ssl', false)
-                ? {
-                    rejectUnauthorized:
-                      configService.get<boolean>(
-                        'dataDatabase.sslRejectUnauthorized',
-                        false,
-                      ),
-                  }
-                : false,
+            ssl: configService.get<boolean>('dataDatabase.ssl', false)
+              ? {
+                  rejectUnauthorized: configService.get<boolean>('dataDatabase.sslRejectUnauthorized', false),
+                }
+              : false,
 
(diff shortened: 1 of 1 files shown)


rmyndharis (changes requested):
Thanks @farrasyakila — and thanks for the precise repro in #204. The fix is correct: it wires SSL into the runtime `data` connection in app.module.ts, matching the pattern already in data-source.ts, and it is secure by default (rejectUnauthorized stays effectively `true` unless DATABASE_SSL_REJECT_UNAUTHORIZED=false, since configuration.ts always resolves a concrete boolean).

Two small things before we can merge:

1. **Formatting (blocking):** `npx prettier --check src/app.module.ts` fails on the new `ssl` block, so the CI format step would go red. Please run `npm run format` and commit the result.
2. **Defense-in-depth (nice-to-have):** change the fallback in `configService.get<boolean>(\"dataDatabase.sslRejectUnauthorized\", false)` to `true`. It is inert today (configuration.ts always provides the value), but `true` is the safer fallback if that ever changes.

Once the formatting is fixed I am happy to merge. Appreciate the contribution!

farrasyakila:
Thanks for the review! @rmyndharis 

I've addressed both points:

- Ran Prettier and pushed the formatting changes for `src/app.module.ts`, `npx prettier --check src/app.module.ts` now passes.

- Updated the fallback value of `dataDatabase.sslRejectUnauthorized` from **false** to **true** as suggested.

Thank you.

PR #215, opened 2026-06-13
fix(deps): bump concurrently to v10 to clear critical shell-quote advisory

rmyndharis:
## Why
The **Lint** CI job's `npm audit --audit-level=critical` step is failing on `main` (last several runs red) and therefore on **every open PR**, blocking all merges. The two critical advisories are `shell-quote` pulled in transitively by `concurrently@9`.

## Fix
Bump `concurrently` 9 → **10.0.3**, which ships a patched `shell-quote`. It's a **devDependency** (only `npm run dev` uses it) and requires Node ≥22 — already the project/CI/Docker baseline (`NODE_VERSION: '22'`, `node:22-slim`).

## Verified
- `npm audit --audit-level=critical` → **exit 0** (was 1)
- `npm run build` → passes

This needs to land first so the rest of the v0.1.7 queue can go green.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,7 +57,7 @@
         "@types/supertest": "^6.0.2",
         "@types/tar-stream": "^3.1.4",
         "@types/uuid": "^11.0.0",
-        "concurrently": "^9.2.1",
+        "concurrently": "^10.0.3",
         "eslint": "^10.4.1",
         "eslint-config-prettier": "^10.0.1",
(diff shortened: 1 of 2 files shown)


rmyndharis:
Folded into #208, which also switches `lint` to check mode. Neither fix passes CI alone (this one's old `--fix` lint surfaces a non-fixable enum error; #208's old concurrently fails the audit), so they need to land together. Closing in favor of #208.

Issue #220, opened 2026-06-13
[Epic] Reflect real session & delivery state to the API (#199, #100)


rmyndharis:
## Theme

Umbrella for "the engine's real state is not faithfully reflected to the API/dashboard." Two open issues share this theme but have **distinct fixes** — tracking them together while keeping each as its own issue.

### #199 — Group send returns 200 + messageId but is silently never delivered (after LID migration)

`sendTextMessage` ([adapter L283-290](https://github.com/rmyndharis/OpenWA/blob/main/src/engine/adapters/whatsapp-web-js.adapter.ts#L283-L290)) calls `client.sendMessage` and immediately returns `msg.id._serialized` with **no ack inspection**, and `MessageService` marks the message `SENT` regardless. The session also stays `ready`. After WhatsApp's LID migration this silently drops messages.
**Needs:** surface send-ack state (don't return a bare 200+messageId for un-ack'd sends), structured 4xx on LID errors, an optional `degraded` session state, and an upstream-LID tracking note.

### #100 — HTTP 500 on dashboard refresh right after disconnecting on the phone

Refreshing the Sessions view during engine teardown/transition returns `Internal server error`.
**Needs:** defensive handling in the read endpoints for transitional/torn-down engine state. (Awaiting the exact failing request + stack trace already requested on the issue.)

### Common thread

The API and dashboard should reflect broken/transitional engine state honestly — neither return success (200) nor crash (500). Keep #199 and #100 as separate issues with distinct fixes; this epic only groups them.

Issue #199, opened 2026-06-11
Group send-text returns 200 + messageId but is silently never delivered after WhatsApp LID migration (session stays "ready")


720pixel:
## Summary

After WhatsApp's **LID (LinkedID) migration** reached one of our linked numbers, the gateway began **silently dropping outgoing messages**. The critical and (as far as I can tell) **undocumented** part is the **group** case: **`send-text` to a `@g.us` group returns `200` with a real `messageId`, the session stays `status: ready`, yet the message is never delivered to the group.** Automated senders (alerting, etc.) get a success response and store a "sent" record while the group receives nothing.

This is related to but distinct from #156 and upstream pedroslopez/whatsapp-web.js#3834 — those cover **individual** (`@c.us`) LID resolution (`No LID for user`), and their workarounds (the `@s.whatsapp.net` swap, the `getChat` open-chat-info / USync overrides) all target a *contact's* LID. **None of them address group sending**, which is the path that silently fails here.

We verified the group failure on a **clean, freshly-relinked session with the phone online** (ruling out congestion, stale credentials, and an offline phone).

## Environment

- `openwa` **0.1.6** (self-hosted, systemd)
- Engine: **whatsapp-web.js 1.34.7** (latest published) via the bundled `whatsapp-web.js` adapter + `puppeteer-core`
- One of the linked numbers had been **LID-migrated server-side by WhatsApp** (rollout is gradual/per-account)

## Observed behavior

| Send target | API result | Actually delivered? |
|---|---|---|
| Group (`<id>@g.us`) | `200` + `{"messageId":"true_..._<sender>@lid", ...}` | **No** |
| Contact with **no prior chat** (`<num>@c.us`) | **hangs**, then `500` after puppeteer `protocolTimeout` | No |
| Contact with existing chat | intermittent | intermittent |
| Same account via the **phone app** | n/a | **Yes** (account is healthy) |

Throughout, `GET /api/sessions` reports the session as `status: "ready"`.

## Errors seen in logs (`journalctl -u openwa.service`)

```
ERROR [ExceptionsHandler] Error: Lid is missing in chat table
    at Client.sendMessage (whatsapp-web.js/src/Client.js:1533)
    at WhatsAppWebJsAdapter.sendTextMessage (dist/engine/adapters/whatsapp-web-js.adapter.js:236)
    at MessageService.sendText (dist/modules/message/message.service.js:44)

ERROR [ExceptionsHandler] Error: No LID for user

[...]

Issue #195, opened 2026-06-08
Webhook/event missing for messages sent manually from linked mobile device


nalmeida84:
## Summary

  When the WhatsApp session owner sends a message manually from the linked mobile WhatsApp app to an existing customer chat, OpenWA does not emit
  any webhook event for that message.

  The message is visible in WhatsApp, but it never reaches the webhook receiver, so the consuming application cannot record it in the conversation
  history.

  ## Environment

  - OpenWA image/API: `ghcr.io/rmyndharis/openwa`
  - Webhook transport: HTTP POST
  - Registered webhook events: `message.received`, `message.sent`, `session.status`
  - Swagger/OpenAPI available event enum includes: `message.received`, `message.sent`, `message.ack`, `message.revoked`, `session.status`,
  `session.qr`, `session.authenticated`, `session.disconnected`, `group.join`, `group.leave`, `group.update`
  - OS: Linux / Docker Compose

  ## Steps to Reproduce

  1. Start an OpenWA session connected to WhatsApp number A.
  2. Register a webhook with `message.received`, `message.sent`, and `session.status`.
  3. Receive or open an existing customer conversation in WhatsApp.
  4. From the linked mobile WhatsApp app using number A, manually send a message to that customer chat, for example: `OPERADOR`.
  5. Check the webhook receiver logs and the OpenWA message/audit endpoints.

  ## Expected Behaviour

  OpenWA should emit a webhook event for the manually sent message, ideally with a flag such as `fromMe: true`, `direction: outgoing`, or
  equivalent metadata.

  This would allow the consuming application to distinguish between:

  - inbound customer messages
  - outbound messages sent by the bot/API
  - outbound messages manually written by the human operator from WhatsApp mobile

  ## Actual Behaviour

  No webhook event is fired for the manually sent mobile WhatsApp message.

  The webhook receiver receives nothing at all for that message, not even a raw payload. The message also does not appear in the API data checked
  by the consuming application.

  ## What Does Work

  - `message.received` webhooks fire correctly for customer messages.
  - `session.status` webhooks fire correctly.
  - Messages sent through the API can be tracked by the application.
  - The webhook registration is active and contains the configured supported events.

  ## Use Case

  This is needed for operator handoff workflows.

  For example, an appointment bot manages a customer conversation automatically, but when a human operator replies manually from the same WhatsApp
  mobile app, that reply should also appear in the bot/operator conversation timeline.

[...]

PR #225, opened 2026-06-13
fix(session): persist message_ack so delivery/read status is tracked

aminebalti55:
## Problem

The whatsapp-web.js engine emits `message_ack` events, and the adapter (`whatsapp-web-js.adapter.ts`) already forwards them through the `onMessageAck` engine callback. But `SessionService` never implemented `onMessageAck`, so every delivery/read receipt was silently dropped.

As a result, an outgoing message's `status` is set to `sent` at send time and **never updated**. `MessageStatus.DELIVERED` and `MessageStatus.READ` exist in the enum but were never written — so the message history and dashboard could never reflect real delivery state (every message appears stuck at `sent` forever, even after it's delivered and read).

## Fix

Implement the `onMessageAck` callback in `SessionService`:

- Map whatsapp-web.js ack levels to `MessageStatus`: `-1 → failed`, `1 → sent`, `2 → delivered`, `3`/`4 → read`.
- Look up the stored message by `(sessionId, waMessageId)` and update its `status`.
- Never downgrade an already-higher status (e.g. a late `delivered` arriving after `read`).
- Ignore acks for messages that aren't tracked (e.g. incoming).

`SessionModule` now registers the `Message` repository so the service can persist updates. It injects the **repository**, not `MessageService`, so there's no new circular dependency.

## Validation

Built and run locally. An outgoing message now advances `sent → delivered → read` in `GET /sessions/:id/messages`, instead of remaining `sent` indefinitely.

## Notes

- No schema/migration change — the `DELIVERED`/`READ` enum values already existed.
- Stats counts are unaffected; they key off `direction`, not `status`.

### DIFF
--- a/src/modules/session/session.module.ts
+++ b/src/modules/session/session.module.ts
@@ -1,12 +1,13 @@
 import { Module, forwardRef } from '@nestjs/common';
 import { TypeOrmModule } from '@nestjs/typeorm';
 import { Session } from './entities/session.entity';
+import { Message } from '../message/entities/message.entity';
 import { SessionService } from './session.service';
 import { SessionController } from './session.controller';
 import { WebhookModule } from '../webhook/webhook.module';
 
 @Module({
-  imports: [TypeOrmModule.forFeature([Session], 'data'), forwardRef(() => WebhookModule)],
+  imports: [TypeOrmModule.forFeature([Session, Message], 'data'), forwardRef(() => WebhookModule)],
   controllers: [SessionController],
   providers: [SessionService],
   exports: [SessionService],
--- a/src/modules/session/session.service.ts
+++ b/src/modules/session/session.service.ts
@@ -9,6 +9,7 @@ import {
 import { InjectRepository, InjectDataSource } from '@nestjs/typeorm';
 import { Repository, In, DataSource } from 'typeorm';
 import { Session, SessionStatus } from './entities/session.entity';
+import { Message, MessageStatus } from '../message/entities/message.entity';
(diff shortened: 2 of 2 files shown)


rmyndharis:
Thank you for this, @aminebalti55 — and for the precise problem write-up. You independently identified exactly the gap we were closing for the upcoming **v0.2.0** release: `SessionService` never implemented `onMessageAck`, so `MessageStatus.DELIVERED`/`READ` were never written and outgoing messages stayed stuck at `sent`.

It is a genuinely good implementation — notably your `statusRank` monotonicity guard (not downgrading a `read` back to `delivered`) matches the design we arrived at independently, which is nice validation that this is the right approach. 👏

We are delivering this as part of a broader delivery-state + webhook change in v0.2.0 that is a **superset** of this PR:

- persists ack → `MessageStatus` (delivered / read / failed), same as here;
- additionally dispatches the `message.ack` webhook (#155) and a new `message.failed` webhook so consumers can react to delivery state without polling;
- also adds `message.sent` for messages composed on a linked phone, via the `message_create` event (#168 / #195);
- and applies the monotonicity guard as an **atomic conditional `UPDATE`** (`WHERE status IN (…lower ranks…)`) rather than a read-modify-write, so it is safe against out-of-order acks landing concurrently.

Because that work covers everything here (and a bit more), I am closing this PR as **superseded** — but the credit is yours for surfacing and fixing it independently, and you will be acknowledged in the v0.2.0 changelog. Thank you for contributing! 🙏

Refs #199, #220, #155.

PR #228, opened 2026-06-14
fix(security): run container process as non-root via gosu entrypoint

A831ARD0:
- Install gosu in the production stage of the Dockerfile
- Add docker-entrypoint.sh: runs as root, chowns /app/data for named-volume mounts, then exec-drops to the openwa user via gosu before starting node
- Fix entrypoint chain: ENTRYPOINT [dumb-init,--,/usr/local/bin/docker-entrypoint.sh] so dumb-init is PID 1, the script runs under it, and node is its direct child for clean SIGTERM forwarding (resolves the missing -- bug from PR #129 review)
- Add scripts/smoke-test-non-root.sh: builds the image and asserts the process runs as openwa (uid != 0) and that dumb-init is PID 1
- Add .gitattributes to enforce LF line endings for shell scripts
- Document the startup chain in README

Splits from PR #129 (entrypoint/non-root only; bull-board-auth via #214).

## Description
Brief description of changes

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] Lint passes
- [ ] Self-reviewed

## Related Issues
Closes #

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# Force LF line endings for files that must run inside Linux containers
+*.sh        text eol=lf
+Dockerfile* text eol=lf
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,6 +49,7 @@ RUN apt-get update && apt-get install -y \
     libxrandr2 \
     xdg-utils \
     dumb-init \
+    gosu \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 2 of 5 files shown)


rmyndharis:

PR #245, opened 2026-06-15
fix(docker): bind dev ports to 0.0.0.0 for remote access

Stanley-blik:
## Problem

The dev compose file binds both the API (2785) and dashboard (2886) to `127.0.0.1`, which prevents access from external hosts. This makes it impossible to use the dashboard when running OpenWA on a remote VPS/server.

## Solution

Changed port bindings from `127.0.0.1:PORT:PORT` to `0.0.0.0:PORT:PORT` in `docker-compose.dev.yml` so the services are accessible externally during development.

The production compose (`docker-compose.yml`) is unchanged — this only affects the dev environment.

## Testing

- Deployed on a remote VPS
- Confirmed dashboard accessible at `http://<vps-ip>:2886`
- Confirmed API accessible at `http://<vps-ip>:2785/api`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,6 +15,12 @@ LOG_LEVEL=info                      # error | warn | info | debug
 DOMAIN=localhost
 DASHBOARD_PORT=2886
 
+# Host interface the dev compose binds the API/dashboard ports to. Defaults to 127.0.0.1
+# (localhost only). Set to 0.0.0.0 to reach the dev stack from another machine (e.g. a remote
+# VPS) — put a TLS reverse proxy in front for anything internet-facing (the API key is sent in
+# cleartext over plain HTTP).
+# BIND_HOST=0.0.0.0
+
 # Public URLs (for startup banner and external access)
 # BASE_URL=https://api.yourdomain.com
 # DASHBOARD_URL=https://dashboard.yourdomain.com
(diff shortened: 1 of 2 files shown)


rmyndharis:
Thanks @Stanley-blik — you're right that pinning the dev stack to `127.0.0.1` makes remote/VPS access impossible, and that's a real gap worth closing. 🙌

One thing I'd like to preserve: that `127.0.0.1` binding is a deliberate safety default (the prod compose pins it too and exposes externally only via Traefik), so flipping it to `0.0.0.0` unconditionally would expose **every** dev stack on the network by default. To keep your use case working *without* changing the default for everyone, I've pushed a small tweak to your branch making the bind host configurable via `.env`:

```yaml
ports:
  - '${BIND_HOST:-127.0.0.1}:2785:2785'
```

Default stays `127.0.0.1`; on your VPS you just set `BIND_HOST=0.0.0.0` in `.env` (now documented in `.env.example`). I verified with `docker compose config` that it resolves from `.env`.

Heads-up for public exposure: over plain HTTP the `X-API-Key` travels in cleartext and the dashboard's secure-context features are limited, so I'd recommend a TLS reverse proxy (or the prod Traefik stack) for anything internet-facing. Credit stays yours — thanks for surfacing this!

Stanley-blik:
Makes sense, configuration via env is cleaner than a hard override.

PR #257, opened 2026-06-16
feat(swagger): add @ApiSecurity decorator to remaining controllers

aharoldk:
## Description
The Swagger docs at `/api/docs` were unusable — every protected endpoint returned `401 Unauthorized` because the `X-API-Key` header was not being attached to requests. Two controllers (`/infra` and `/auth/validate`) were missing the `@ApiSecurity('X-API-Key')` decorator, which meant Swagger had no way to know these endpoints required authentication. Without it, Swagger does not include the header and all requests from the docs fail.

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] Lint passes
- [X] Self-reviewed

## Related Issues
Closes #

### DIFF
--- a/src/modules/audit/audit.controller.ts
+++ b/src/modules/audit/audit.controller.ts
@@ -1,9 +1,10 @@
 import { Controller, Get, Query } from '@nestjs/common';
-import { ApiTags, ApiOperation, ApiResponse, ApiQuery } from '@nestjs/swagger';
+import { ApiTags, ApiOperation, ApiResponse, ApiQuery, ApiSecurity } from '@nestjs/swagger';
 import { AuditService, AuditQueryOptions } from './audit.service';
 import { AuditLog, AuditAction, AuditSeverity } from './entities/audit-log.entity';
 
 @ApiTags('audit')
+@ApiSecurity('X-API-Key')
 @Controller('audit')
 export class AuditController {
   constructor(private readonly auditService: AuditService) {}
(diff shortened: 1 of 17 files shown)


rmyndharis:
Thanks for the contribution, @aharoldk! Unfortunately this is already handled.

[#109](https://github.com/rmyndharis/OpenWA/pull/109) (`fix(swagger): apply X-API-Key security scheme globally`, merged 2026-06-13) added `.addSecurityRequirements('X-API-Key')` in [`src/config/swagger.config.ts`](https://github.com/rmyndharis/OpenWA/blob/main/src/config/swagger.config.ts), which applies the scheme **globally to every operation** — so Swagger UI already sends the `X-API-Key` header on all protected endpoints. There's a regression test guarding this in [`src/config/swagger.config.spec.ts`](https://github.com/rmyndharis/OpenWA/blob/main/src/config/swagger.config.spec.ts).

Adding `@ApiSecurity('X-API-Key')` per controller just duplicates that global requirement, so on current `main` this is effectively a no-op.

If you were hitting a `401` in `/api/docs`, please make sure you're on the latest `main`, then click **Authorize** in the Swagger UI and enter your API key — every endpoint should work after that.

Closing as already-resolved — but I genuinely appreciate you digging into the API docs, and I hope you'll send more PRs. 🙏

Issue #254, opened 2026-06-16
[Bug]: Error: Failed to launch the browser process


Znuff:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm on the latest released version (or noted my version below)

### OpenWA version

6769690bf5d52298c06e58968d145d2e7e9cb110

### Deployment

Docker Compose

### Database

SQLite (default)

### WhatsApp engine

whatsapp-web.js

### What happened?

```
POST http://10.11.19.66:2886/api/sessions/f9cb86d0-58ea-4a97-a201-3abd48521989/start 500 (Internal Server Error)

Failed to start: Error: Internal server error
    at H (index-BUP2EOcR.js:12:66666)
    at async X (Sessions-CJWTde16.js:1:3579)
X @ Sessions-CJWTde16.js:1
await in X
onClick @ Sessions-CJWTde16.js:1
yd @ index-BUP2EOcR.js:9
(anonymous) @ index-BUP2EOcR.js:9
hn @ index-BUP2EOcR.js:9
wd @ index-BUP2EOcR.js:9
up @ index-BUP2EOcR.js:10
cp @ index-BUP2EOcR.js:10
```

### Steps to reproduce

1. `docker compose --profile with-proxy --profile with-dashboard up --build`
2. log on with the API key
3. create session
4. click on "Start"

### Relevant logs

```shell
openwa-api           | {"timestamp":"2026-06-16T01:22:56.842Z","level":"info","context":"SessionService","message":"Session created: 123","sessionId":"f9cb86d0-58ea-4a97-a201-3abd48521989","action":"create"}
openwa-api           | {"timestamp":"2026-06-16T01:22:58.867Z","level":"info","context":"SessionService","message":"Initializing engine for session: 123","sessionId":"f9cb86d0-58ea-4a97-a201-3abd48521989","action":"engine_init","proxyEnabled":false}
[...]

Issue #251, opened 2026-06-15
[Bug]: v0.2.x sessions hang at "authenticating", never reach "ready" on Raspberry Pi 5 / ARM64; crashpad "--database is required" returns (#242); 0.1.8 works


ulises2k:
## Summary
After upgrading 0.1.8 → 0.2.3, every WhatsApp session reaches `authenticating` but
never transitions to `ready` on a Raspberry Pi 5 (ARM64). The engine authenticates
(the phone shows the device linked) but the whatsapp-web.js sync never completes, so
the session is unusable. Rolling back to 0.1.8 (whatsapp-web.js 1.26.1-alpha.3) fixes
it — the same LocalAuth reaches `ready` in ~3s. The `chrome_crashpad_handler:
--database is required` error from #242 reappears on v0.2.3, though Chromium still launches.

## Environment
- Raspberry Pi 5, ARM64 (aarch64), Debian 12 bookworm, kernel 6.12
- Docker Compose, image built locally from the repo, `--profile full`
- OpenWA v0.2.3, engine whatsapp-web.js 1.34.7 (#222), image Chromium (`/usr/bin/chromium`)
- sqlite + local storage, `NODE_ENV=production`
- Container as shipped: non-root `openwa`, `read_only` rootfs, `HOME=/tmp` (tmpfs)

## Steps to reproduce
1. Deploy v0.2.3 on a Pi 5 (`docker compose --profile full up -d --build`).
2. Create + start a session (or AUTO_START a previously authenticated one).
3. Scan the QR — the phone links the device.
4. Status goes `initializing → authenticating` and stays there indefinitely; `ready`
   never fires. `GET /sessions/:id/qr` returns 400 while stuck.

## Actual
Status stuck at `authenticating`, `lastError: null`, no `ready` in logs. First launch
attempt logs (note crashpad, #242):

```
Failed to launch the browser process: Code: 21
chrome_crashpad_handler: --database is required
[ERROR ...filesystem_posix.cc:63] mkdir : No such file or directory (2)
[ERROR ...process_singleton_posix.cc:365] The profile appears to be in use by another
Chromium process (...) on another computer (<old-container-hostname>) ...
```

(The SingletonLock part is a separate recreate-hostname artifact; clearing
`session-*/Singleton*` lets Chromium launch, then it reaches `authenticating` but still
never `ready`.)

## Expected
Session reaches `ready` after scanning, as on 0.1.8.

## Tried (no effect on the hang)
[...]

Issue #242, opened 2026-06-15
v0.2.x: headless Chromium fails to launch as non-root user — "chrome_crashpad_handler: --database is required"


rcitelli:
### Summary

Since **v0.2.0**, the WhatsApp engine fails to start in the official Docker image. The session goes to `failed` and Puppeteer cannot launch the browser:

```
Session engine failed: Failed to launch the browser process:
stderr:
chrome_crashpad_handler: --database is required
Try 'chrome_crashpad_handler --help' for more information.
```

The HTTP API and `/api/health` respond fine — only the browser/session is broken — so the container looks "healthy" while the gateway is actually unusable for sending/receiving messages on a fresh v0.2.0 / v0.2.1 deploy.

### Environment

- OpenWA **v0.2.1** (also reproduces on v0.2.0), official `Dockerfile` / `docker compose up`
- Core profile: SQLite, `whatsapp-web.js` engine, `PLUGINS_ENABLED=true`
- System Chromium **149.0.7827.114** (`PUPPETEER_EXECUTABLE_PATH=/usr/bin/chromium`, `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true`)
- Host: Debian 12, Docker; container reports healthy

### Root cause

Two changes between v0.1.8 and v0.2.x combine to break Chromium:

1. **#229 (container hardening, v0.2.0):** the container now drops to a **non-root `openwa` user** via `gosu`. The user is created with `useradd -r -g openwa openwa` in the `Dockerfile` — **no `-m`, so it has no writable `$HOME`** (defaults to `/`).
2. **`whatsapp-web.js` bumped `^1.26.1-alpha.3` → `^1.34.7`**, which drives a much newer bundled Chromium (149).

Newer Chromium spawns `chrome_crashpad_handler`, which needs a **writable crash-dump database directory** (derived from `$HOME` / the user-data-dir). With a non-root user that has **no writable home**, crashpad receives no `--database` and the browser process dies. In v0.1.x this worked because the process ran as **root** (`HOME=/root`, writable).

### Minimal reproduction (isolated — no WhatsApp involved)

Inside the running container, launch Chromium directly as the `openwa` user:

```sh
# FAILS — non-writable HOME, no user-data-dir:
docker exec -u openwa -e HOME=/nonexistent openwa-api \
  /usr/bin/chromium --headless=new --no-sandbox --dump-dom about:blank
# -> chrome_crashpad_handler: --database is required   (core dumped)

[...]

PR #259, opened 2026-06-16
fix: Chromium singleton locks, crashpad --database, auth hang, and 502 Bad Gateway

Abhishekrajpurohit:
## Summary

Fixes 4 runtime issues that block session start on a clean Docker setup:

### 1. `Dockerfile` — sqlite3 native module build failure
`npm ci --omit=dev` cannot build sqlite3's native addon (needs dev toolchain). Fixed by copying the complete `node_modules` from the builder stage, which has all native modules pre-built.

### 2. `dashboard/nginx.conf` — 502 Bad Gateway
Dashboard nginx proxied to `http://openwa:2785` but the container is named `openwa-api`. Fixed in both the `/api/` and `/socket.io/` location blocks.

### 3. `docker-compose.dev.yml` — Chromium auth hang + crashpad
- Added `WWEBJS_WEB_VERSION=2.3000.1023204257` (known-good WA Web version, confirmed in #251) to prevent session hanging at \`authenticating\` indefinitely
- Added `XDG_CONFIG_HOME=/tmp/.chromium` for explicit Chromium config path
- Added `--no-zygote` to `PUPPETEER_ARGS` to prevent zygote process from interfering with crashpad hooks

### 4. `docker-entrypoint.sh` — Chromium singleton lock files + crashpad handler
Chromium doesn't clean up `SingletonLock`, `SingletonSocket`, or `SingletonCookie` files on crash. These stale locks block subsequent launches (Code 21: "profile appears to be in use by another Chromium process"). Fixed by wiping all lock files and `/tmp/.chromium` on every container start, and creating `/tmp/.config/Crashpad` so the crashpad handler can initialise without throwing \`--database is required\` (refs #254, #242).

## Testing
- Session start: \`initializing → authenticating → ready\` completes in ~15s
- No crashpad \`--database is required\` errors
- No singleton lock errors after container restart
- Dashboard API proxy: 200 OK on all endpoints

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,8 +65,8 @@ WORKDIR /app
 # Copy package files
 COPY package*.json ./
 
-# Install production dependencies only
-RUN npm ci --omit=dev && npm cache clean --force
+# Copy node_modules from builder (includes native modules like sqlite3 built with full toolchain)
+COPY --from=builder /app/node_modules ./node_modules
 
 # Copy built application from builder stage
 COPY --from=builder /app/dist ./dist
--- a/dashboard/nginx.conf
+++ b/dashboard/nginx.conf
@@ -15,7 +15,7 @@ server {
 
     # Proxy API requests to backend
     location /api/ {
-        proxy_pass http://openwa:2785;
+        proxy_pass http://openwa-api:2785;
         proxy_http_version 1.1;
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection 'upgrade';
@@ -25,7 +25,7 @@ server {
 
     # WebSocket proxy
     location /socket.io/ {
-        proxy_pass http://openwa:2785;
+        proxy_pass http://openwa-api:2785;
         proxy_http_version 1.1;
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection 'upgrade';
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -38,13 +38,15 @@ services:
       - NODE_ENV=development
       - PORT=2785
       - HOME=/tmp
+      - XDG_CONFIG_HOME=/tmp/.chromium
+      - WWEBJS_WEB_VERSION=2.3000.1023204257
       - DATABASE_TYPE=sqlite
       - DATABASE_NAME=/app/data/openwa.sqlite
       - DATABASE_SYNCHRONIZE=true
       - ENGINE_TYPE=whatsapp-web.js
       - SESSION_DATA_PATH=/app/data/sessions
       - PUPPETEER_HEADLESS=true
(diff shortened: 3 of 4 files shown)


rmyndharis:
Thanks for digging into the Docker runtime issues, @Abhishekrajpurohit — there's real signal in here. I'm going to close this PR as-is, but two of your findings are good and **I'm implementing them directly** (with credit to you in the changelog), so you don't need to redo them:

- ✅ **nginx hostname** — `dashboard/nginx.conf` proxied to a bare `openwa`, but the canonical backend service is **`openwa-api`** (the production compose service name, the dev `container_name`, and what `dashboard/Dockerfile.traefik` already uses). Aligning both the `/api/` and `/socket.io/` blocks to `openwa-api`.
- ✅ **Chromium singleton-lock cleanup** — clearing stale `SingletonLock`/`Socket`/`Cookie` in session profiles on start (the "profile appears to be in use" / Code 21 issue after an unclean shutdown) is a genuine fix. Adding a clean version to the entrypoint.

The reasons I can't merge the PR itself:

1. **Accidental submodule commit (hard blocker).** The diff adds a git **gitlink** (`mode 160000`) named `OpenWA` pointing at commit `b90b4d13` — which is OpenWA's own **v0.2.6** release commit. A nested clone got `git add`ed by mistake; merging it would create a broken submodule in the repo root.
2. **Duplicate `XDG_CONFIG_HOME` from a merge slip.** After merging `main`, `docker-compose.dev.yml` ends up with two `XDG_CONFIG_HOME` entries (`/tmp/.config` and `/tmp/.chromium`); YAML last-wins picks `/tmp/.chromium`, which the entrypoint then `rm -rf`s and never recreates — self-contradictory.
3. **Dockerfile would ship dev dependencies to production (independent reason to decline).** Replacing `npm ci --omit=dev` with `COPY --from=builder /app/node_modules` pulls the full builder toolchain (TypeScript, jest, eslint, …) into the `node:22-slim` runtime image — image bloat + attack surface. Note the Docker image build is **push-gated** in CI (it doesn't run on PRs), so this wouldn't be caught automatically. `sqlite3` ships prebuilt binaries, so `npm ci --omit=dev` already builds fine on slim — that's how the released image is produced; there's no native-build failure to work around this way.
4. **The crashpad / `--database is required` part is already fixed** on `main` in **v0.2.6** (#254), using a verified `XDG_CONFIG_HOME=/tmp/.config` + `/tmp/.cache` approach (the entrypoint pre-creates them). `--no-zygote` was intentionally left out of the compose `PUPPETEER_ARGS` override, since the reporter confirmed the launch succeeds without it.

So even setting aside the accidental submodule, the Dockerfile change alone would block a merge. A good pattern for next time is one small, focused PR per concrete fix against current `main`. But since the nginx + lock-cleanup pieces are quick and unblock people now, I'm taking those directly. Really appreciate the investigation — please keep them coming. 🙏

Issue #265, opened 2026-06-16
Engine pluggability: decouple whatsapp-web.js specifics from the engine-neutral layer


rmyndharis:
## Context

OpenWA has a pluggable engine architecture — the seam is `IWhatsAppEngine` (neutral types), and **only the adapter layer** (`src/engine/adapters/*`, `src/plugins/engines/whatsapp-web-js/*`) should know whatsapp-web.js. A code audit found several places **outside** that seam that assume whatsapp-web.js specifics and would break (or silently misbehave) under a different engine (e.g. Baileys, which uses `@s.whatsapp.net` JIDs and a different ack/presence/type model).

PR #264 already addressed the lightweight items (engine-neutral typing via `sendChatState`, `mark-chat-read` JID validation relaxed, real engine library version surfaced). This issue tracks the remaining, larger decouplings — each is its own PR to keep regression surface contained.

### 🔴 High — delivery ack scale leaks whatsapp-web.js integers end-to-end
The raw wwebjs `MessageAck` integer (`-1..4`) flows through the neutral interface (`MessageResult.ack`, `onMessageAck(id, ack: number)`), is interpreted in `message-status.util.ts` (`ackToMessageStatus`) and `session.service.ts` (`ack < 0` → `message.failed`), embedded in webhook payloads + idempotency keys, re-emitted by `events.gateway.ts`, and re-mapped **again** in the dashboard (`Chats.tsx` `statusMap`, `useWebSocket.ts`). Baileys uses a different/offset integer scale → every delivery/read state would mis-map.
- **Fix:** introduce a neutral `DeliveryStatus` enum (`pending|sent|delivered|read|played|failed`) at the seam; each adapter maps its native codes; services/webhook/ws/dashboard consume the neutral string.

### 🔴 High — `IncomingMessage.type` is the raw wwebjs `MessageTypes` string
`message-mapper.ts` assigns `type: msg.type` (`chat`/`ptt`/`voice`/`revoked`/…) directly; the dashboard switches on those tokens (`Chats.tsx` media render, `'ptt'`/`'voice'`, and a latent `'chat'` vs `'text'` mismatch to verify).
- **Fix:** normalize to a neutral `MessageType` union/enum at the adapter boundary (`'chat' → 'text'`, `'ptt'/'voice' → 'voice'`, …); consumers switch on the neutral type.

### 🟠 Medium — JID construction outside the adapter
- `contact.controller.ts` builds `${number}@c.us` in the check-number response.
[...]

PR #268, opened 2026-06-16
refactor(engine): neutral DeliveryStatus instead of raw whatsapp-web.js ack int (#265)

rmyndharis:
First item of the **#265** engine-pluggability decoupling: stop the whatsapp-web.js `MessageAck` integer from leaking past the engine adapter.

## What changed
- **Neutral `DeliveryStatus`** (`pending`|`sent`|`delivered`|`read`|`failed`) added to `IWhatsAppEngine`; `onMessageAck` now delivers a `DeliveryStatus`, not a raw int.
- The wwebjs ack integer is mapped to `DeliveryStatus` **only inside the adapter** (`wwebjsAckToDeliveryStatus`). No consumer outside the adapter sees engine ack codes anymore.
- `message-status.util`: `ackToMessageStatus(int)` → `deliveryStatusToMessageStatus(status)`; added deprecated `deliveryStatusToAck(status)` to derive the legacy webhook `ack` int. **Behaviour is byte-identical for every ack value** (verified end-to-end).
- **Webhook (non-breaking):** `message.ack`/`message.failed` now carry a neutral **`status`**; the legacy **`ack`** int is kept (deprecated). Idempotency keys prefer `status`.
- **Live ws delivery ticks (bonus fix):** `session.service` now actually emits `message.ack` over the websocket — `emitMessageAck` was previously **never called**, so the dashboard's live tick path was dead. The dashboard consumes the neutral status string; the integer `statusMap {-1..4}` is removed.

A Baileys/other engine adapter now only needs to map its native delivery codes to `DeliveryStatus`.

## Decisions (confirmed with maintainer)
- Webhook kept **non-breaking** (add `status`, keep deprecated `ack`).
- The dead ws delivery-tick path **wired live** with the neutral status.
- `PLAYED(4)` collapses to `read` (the DB already did this). Documented deltas: the deprecated webhook `ack` reports `3` not `4` for a played receipt, and play-after-read no longer double-fires `message.ack`.

## Verification
- Backend build ✅ · **445 tests** ✅ (added a `wwebjsAckToDeliveryStatus` boundary test regression-locking the int→status map incl. PLAYED→read) · lint ✅ · dashboard build + lint ✅.
- Adversarial review (3 dimensions, each independently verified): all **hold** — exact behavior preservation, type-sound live ws path, no remaining raw-ack leak outside the adapter.

Part of #265.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- **Message delivery status is now engine-agnostic** (engine-pluggability decoupling, #265). The raw whatsapp-web.js
+  ack integer no longer leaks past the engine adapter — a neutral `DeliveryStatus`
+  (`pending`/`sent`/`delivered`/`read`/`failed`) flows through the interface, services, webhooks, websocket, and
+  dashboard, so a non-whatsapp-web.js engine (e.g. Baileys) can map its own delivery codes at the adapter boundary.
+  - The `message.ack`/`message.failed` webhooks now include a neutral **`status`** field. The legacy **`ack`** integer
+    is **kept (deprecated)** for backward compatibility — new consumers should read `status`.
+  - Dashboard chat delivery ticks now update **live** over the websocket (the ack push was previously never emitted).
+  - Minor deprecated-surface deltas: the legacy webhook `ack` reports `3` (not `4`) for a "played" voice/video receipt,
+    and a play-after-read no longer emits a second `message.ack` (both map to `status: 'read'`).
+
 ## [0.2.7] - 2026-06-16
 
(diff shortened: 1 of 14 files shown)


rmyndharis:
### Post-review hardening (deep self-review of this PR)

Adversarial review (source-traced) flagged one real gap, now fixed in 4af6aca:

- **Dashboard tick downgrade.** `emitMessageAck` now fires on *every* ack (incl. `pending`/`sent`), and engine acks can arrive out of order / be replayed on reconnect. The dashboard applied `event.status` unconditionally while the backend DB update is forward-only — so a late/duplicate lower ack could visually revert a `read` tick to `delivered`/`sent`. Added `mergeDeliveryStatus` that mirrors the backend's `ackStatusTransitionFrom` **exactly** (rank advance; `failed` only from pending/sent and terminal — a late failure must not clobber a confirmed delivered/read).
- **Docs accuracy.** `docs/06` `message.ack` example showed a never-emitted `ackName` and a wrong ack-code table; corrected to the real `{id, messageId, status, ack}` payload, documented `status` as canonical / `ack` deprecated, and fixed the idempotency-key examples.

Confirmed non-issues: stored-status mapping is byte-identical to `main`; the PLAYED(4)→read collapse is intentional/documented; the idempotency key isn't stored internally (consumer-side header) so no migration concern.

PR #270, opened 2026-06-16
refactor(engine): neutral MessageType instead of raw whatsapp-web.js type tokens (#265)

rmyndharis:
Second item of the **#265** engine-pluggability decoupling: stop raw whatsapp-web.js `MessageTypes` tokens leaking past the engine adapter.

## What changed
- **Neutral `MessageType`** union (`text|image|video|audio|voice|document|sticker|location|contact|revoked|unknown`) on `IncomingMessage.type`.
- `message-mapper`: new **`mapWwebjsMessageType`** (`chat`→`text`, `ptt`→`voice`, `vcard`/`multi_vcard`→`contact`, others passthrough/`unknown`) is the **sole** translation point. So incoming live + `getChatHistory` + persisted rows + the `message.received`/`message.sent` webhooks are now neutral — **consistent with outgoing** (which `message.service` already wrote as neutral).
- **Migration `NeutralizeMessageTypes`** backfills existing `messages.type` rows. This matters because the dashboard chat view reads **persisted rows** (`GET /messages`), and stats group by `type` — without it, old `chat`/`ptt`/`vcard` rows would mis-render and split the by-type stats. Collision-free, idempotent, portable (SQLite/Postgres).
- **Dashboard:** removed the dead wwebjs `'ptt'` case from `renderMedia`; the **latent chat-vs-text bug** (incoming text mis-styled as media + shown as `[chat]` in reply previews) is fixed.
- Docs + CHANGELOG updated.

## Decisions (confirmed)
Neutralize `type` **everywhere incl. the webhook** — a contract change toward consistency. The adversarial review corrected my initial assumption twice (both load-bearing, both fixed here):
1. The dashboard reads **persisted** rows (`getMessages`), *not* engine `getChatHistory` → hence the **backfill migration** (without it, old voice notes would regress to document links).
2. The webhook change affects **both** `message.received` *and* `message.sent` (the sent webhook is dispatched from `buildIncomingMessageBase`, not the neutral DB write).

**Webhook contract change for integrators:** incoming/history `type` is now neutral — `chat`→`text`, `ptt`→`voice`, `vcard`→`contact`.

## Verification
- Backend build ✅ · **449 tests** ✅ (added a `mapWwebjsMessageType` boundary test) · lint ✅ · dashboard build + lint ✅.
- Migration SQL verified against an in-memory SQLite DB (legacy tokens mapped, neutral/passthrough untouched).
- Adversarial review (3 dimensions, independently verified): all required changes (backfill migration, docs, dual-direction webhook note) incorporated.

Part of #265.


phendricks7-Pisces (approved):

rmyndharis:
### Post-review hardening (deep self-review of this PR)

Adversarial review caught one real gap, now fixed in d8bc119:

- **The backfill migration never ran in the default SQLite config.** `synchronize: true` (the zero-config default) ⇒ `migrationsRun: false`, so `NeutralizeMessageTypes` only ran on Postgres / `DATABASE_SYNCHRONIZE=false`. An upgraded default-SQLite install kept old `chat`/`ptt`/`vcard` rows — and because this PR removes the dashboard's `case 'ptt'`, old voice notes regressed to document bubbles.
- **Fix:** replaced the migration with an idempotent `MessageTypeBackfillService` that runs on app bootstrap in **every** DB mode (chat→text, ptt→voice, vcard/multi_vcard→contact). Forward-only + collision-free ⇒ no-op once converted. +3 tests.

Confirmed solid: `mapWwebjsMessageType` covers the full wwebjs enum with a safe `unknown` fallback; the mapper is the sole boundary across live/history/persisted paths; webhook `type` contract change (chat→text, etc.) is documented — **needs a minor version bump at release**.

PR #295, opened 2026-06-17
docs: align Node.js runtime version

hsnyvsh:
## Summary

- Update the documentation tech stack to consistently reference Node.js 22 LTS.
- Add a package `engines.node` requirement so the runtime expectation is visible to package managers and contributors.

## Context

The documentation badge already references Node.js 22 LTS, while the Tech Stack table still referenced Node.js 20 LTS. This PR aligns those references and makes the runtime requirement explicit in `package.json`.

## Testing

Not run. Documentation/package metadata only.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -188,7 +188,7 @@ socket.on('message', msg => {
 
 | Layer     | Technology                    |
 | --------- | ----------------------------- |
-| Runtime   | Node.js 20 LTS                |
+| Runtime   | Node.js 22 LTS                |
 | Framework | NestJS 11.x                   |
 | Language  | TypeScript 5.x                |
 | WA Engine | whatsapp-web.js               |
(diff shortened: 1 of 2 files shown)


rmyndharis:
Thanks for catching the docs drift. Closing this one as superseded by the v0.3.0 engine work that just landed.

As of v0.3.0, OpenWA intentionally **does not declare a global Node version floor**. The Baileys engine (which is ESM-only) is now lazy-loaded via dynamic `import()` only when it's actually selected (`ENGINE_TYPE=baileys`), so the default whatsapp-web.js engine runs on any supported Node and we removed `engines.node` from `package.json` on purpose. This PR re-introduces a `>=22` floor that contradicts that direction (and now conflicts on `package.json`). The runtime-version guidance is covered in the updated CHANGELOG/README. Appreciate the attention to the docs!

hsnyvsh:
Thanks for the clarification. That makes sense with the v0.3.0 engine split and the package-specific dependency handling.

I’ll take a look at the updated README/CHANGELOG before opening another docs-focused PR.

PR #305, opened 2026-06-17
Quality: Path Traversal Vulnerability in isPathWithin

tuanaiseo:
## Problem

The `isPathWithin` utility in `path-safety.ts` uses `path.resolve(resolvedRoot, target)` which follows symlinks. An attacker could bypass the path check by providing a symlink that points outside the root directory. Additionally, the function does not handle cases where `root` contains trailing spaces or special characters that might affect path resolution on certain platforms.

**Severity**: `critical`
**File**: `src/common/utils/path-safety.ts`

## Solution

Use `fs.realpath` to resolve symlinks before comparison, and normalize the root path. Consider using a library like `resolve-path` or implementing a more robust check that verifies the resolved target is actually within the root directory after symlink resolution.

## Changes

- `src/common/utils/path-safety.ts` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/src/common/utils/path-safety.ts
+++ b/src/common/utils/path-safety.ts
@@ -1,3 +1,4 @@
+import * as fs from 'fs';
 import * as path from 'path';
 
 /**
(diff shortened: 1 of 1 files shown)


rmyndharis:
Thank you for the security focus here — symlink/traversal hardening is the right instinct. After review, though, this change would introduce a regression rather than fix one, so I'm going to close it.

`isPathWithin` is deliberately pure string math and always returns a boolean. Swapping in `fs.realpathSync(...)` makes it **throw `ENOENT` on any path that doesn't exist yet** — and we call `isPathWithin` *before* creating files (e.g. `StorageService.putLocalFile` validates the target path before writing it). With this change, every local media write would throw `ENOENT` instead of returning a boolean, and the existing `path-safety.spec.ts` cases (over a non-existent root) would throw rather than assert.

If you'd like to pursue symlink-resolution hardening, it needs to be `ENOENT`-safe (resolve only the existing ancestor directory, not the full target) and come with tests. Happy to review a revised version. Thanks again for contributing.

PR #314, opened 2026-06-18
docs: add n8n appointment booking workflow example

hsnyvsh:
## Summary

- Add a complete n8n appointment booking workflow example for OpenWA.
- Link the example from the existing n8n integration documentation.
- Cover trigger setup, booking intent detection, request normalization, availability checks, confirmation messages, alternative slots, and operational notes.

## Context

The existing n8n documentation includes short examples for auto-replies, lead collection, session monitoring, order notifications, and scheduled reminders. This PR adds a more practical end-to-end appointment booking example, which is a common WhatsApp automation use case.

## Testing

Not run. Documentation-only change.

### DIFF
--- a/docs/22-n8n-integration.md
+++ b/docs/22-n8n-integration.md
@@ -172,6 +172,24 @@ Send daily reminders to a list of contacts.
      └── Daily 9AM          └── Get contacts                     └── Send reminder
 ```
 
+### 6. Appointment Booking
+
+Collect appointment requests over WhatsApp, check availability in an external scheduling source, and send a confirmation or alternative time slots.
+
+See [n8n Appointment Booking Workflow](./examples/n8n-appointment-booking.md) for a complete example.
+
+```
+[OpenWA Trigger] → [IF: Booking intent?] → [Set: Normalize request]
+                                               │
+                                               ▼
+                                      [Availability Source]
+                                               │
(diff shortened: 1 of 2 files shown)


rmyndharis:
Merged — thanks for this, @hsnyvsh! 🙏 The appointment-booking example is well-structured and engine-neutral, and the field references (`data.body`, `data.chatId`, `data.from`) all map to the real `message.received` payload.

One small follow-up I pushed on top in [`c67cd2b`](https://github.com/rmyndharis/OpenWA/commit/c67cd2b8ed1f50aa3b9b3262c682bae4b427735f): the Operational Note suggested keying idempotency on `messageId`, but the dispatched `message.received` payload is a spread `IncomingMessage` whose message-id field is actually **`id`** (`messageId` is only tolerated as a legacy fallback in `idempotency.util.ts`). I corrected it in both the new example and the pre-existing "Output Data Format" block (which already had the same `messageId` typo), and restored the trailing newline on `docs/22-n8n-integration.md`. Nothing for you to do — just flagging it for transparency. Thanks again!

hsnyvsh:
Thanks for the quick merge and the correction. Noted on `message.id` being the canonical field for idempotency; I’ll use that convention in future docs/examples.

Issue #297, opened 2026-06-17
[Bug]: Creating of sessions stucked at "authenticating" state/status after scanning qr code


piusshedrach:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm on the latest released version (or noted my version below)

### OpenWA version

0.2.10

### Deployment

Docker Compose

### Database

SQLite (default)

### WhatsApp engine

whatsapp-web.js

### What happened?

After starting a session and scanning the qr code the status is stucked at authenticating with my device showing "last active at -  [time is just scanned]".
I tried reconnecting but it wont resolve it.

### Steps to reproduce

- clone repo,
- install deps via sudo docker comopse -f docker-compose.dev.yml up
- create a session
- start session
- scan qr code

### Relevant logs

```shell

```

gp381:
+1

gp381:
in side OpenWA folder, create an empty `.env` file, and put the following config
```
WWEBJS_WEB_VERSION=2.3000.1023204257
```

then run this command to recreate containers `docker compose up -d --force-recreate`

it worked for me

Janeirohurley:
`[+] up 12/12king to docker.io/library/openwa-openwa-api:latest                                                                                                                                                                           0.1s
 ✔ Image tecnativa/docker-socket-proxy Pulled                                                                                                                                                                                            33.1s
 ✔ Image openwa-openwa-api             Built                                                                                                                                                                                             35.0s
network openwa-network was found but has incorrect label com.docker.compose.network set to "default" (expected: "openwa-network")
janeirohurley@janeirohurley /h/p/w/OpenWA (main) [1]> 
`
me also i have the same bug the sasion stay in connecting 

<img width="762" height="422" alt="Image" src="https://github.com/user-attachments/assets/3a8abfc3-6388-4825-b2cc-48412f4755ae" />

rmyndharis:
Thanks for the report — and thanks @gp381 for posting the workaround above.

This is the same root cause as #251 and #273: on `whatsapp-web.js` 1.34.x, the **auto-selected WhatsApp Web version** is sometimes incompatible, so the post-link sync never completes and the session hangs at `authenticating` (your phone shows it as linked, but Web never finishes syncing). It isn't specific to your setup — it's the upstream auto-version behavior, and 1.34.7 is currently the latest `whatsapp-web.js` release, so pinning a known-good version is the recommended remedy.

[...]

Issue #273, opened 2026-06-16
[Bug]: Session stuck at "Authenticating" after scanning QR (v0.2.7)


JasonSecurity:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm on the latest released version (or noted my version below)

### OpenWA version

0.2.7

### Deployment

Docker Compose

### Database

SQLite (default)

### WhatsApp engine

whatsapp-web.js

### What happened?

After scanning the QR code, the session gets stuck indefinitely in the "Authenticating" state in the Dashboard.
The QR code is generated and displayed correctly, and it can be scanned successfully with the WhatsApp iOS mobile app. However, the session never transitions to the authenticated/ready state.

<img width="408" height="282" alt="Image" src="https://github.com/user-attachments/assets/0edab847-31e0-47be-90df-84a9ad16ee62" />

### Steps to reproduce

1. Fresh install / update to latest version (v0.2.7)
2. Start a new session via Dashboard
3. Scan the QR code with WhatsApp mobile
4. Dashboard shows "Authenticating" indefinitely
5. No further progress / Backend logs stop after "QR code generated"

### Expected behavior

After scanning the QR, the session should transition to authenticated → ready state.

### Relevant logs

```shell
openwa-api  | {"timestamp":"2026-06-16T21:21:35.757Z","level":"info","context":"PluginLoaderService","message":"[whatsapp-web.js] WhatsApp-web.js engine plugin loaded","pluginId":"whatsapp-web.js"}
openwa-api  | {"timestamp":"2026-06-16T21:21:35.757Z","level":"info","context":"PluginLoaderService","message":"[whatsapp-web.js] WhatsApp-web.js engine plugin enabled","pluginId":"whatsapp-web.js"}
openwa-api  | {"timestamp":"2026-06-16T21:21:35.757Z","level":"info","context":"PluginLoaderService","message":"Plugin enabled: WhatsApp Web.js Engine","pluginId":"whatsapp-web.js","action":"plugin_enabled"}
openwa-api  | {"timestamp":"2026-06-16T21:21:35.757Z","level":"info","context":"EngineFactory","message":"Engine plugin enabled: whatsapp-web.js","action":"engine_enabled","engineType":"whatsapp-web.js"}
openwa-api  | [Nest] 7  - 06/16/2026, 9:21:35 PM    WARN [DockerService] Docker not available. Container orchestration disabled.
[...]

PR #316, opened 2026-06-18
fix: add force-kill endpoint to terminate orphaned Chromium processes

Abhishekrajpurohit:
## Summary
- Adds `POST /api/sessions/:id/force-kill` — kills orphaned Chromium processes, resets session to `disconnected` without wiping credentials or DB record
- Only authenticated sessions can trigger force-kill
- Closes session gateway connection before `pkill` to avoid stale auth state
- Logs action as `force_kill` for audit trail

## Files
- `src/modules/session/session.controller.ts` — new endpoint
- `src/modules/session/session.service.ts` — `forceKillSession()` implementation
- `docker-entrypoint.sh` — ensures Chromium profile cleanup on restart

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -66,7 +66,8 @@ PUPPETEER_ARGS=--no-sandbox,--disable-setuid-sandbox,--disable-dev-shm-usage,--d
 # WWEBJS_WEB_VERSION=2.3000.1023204257
 
 # Baileys engine (used when ENGINE_TYPE=baileys). WebSocket client, no Chromium.
-# NOTE: the Baileys engine dependency is ESM-only; OpenWA therefore requires Node >= 20.19 (require-of-ESM support).
+# NOTE: the Baileys engine is loaded lazily (dynamic import, only when ENGINE_TYPE=baileys), so there is no
+# global Node.js version floor. Node.js 22 LTS is recommended for all deployments.
 BAILEYS_AUTH_DIR=./data/baileys
 # NOTE: proxy (PROXY_URL/PROXY_TYPE) is not yet supported by the baileys engine; it is ignored.
 
--- a/docs/01-project-overview.md
+++ b/docs/01-project-overview.md
@@ -220,7 +220,7 @@ quadrantChart
 ```mermaid
 flowchart TB
     subgraph Runtime["Runtime & Framework"]
-        N[Node.js 20 LTS] --> |"TypeScript support, async/await"| NE[NestJS]
+        N[Node.js 22 LTS] --> |"TypeScript support, async/await"| NE[NestJS]
         NE --> |"Modular, scalable, DI"| API[REST API]
     end
     
@@ -247,7 +247,7 @@ flowchart TB
 
 | Layer | Technology | Rationale |
 |-------|------------|-----------|
-| Runtime | Node.js 20 LTS | Stable, long-term support |
+| Runtime | Node.js 22 LTS | Stable, long-term support |
 | Language | TypeScript | Type safety, better developer experience |
 | Framework | NestJS | Enterprise-grade, modular |
 | WA Engine | whatsapp-web.js (default) or baileys | Pluggable via `ENGINE_TYPE` env var; wwebjs = Puppeteer/stealth, baileys = browser-free |
--- a/docs/02-requirements-specification.md
+++ b/docs/02-requirements-specification.md
@@ -365,7 +365,7 @@ flowchart TB
 
 | ID | Requirement | Description | Priority |
 |----|-------------|-------------|----------|
-| NFR-CP-001 | Node.js version | 18.x, 20.x LTS | High |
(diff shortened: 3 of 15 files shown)


rmyndharis:
Thanks for tackling this, @Abhishekrajpurohit — stuck sessions are a real and recurring pain (#297, #273, #251), so the motivation is solid. Before iterating on the implementation, though, I'd like to pin down the **root cause**, because I think the diagnosis here may be broader than the actual failure mode.

**Where I agree the problem is real:** whatsapp-web.js / Puppeteer teardown genuinely isn't reliable. The codebase already reflects this — `disconnect()` wraps `client.destroy()` in a try/catch that warns `"Destroy client failed"` ([whatsapp-web-js.adapter.ts L394-L406](https://github.com/rmyndharis/OpenWA/blob/5dd8c28e9b38ad9fed6ebed0421ecf96871ff826/src/engine/adapters/whatsapp-web-js.adapter.ts#L394-L406)), `onModuleDestroy` puts a 10s timeout on `engine.destroy()`, and the Docker entrypoint already clears stale `Singleton*` locks on every start (#259). So an orphaned Chromium after a *failed/hung* `destroy()` is a genuine (if narrow) edge case.

**Where I think the diagnosis is off:**
- In this project, the most common cause of "stuck at **authenticating**" is **not** an orphaned Chromium — it's the whatsapp-web.js **web-version sync hang**. The documented fix is pinning `WWEBJS_WEB_VERSION` (see the note in `.env.example`), and that's how #297 (the issue this PR cites) was triaged. A force-kill won't help that case.
- The **container-restart** scenario the entrypoint `pkill -x chromium` targets is mostly already covered: a container restart kills every process (so there's no surviving Chromium to kill), and the existing entrypoint already clears the stale lock files. So that line is largely a no-op in the path it's aimed at.
- Under **normal** operation there's no orphan: `stop()` → `disconnect()` → `client.destroy()` closes the browser. Orphans only appear when `destroy()` throws/hangs, or the Node process is hard-killed (SIGKILL/OOM/crash) — i.e. the failure path, not the happy path.
- This is also **whatsapp-web.js-specific**: the Baileys engine has no Chromium at all, so the whole class of problem doesn't exist there.

**So my ask:** could you confirm what you're actually hitting? Specifically:
1. Is there a surviving `chromium` process after a stuck session (`ps aux | grep chromium`), holding the profile dir — i.e. a true in-process orphan?
2. Do you see the `onQRChangedEvent already exists` error on re-init (the narrow orphan path), or is the session simply stuck at *authenticating* (more likely the version-sync hang → try `WWEBJS_WEB_VERSION`)?

If it's confirmed to be an in-process orphaned Chromium that survives a failed teardown while the gateway keeps running, then there *is* a real gap worth a **targeted** fix — but it's a narrower problem than a global kill, and it should live behind an engine method in the wwebjs adapter (it's a no-op for Baileys). I also have separate implementation-level notes on the current code (the engine-neutral boundary, the `pkill -f "session-<name>"` substring match that can hit sibling sessions like `test`/`test2`, and the red Lint) that I'm happy to lay out once we've agreed on the actual problem we're solving. 🙏

Abhishekrajpurohit:
Thanks for the detailed review, @rmyndharis — fair points, and you're right that a lot of this PR's motivation came from triaging #297, which turned out to be the WWEBJS_WEB_VERSION issue (not a Chromium orphan). Let me address your questions directly:
    
    What we were actually hitting:
    We encountered the issue when the Docker container was rebuilt but the Node process inside was hard-killed (SIGKILL / OOM / crash) — specifically during development iterations where the container was repeatedly stopped/started without a graceful shutdown. The WhatsApp session would be in a stuck state, and on restart we'd see Chromium binding errors ("onQRChangedEvent already exists"). We confirmed orphaned chromium processes with ps aux | grep chromium — they were surviving inside the container while the gateway was still running.
    
    Why we also hit it under normal operation:
    You're right that a container restart kills all processes cleanly. But during development (and potentially under OOM/recovery scenarios in production), the pkill -x chromium on startup acts as a belt-and-suspenders cleanup.
    
    On the diagnosis being too broad:
    Agreed. The entrypoint pkill -x chromium is largely redundant for container restarts, and the web-version hang (WWEBJS_WEB_VERSION) is the real cause of "stuck at authenticating" — which this PR doesn't address.
    
    On the implementation being engine-neutral:
    Fair critique. pkill -f "session-<name>" is WhatsApp-web.js-specific, and the substring match is indeed too loose (test vs test2). A better approach would be to add a killOrphanedBrowser() method to the IWhatsAppEngine interface (no-op for Baileys, pkill with a tighter pattern for whatsapp-web.js) and call that from the force-kill logic.
    
    On next steps:
    Would you prefer we:
    1. Scope this down to just the IWhatsAppEngine.killOrphanedBrowser() approach (engine-specific, safer pattern matching), or
    2. Close this PR and address it in a future issue specifically for the whatsapp-web.js orphan scenario with a targeted engine adapter fix

PR #317, opened 2026-06-18
feat(dashboard): add Kill Stuck button to session cards

Abhishekrajpurohit:
## Summary
- Adds `POST /api/sessions/:id/force-kill` wired to a **Kill Stuck** button (Skull icon) on session cards with `failed` status
- `handleForceKill()` updates session status and clears `lastError` on success
- Button only visible when session status is `'failed'` and user has write permission
- All three locales updated (en, zh-CN, he): button label + success/error toasts

## Files
- `dashboard/src/services/api.ts` — `forceKill` API call
- `dashboard/src/pages/Sessions.tsx` — `Skull` icon, `handleForceKill()`, Kill Stuck button
- `dashboard/src/i18n/locales/{en,zh-CN,he}.json` — i18n strings

### DIFF
--- a/dashboard/src/i18n/locales/ar.json
+++ b/dashboard/src/i18n/locales/ar.json
@@ -226,7 +226,8 @@
       "start": "بدء",
       "stop": "إيقاف",
       "reconnect": "إعادة الاتصال",
-      "delete": "حذف"
+      "delete": "حذف",
+      "killStuck": "Kill Stuck"
     },
     "toasts": {
       "readyTitle": "الجلسة جاهزة",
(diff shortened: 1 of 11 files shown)


rmyndharis:
Thanks @Abhishekrajpurohit — the i18n coverage across all nine locales is great.

Closing because it's blocked: the "Kill Stuck" button calls `POST /api/sessions/:id/force-kill`, which doesn't exist on `main`. Its backend (#329) is empty and I've closed it — the `pkill` approach doesn't fit the engine-neutral session layer, and the stuck-session case it targeted is already handled by the time-bounded engine teardown shipped in v0.4.x.

If we add a proper engine-level force-recover endpoint later, this UI would be a great fit to rebase onto it. Closing for now so it isn't merged calling a missing endpoint — thanks for the contribution!

PR #329, opened 2026-06-19
fix: add force-kill endpoint to terminate orphaned Chromium processes

Abhishekrajpurohit:
## Summary
- Adds `POST /api/sessions/:id/force-kill` — kills orphaned Chromium processes, resets session to `disconnected` without wiping credentials or DB record
- Only authenticated sessions can trigger force-kill
- Closes session gateway connection before `pkill` to avoid stale auth state
- Logs action as `force_kill` for audit trail

## Files
- `src/modules/session/session.controller.ts` — new endpoint
- `src/modules/session/session.service.ts` — `forceKillSession()` implementation  
- `docker-entrypoint.sh` — ensures Chromium profile cleanup on restart


rmyndharis:
Thanks for tackling stuck-session recovery, @Abhishekrajpurohit.

I'm going to close this, for two reasons:

**1. The branch is empty.** `fix/session-orphaned-chromium-kill` currently has no commits ahead of `main` — the controller endpoint, `forceKillSession()`, and `docker-entrypoint.sh` changes described in the PR aren't in the diff, so there's nothing here to merge. Looks like the commits were lost in a rebase/force-push.

**2. The `pkill` approach doesn't fit the engine abstraction.** The session layer is engine-neutral (it drives both whatsapp-web.js and Baileys); `pkill chromium` is Puppeteer-specific — meaningless for Baileys, and it risks killing *other* sessions' (or the system's) Chromium, not just the stuck one.

The underlying need is also largely covered now: as of v0.4.x, `delete()` and `stop()` do a **time-bounded, isolated engine teardown** (force-destroy with a 10s cap), so a hung Chromium no longer wedges a session. If there's still a gap, the right shape is a `forceDestroy()` on the engine interface (each adapter cleans up its own resources) rather than a process-level `pkill` in the shared service — happy to review that. Thanks again!

PR #335, opened 2026-06-19
fix(webhook): deliver session lifecycle events and key them per occurrence

rmyndharis:
## Problem

Four webhook events — `session.qr`, `session.authenticated`, `session.disconnected`, `session.status` — are declared in `WEBHOOK_EVENTS` and accepted at subscription, but **no code path ever dispatched them**. A subscriber (the n8n trigger node, the dashboard, a custom integration) registers for `session.disconnected`, the webhook is stored active, and the events never arrive — silent non-delivery on a documented happy path. The n8n docs compounded it by advertising two names (`session.connected`, `session.qr_ready`) that aren't in the event set at all, so they were rejected with a 400 at registration.

## Changes

- **Wire the four `session.*` events** from their existing engine-callback emit sites in `session.service.ts` (`onQRCode` → `session.qr`, `onReady` → `session.authenticated`, `onDisconnected` → `session.disconnected`, and the `updateStatus()` chokepoint → `session.status`). Delivery is fire-and-forget (`void dispatch(...)`), so a webhook failure never disrupts the engine.
- **Per-occurrence idempotency keys for lifecycle events.** Lifecycle events recur with identical content (the same phone on every reconnect, a constant `'logged out'` disconnect reason, the same status across cycles). The content-hash key collapsed these onto one value, so a consumer deduping on `X-OpenWA-Idempotency-Key` would silently drop every recurrence. Recurring lifecycle keys are now salted with a per-dispatch occurrence timestamp captured **once** and reused across retries — distinct occurrences get distinct keys, retries of the same delivery stay stable. Message keys remain purely content-based.
- **No duplicate `session.status`.** Some engines (whatsapp-web.js) signal one transition via both `onStateChanged` and a dedicated callback, which would POST `session.status` twice. The webhook now fires only when the status actually changes from the last one sent.
- **Docs corrected.** The n8n integration guide now lists the real, firing event names; `group.*` is documented as reserved (accepted on subscribe but not emitted yet — no engine source exists).

## Compatibility

Non-breaking. No declared event was removed and no response shape changed; previously-dead events now deliver, and the idempotency key is an opaque dedup token whose format consumers don't depend on. SemVer: patch.

## Tests

- `session.service.spec.ts`: each `session.*` event dispatches with the right payload from its emit site; `session.status` is not double-dispatched when `onStateChanged` and a dedicated callback report the same status.
- `idempotency.util.spec.ts`: recurring lifecycle keys are distinct per occurrence, retry-stable for the same occurrence, and message keys are unaffected by the occurrence salt.
- Full backend suite 738/738; dashboard build + lint clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Webhook subscriptions for session lifecycle events now deliver.** `session.status`, `session.qr`,
+  `session.authenticated` and `session.disconnected` were accepted on subscribe but were never dispatched, so
+  consumers (including the n8n trigger node) waited indefinitely. They now fire from the engine lifecycle. The
+  n8n integration docs are corrected to the real event names (`session.authenticated`, `session.qr` —
+  previously documented as the non-existent `session.connected`/`session.qr_ready`, which were rejected at
+  registration). `group.*` events remain accepted on subscribe but are documented as reserved (not emitted
+  yet).
+
 ## [0.4.2] - 2026-06-19
 
 Bug-fix and hardening release: access-control tightening, session-lifecycle resilience, data-migration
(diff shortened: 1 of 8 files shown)


rmyndharis:

PR #338, opened 2026-06-19
fix(security): pin outbound webhook and media fetches to the SSRF-validated IP

rmyndharis:
## Problem

The SSRF guard and the actual outbound connection resolved DNS **independently**: `assertSafeFetchUrl(url)` did its own `dns.lookup` to validate the host, then `fetch(url)` performed a *second*, unrelated resolution at connect time. An attacker controlling a domain's authoritative DNS with a ~0 TTL could answer a public address to the guard and an internal one (loopback, RFC1918, `169.254.169.254` cloud metadata) to the fetch — a classic DNS-rebinding TOCTOU. This affected webhook delivery (the OPERATOR-set URL) and server-side media fetches.

## Fix

A single guarded helper, `withSafeFetch`, that resolves and validates the host **once**, then pins the connection to the vetted IP via an undici dispatcher whose `connect.lookup` returns only those addresses — so the connection cannot be re-resolved between check and connect. Key invariants preserved:

- **TLS SNI + `Host` header stay the original hostname** (not the IP), so certificate validation and virtual hosting are unaffected.
- **All vetted addresses are offered**, so A-record round-robin / failover still works.
- **Redirects are refused** (`redirect: 'manual'` + `assertNoRedirect`) — the guard only validated the original host.
- The `WEBHOOK_SSRF_PROTECT` opt-out and `SSRF_ALLOWED_HOSTS` escape-hatch behave exactly as before; literal-IP and allowlisted-host targets need no pin.

Routed through it: webhook delivery (direct, queued, and the test endpoint), the engine-neutral media download, **and** the whatsapp-web.js send-media-by-URL path — which previously used `MessageMedia.fromUrl` (bundled `node-fetch`, its own unpinned re-resolution) and is the default engine, so it was still exploitable. It now fetches via the pinned helper and builds `MessageMedia` from the bytes, mirroring the Baileys path.

## Dependency

Adds `undici` (the official Node HTTP client — the engine behind global `fetch`), pinned to v6 to match Node 22's bundled major. Global `fetch` exposes no per-request DNS hook; undici's dispatcher is the documented way to pin a connection's IP while keeping SNI. No audit advisories.

## Compatibility

No wire-format or response-shape change; non-breaking. SemVer: patch.

## Tests

- `ssrf-guard.spec.ts`: `pinnedLookup` returns the captured addresses and never re-resolves (rebind-immune); `withSafeFetch` passes a real pinned dispatcher for a hostname target (a regression that removes the pin fails this), refuses redirects on the pinned path, and is fail-closed before any socket.
- `load-remote-media.spec.ts` / `whatsapp-web-js.adapter.spec.ts`: media now flows through the pinned fetch and never via `MessageMedia.fromUrl`.
- Full backend suite 739/739; dashboard build + lint clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- **Outbound webhook and media fetches are pinned to the SSRF-validated IP.** The host safety check
+  and the actual connection previously resolved DNS independently, leaving a narrow window in which a
+  hostile low-TTL DNS answer could pass validation as a public address and then have the connection
+  re-resolve to an internal one (DNS rebinding). The connection now reuses the exact address vetted by
+  the guard — preserving the original hostname for TLS SNI and the `Host` header, and offering all
+  vetted addresses so A-record failover still works — closing the window across webhook delivery
+  (direct, queued, and test) and server-side media downloads.
+
 ## [0.4.2] - 2026-06-19
 
 Bug-fix and hardening release: access-control tightening, session-lifecycle resilience, data-migration
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "openwa",
-  "version": "0.4.1",
+  "version": "0.4.2",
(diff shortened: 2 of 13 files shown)


rmyndharis:

PR #339, opened 2026-06-19
fix(plugins): persist plugin enable/config and restore configuration on restart

rmyndharis:
## Problem

Enabling, disabling, or configuring a plugin only changed in-memory state. The storage setters (`setPluginStatus`/`setPluginConfig`) update an **existing** registry entry, but no entry was ever created — `setPluginEntry` had zero callers — so every write was a silent no-op. The API still returned `{success: true}`, yet on restart an enabled plugin came back disabled and its saved configuration (including secrets such as an API key) was gone.

## Fix

- **A registry entry is created when a plugin loads** (`ensureRegistryEntry`), so subsequent enable/disable/config writes actually persist.
- **Configuration is restored on the next start** — a plugin's saved config survives a restart instead of resetting to empty.
- **Boot never auto-enables a plugin.** The entry's status is reconciled to `INSTALLED` to match the freshly-loaded runtime, so a previously-enabled plugin is re-enabled by an explicit admin action (which runs the lifecycle) rather than re-executing plugin code at startup. The runtime and the registry never disagree about status.
- **Built-in engine config stays env-driven each boot.** The runtime config is `{...envDefault, ...persistedOperatorOverrides}`, so a changed `.env` value (e.g. `PUPPETEER_EXECUTABLE_PATH` on a new image) always wins and is never frozen to a first-boot snapshot; an operator's explicit overrides still win for the keys they set.
- **`registry.json` is written owner-only (`0600`)**, since plugin config can now hold secrets at rest.

## Compatibility

No API or response-shape change; non-breaking. Behavior change: after a restart, an enabled plugin must be re-enabled (its configuration is preserved) — a deliberate, security-conscious choice so startup never auto-runs plugin code.

## Tests

- A registry entry is created on load (complete, `INSTALLED`, `builtIn`) and a status write then persists across a simulated restart.
- An operator's config is restored on the next load; an env-derived built-in config is **not** frozen (a changed default wins on the second boot).
- A re-registered plugin reports `INSTALLED` after restart (no auto-enable, no registry/runtime divergence).
- `registry.json` is written without group/other access.
- Full backend suite 733/733; dashboard build + lint clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Plugin enable/disable and configuration now persist.** Enabling, disabling, or configuring a plugin
+  updated only in-memory state — the registry on disk was never written, so the change was silently lost
+  on restart (an enabled plugin came back disabled and its saved configuration, including secrets such as
+  an API key, was gone), while the API still reported success. A registry entry is now created when a
+  plugin loads, so these writes are durable, and a plugin's saved configuration is restored on the next
+  start. For safety, plugins are not auto-enabled on boot — re-enable them after a restart; their
+  configuration is preserved.
+
 ## [0.4.2] - 2026-06-19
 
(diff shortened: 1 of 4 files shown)


rmyndharis:

PR #340, opened 2026-06-19
fix(message): persist bulk-sent messages, sanitize SSRF errors, honor cross-instance cancel

rmyndharis:
## Problem

Bulk batches called the engine's send methods directly, bypassing the single-send persistence path. Three consequences:

1. **No DB persistence.** Bulk-sent messages were never written to the `messages` table, so they were invisible to chat history and statistics (the engine echo fires the webhook/WS but does not write the DB).
2. **Internal-address disclosure.** On a blocked-destination (SSRF) failure for a media URL, `String(error)` — which names the refused internal host/IP — was stored verbatim in the batch result and returned by the batch-status endpoint.
3. **Cancellation was process-local.** It was tracked only in an in-memory map, so a cancel issued by another replica or after a restart was ignored, and the loop kept sending.

## Fix

- **Persistence.** Bulk reuses the shared outgoing-message persistence (`saveOutgoingMessage`, now public) after each successful send — best-effort, so a persistence failure never flips a sent message to failed. Confirmed no double-write (the engine echo does not persist).
- **Sanitization.** A pure `sanitizeBatchError` maps a blocked-destination error to a generic `SEND_BLOCKED` code; the internal address is no longer stored, returned, or logged verbatim.
- **Cross-instance cancel.** `processBatch` re-reads `batch.status` from the database at its save cadence — and once more before the final write — so a cancel from another instance/restart stops the batch and is never clobbered back to a completed status. The status is read *before* each save so a concurrent `CANCELLED` is never overwritten.

## Scope notes

Implemented additively (bulk-only) rather than re-routing through `message.service`'s full send pipeline, to avoid changing bulk's plugin-hook/typing-humaniser behaviour. Multi-replica cancel is honoured at the save cadence (the existing checkpoint granularity), matching the batch's periodic-save design.

## Compatibility

No API or response-shape change; non-breaking. SemVer: patch.

## Tests

- `sanitizeBatchError`: SSRF → generic message (no internal address); ordinary error passes through under `SEND_FAILED`.
- `processBatch`: persists each sent message; stops when the batch is cancelled in the DB by another instance; does not clobber a `CANCELLED` that lands after the last cadence read.
- Full backend suite 733/733; dashboard build + lint clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Bulk-sent messages are now recorded, their errors no longer leak internal addresses, and a running
+  batch can be cancelled across instances.** Messages sent via a bulk batch went straight to the engine
+  and were never written to the messages table, so they were invisible to chat history and statistics; a
+  blocked-destination (SSRF) failure stored the refused internal address verbatim in the batch result
+  (readable via the batch-status endpoint); and a cancellation was only honoured by the process that
+  created the batch. Bulk sends now persist like single sends, a blocked-destination error is reported as a
+  generic code, and the batch is re-checked against the database as it runs so a cancel issued by another
+  instance (or after a restart) stops it.
+
 ## [0.4.2] - 2026-06-19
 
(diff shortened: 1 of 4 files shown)


rmyndharis:

PR #341, opened 2026-06-19
fix(engine): return the real id for forwarded messages on whatsapp-web.js so delivery status advances

rmyndharis:
## Problem

Forwarding a message on the whatsapp-web.js engine returned a **synthetic** `fwd_<id>` — the underlying library's `Message.forward()` resolves to `void`, so the adapter fabricated an id. The delivery-ack matcher keys strictly on the stored `waMessageId`, and the real ack carries the sent copy's *real* id, so it never matched: forwarded messages were stuck at `SENT` forever and webhook / n8n consumers (keyed on the returned message id) could never correlate the ack. Baileys already returns the real id, so this was also a cross-engine contract drift.

## Fix

After `forward()`, the adapter reads the sent copy back from the destination chat (the most recent outgoing message) and returns its **real** `id._serialized`, so the ack matches and the status advances — aligning whatsapp-web.js with Baileys.

Hardened per review:

- **Best-effort recovery.** The forward has already happened, so the read-back is wrapped so it can never turn a successful forward into a reported failure (e.g. a transient browser/page error). It also null-guards the destination chat lookup.
- **Explicit-unknown fallback.** When the sent copy can't be identified, the adapter returns an *empty* id and `message.service` leaves the row's `waMessageId` unset (NULL). A null id matches no ack and collides with nothing — unlike a synthetic `fwd_` id (un-correlatable) or the source id (which would share `waMessageId` with the source message's own row and let a single ack mis-drive both).

## Known limitation

Recovery selects the most-recent outgoing message by timestamp, so two near-simultaneous forwards/sends to the *same* destination chat could mis-identify the copy. This is documented inline and is a clear net improvement over the previous always-broken behaviour (it is correct in the dominant single-forward case); a content/source-id match could tighten it later.

## Compatibility

`MessageResult` shape unchanged; non-breaking. SemVer: patch.

## Tests

- Returns the real id of the forwarded copy read back from the destination chat (not `fwd_<id>`).
- Returns an explicit-unknown (empty) id when the copy can't be identified.
- A throw during post-forward recovery does not surface as a failed forward.
- Full backend suite 731/731; dashboard build + lint clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Forwarded messages on the whatsapp-web.js engine now report a real WhatsApp message id, so their
+  delivery status advances.** A forward returned a synthetic `fwd_<id>` (the underlying library's forward
+  call yields no id), which the delivery-ack matcher could never match — leaving the forward stuck at
+  "sent" and uncorrelatable by webhook / n8n consumers. The adapter now reads the sent copy back from the
+  destination chat and returns its real id (matching the Baileys engine), falling back to the source id
+  rather than a fabricated one.
+
 ## [0.4.2] - 2026-06-19
 
 Bug-fix and hardening release: access-control tightening, session-lifecycle resilience, data-migration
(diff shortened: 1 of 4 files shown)


rmyndharis:

PR #344, opened 2026-06-19
fix(security): harden outbound requests — IPv6 SSRF embeddings, LibreTranslate guard, proxy validation

rmyndharis:
Three small outbound-request hardening fixes (grouped — all touch the SSRF/egress surface).

## 1. IPv6 SSRF blocklist closes embedded-IPv4 gaps
`isBlockedAddress` handled `::ffff:` (v4-mapped) but allowed every other IPv6 form, so a literal that **embeds an internal IPv4** slipped through:
- **6to4** (`2002::/16`), e.g. `2002:7f00:1::` → `127.0.0.1`, and the compressed-zero form `2002:7f00::` → `127.0.0.0`
- **NAT64** (`64:ff9b::/96`), e.g. `64:ff9b::a9fe:a9fe` → `169.254.169.254`
- deprecated **IPv4-compatible** `::/96`, e.g. `::127.0.0.1`

The literal is now expanded to its 8 hextets (so a compressed all-zero embedded segment is read as `0`, not skipped) and classified by the embedded address. A 6to4/NAT64/compat of a genuinely public IPv4 stays allowed. Reachability is topology-conditional (needs NAT64/6to4 routing on the host), but the guard is meant to be fail-closed.

## 2. LibreTranslate plugin client is SSRF-guarded
Its outbound request carries the configured `api_key` but used `redirect: follow` with no guard. It now runs `assertSafeFetchUrl` and `redirect: error` (gated on the SSRF flag), so the key can't be replayed to a redirect-controlled internal host. A blocked-host error is a deterministic config error, so it no longer trips the client's circuit breaker. **Behavior note:** a loopback LibreTranslate sidecar (`http://localhost:7001`) must be added to `SSRF_ALLOWED_HOSTS` when SSRF protection is on — documented in the plugin's config schema.

## 3. Per-session `proxyUrl` is validated
Previously any ≤255-char string. Now `@IsUrl` requires an `http(s)`/`socks4`/`socks5` URL — accepting credentialed (`http://user:pass@host`), SOCKS, **single-label/container** (`http://squid:3128`) and IP-literal proxies, rejecting malformed/other-scheme values. The engine re-checks it (`isSupportedProxyUrl`, same accept set) and skips a bad value with a warning instead of crashing the browser launch.

## Compatibility
No wire-format change; non-breaking except the documented LibreTranslate-sidecar allowlisting.

## Tests
SSRF blocklist gains 13 IPv6 cases (incl. the compressed-LO and v4-compatible bypasses); proxy DTO validation (single-label + IP-literal accepted, malformed rejected); LibreTranslate guard (internal blocked pre-fetch, `redirect: error`, SSRF error doesn't open the circuit). Full backend 766/766; dashboard build + lint clean.

> The IPv6-embedding tests were green on the first attempt while a compressed-low-hextet bypass (`2002:7f00::`) was still live — an adversarial review caught it; the hextet-expansion rewrite + new cases close it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- **The IPv6 SSRF blocklist now catches NAT64 and 6to4 literals.** A `64:ff9b::/96` (NAT64) or `2002::/16`
+  (6to4) URL embedding an internal IPv4 (loopback, RFC1918, link-local/metadata) previously slipped past the
+  guard; the embedded address is now extracted and checked, on hosts where such routing exists. Genuinely
+  public IPv6 — including a 6to4/NAT64 of a public address — is unaffected.
+- **The LibreTranslate plugin client validates its target and refuses redirects.** Its outbound requests
+  carry the configured `api_key`; they now go through the SSRF guard and use `redirect: error`, so a redirect
+  can't replay the key to an internal host. A localhost LibreTranslate sidecar must be added to
+  `SSRF_ALLOWED_HOSTS` when SSRF protection is on.
+- **Per-session `proxyUrl` is validated.** It must parse to an `http(s)`/`socks4`/`socks5` URL (credentialed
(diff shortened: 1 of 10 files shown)


rmyndharis:

PR #348, opened 2026-06-19
fix(session): reconcile late acks, serialize concurrent reactions, harden hook + ack error paths

rmyndharis:
Four small reliability fixes on the engine→session event path (delivery status, reactions, hooks). No wire-format change.

## 1. A delivery/read receipt that arrives before the send is recorded is no longer lost
A send stores the row optimistically (`PENDING`, no `waMessageId`) and only writes `waMessageId` + `SENT` in a second save *after* the engine returns. The ack callbacks fire from independent engine event handlers, not gated on that send promise — so a fast `delivered`/`read`/`failed` ack could run the status `UPDATE` while the row's `waMessageId` was still null, matching 0 rows. Each ack is one-shot, so the message stayed stuck at "sent" forever. `onMessageAck` now retries the **forward-only, session-scoped** advance **once** after a short delay, reconciling the status. The transition guard is unchanged, so the retry can't downgrade a higher status.

## 2. Concurrent reactions on the same message no longer overwrite each other
`onMessageReaction` did a read-modify-write of the single `metadata.reactions` JSON column with an `await` between the read and the full-row save. Two reactions on the same message interleaved across that await: both read the same snapshot, both saved the whole map, last writer won — silently dropping a reaction. Reaction writes are now **serialized per `(sessionId, waMessageId)`** through a small promise-chain map that cleans up its own entries (no leak), so every reaction is applied on top of the previous one.

## 3. The ack DB write no longer escapes to the global unhandled-rejection backstop
The ack status `UPDATE` was the only `message.*` handler without a `.catch` — a connection-level rejection (pool exhaustion, `SQLITE_BUSY`) surfaced through the global backstop with no per-message context. It now logs locally like its sibling handlers.

## 4. A plugin hook that reports an error no longer has its partial output applied
`HookResult` allows returning `data` and `error` together. The pipeline applied `currentData = result.data` **before** checking `result.error`, so a handler that mutated the payload *and* signalled failure had its (possibly half-transformed) data propagated to persistence, webhooks, and the WebSocket as if it succeeded. The data assignment is now gated on `result.error === undefined` — an errored handler's mutation is discarded (on both the continue and the stop path). Error isolation (catch-and-continue) is unchanged.

## Tests
New unit tests: ack retry fires once on `affected:0` and **not** when a row is advanced (fake timers); a rejected ack update produces no unhandled rejection; concurrent reactions from two senders both survive (DB-aliasing simulated with per-read snapshots); the reaction **remove/delete** branch; a failed reaction write doesn't block a later one; the serialization map is emptied after draining; the hook gate discards an errored mutation on both the continue and `continue:false` paths. Full backend 737/737; dashboard build + lint clean.

> An adversarial review of this diff found **no must-fix issues** and mutation-verified that the regression tests genuinely fail without each fix; the added remove-branch / chain-after-failure / map-cleanup / `continue:false` tests close the coverage gaps it surfaced.

## Deferred (not in this PR)
Persisting **phone-composed self-messages** (sends typed on the linked phone) into the local `messages` table is intentionally left out: it requires a `UNIQUE(sessionId, waMessageId)` index migration, a boot-time dedup of legacy rows, and an upsert across two write paths (a naive insert double-persists API sends). That's a higher-risk change better done as its own PR with the migration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **A delivery/read receipt that arrives before the send is fully recorded is no longer lost.** A fast
+  `delivered`/`read`/`failed` ack could reach the server before the just-sent message's id (`waMessageId`)
+  was committed, so the status update matched no row and the message stayed stuck at "sent" indefinitely.
+  The ack now retries once after a short delay (still forward-only and session-scoped, so it can't
+  downgrade a higher status), reconciling the stored status.
+- **Concurrent reactions on the same message no longer overwrite each other.** Two reactions arriving
+  together both read the same stored snapshot and the last save won, silently dropping a reaction.
+  Reaction writes are now serialized per message, so every reaction is preserved.
+- **A failed delivery-status database write is logged with per-message context** instead of escaping to the
+  global unhandled-rejection backstop (the ack update was the one message handler missing its `.catch`).
+- **A plugin hook that reports an error no longer has its partial output applied.** A hook returning both
+  mutated data and an error had the (possibly half-transformed) data propagated to persistence, webhooks,
+  and the WebSocket as if it succeeded; an errored hook's data mutation is now discarded.
+
 ## [0.4.2] - 2026-06-19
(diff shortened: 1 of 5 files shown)


rmyndharis:

PR #350, opened 2026-06-19
fix(contract): honor webhook timeout, bound shutdown, unify unsupported-op 501, validate engine/storage env

rmyndharis:
Eight small contract / lifecycle / config / doc fixes. No breaking changes; defaults unchanged.

## Fixed
- **`WEBHOOK_TIMEOUT` is honored on the primary (queued) path and the test endpoint.** The configured timeout was applied only on the deprecated direct path; the BullMQ webhook processor and `POST /webhooks/:id/test` hardcoded 10s, so raising `WEBHOOK_TIMEOUT` had no effect when `QUEUE_ENABLED=true` (the recommended production path). All three sites now read `webhook.timeout`. *(The processor gains a `ConfigService` constructor dependency; `ConfigModule` is global so DI resolves it.)*
- **Graceful shutdown is bounded.** `CacheService.onModuleDestroy` awaited `redis.quit()` with no deadline; on a half-open socket the QUIT reply never arrives, blocking `app.close()` until the orchestrator `SIGKILL`s the process. It now force-disconnects after a short timeout (timer cleared on a clean quit) so shutdown always completes.
- **Unsupported status/catalog operations return a consistent `501`.** Six whatsapp-web.js stubs threw a raw `Error` → HTTP 500, while Baileys returned 501 for the same ops. They now throw `EngineNotSupportedError` (501) uniformly. *(The silent `return null/[]` read stubs are intentionally left for a capability-gated follow-up — flipping a 200-empty to 501 would change behavior for clients that treat empty as success.)*
- **A misconfigured `ENGINE_TYPE` / `STORAGE_TYPE` fails fast at boot.** A typo (`ENGINE_TYPE=bailys`) was swallowed and silently fell back to the legacy engine; `validateEnv` now whitelists the registered ids.

## Changed
- **`/api/metrics` is memoized for a few seconds.** Each scrape ran a full session scan + several aggregate queries; back-to-back scrapes (or multiple Prometheus replicas) now share a short-lived rendered result. Stale-by-a-few-seconds metrics are expected for Prometheus. The `METRICS_TOKEN` gate and `@SkipThrottle` are unchanged.
- Internal: removed a dead `if (!validKey)` branch in the WebSocket connect handler — `validateApiKey` always throws on failure, so the surrounding `catch` is the real rejection path. The load-bearing `handleSubscribe` null-branch (revoked-mid-connection) is untouched.

## Documentation
- Documented the webhook `idempotencyKey` / `deliveryId` fields (body + `X-OpenWA-Idempotency-Key` / `X-OpenWA-Delivery-Id` headers) and the dedup-on-`idempotencyKey` rule in the n8n integration guide.
- Corrected `.env.example` rate-limit variables to the names the app actually reads (`RATE_LIMIT_MEDIUM_TTL` / `RATE_LIMIT_MEDIUM_LIMIT`, in milliseconds); the advertised `RATE_LIMIT_TTL` / `RATE_LIMIT_MAX` were read by nothing.

## Tests
New/updated unit tests: webhook timeout flows from config on the processor **and** both `webhook.service` sites (`AbortSignal.timeout` spy, distinct value); cache shutdown force-disconnects when `quit()` hangs and doesn't when it's clean; `validateEnv` accepts/rejects engine & storage ids; metrics `render()` is memoized within the TTL and recomputes after; the WS connect rejection asserts the real throw contract. Full backend 736/736; dashboard build + lint clean.

> An adversarial review found **no must-fix issues**; its one nice-to-have (the two `webhook.service` timeout sites weren't value-asserted) was closed before opening this PR.

## Deferred (own follow-up)
- **Settings `PUT`**: `PUT /settings` mutates an in-memory field that's lost on restart and never re-applied to runtime config. Making it honest is an API-contract decision (persist + re-apply the truly-mutable values vs. return `501`/read-only for env-derived ones), so it's out of scope here.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -168,10 +168,15 @@ WEBHOOK_SSRF_PROTECT=true
 # MEDIA_DOWNLOAD_TIMEOUT_MS=30000     # abort a slow media download (default 30s)
 
 # =============================================================================
-# RATE LIMITING
+# RATE LIMITING  (all TTLs are in MILLISECONDS)
 # =============================================================================
-RATE_LIMIT_TTL=60                   # Time window in seconds
-RATE_LIMIT_MAX=100                  # Max requests per window
+# The "medium" tier is the one enforced on the API; short/long are optional extra tiers.
+RATE_LIMIT_MEDIUM_TTL=60000         # window in ms (default 60000 = 60s)
+RATE_LIMIT_MEDIUM_LIMIT=100         # max requests per window
+# RATE_LIMIT_SHORT_TTL=1000         # 1s burst window (default)
+# RATE_LIMIT_SHORT_LIMIT=10
+# RATE_LIMIT_LONG_TTL=3600000       # 1h window (default)
+# RATE_LIMIT_LONG_LIMIT=1000
 
 # =============================================================================
 # PLUGINS
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`WEBHOOK_TIMEOUT` is honored on the primary (queued) delivery path and the test endpoint.** The
+  configured timeout was applied only on the deprecated direct path; the BullMQ webhook processor and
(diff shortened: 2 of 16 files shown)


rmyndharis:

PR #351, opened 2026-06-19
feat(dashboard): Kill Stuck button for failed sessions

Abhishekrajpurohit:
Adds a **Kill Stuck** button to session cards with `status === 'failed'`.

Clicking shows a confirmation modal, then calls `POST /api/sessions/:id/force-kill` which triggers the engine-level `forceDestroy()` — killing Chromium processes holding the session's profile directory, then tearing down the client.

This PR was previously part of #317 (closed) which depended on a backend PR that was also closed. The backend is now on main as of v0.4.2+.

### DIFF
--- a/dashboard/src/i18n/locales/ar.json
+++ b/dashboard/src/i18n/locales/ar.json
@@ -226,7 +226,8 @@
       "start": "بدء",
       "stop": "إيقاف",
       "reconnect": "إعادة الاتصال",
-      "delete": "حذف"
+      "delete": "حذف",
+      "killStuck": "Kill Stuck"
     },
     "toasts": {
       "readyTitle": "الجلسة جاهزة",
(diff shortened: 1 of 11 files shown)


rmyndharis:
Thanks @Abhishekrajpurohit! Heads-up: the backend this depends on (`POST /api/sessions/:id/force-kill` → engine `forceDestroy()`) wasn't actually on `main` — the original backend PR (#316/#329) was closed with an empty branch, so this button would currently 404.

I've just opened **#352** with that backend, implemented the engine-neutral way: each adapter force-kills only its **own** resources (whatsapp-web.js SIGKILLs *its own* Chromium process — not a process-wide `pkill` that would take down other sessions; Baileys ends its socket), wrapped in the existing time-bounded/isolated teardown. The route + payload match what this PR calls.

Plan: merge #352 first, then this. Could you rebase on `main` once #352 lands so CI runs against the real endpoint? The dashboard side here looks good.

rmyndharis:

PR #352, opened 2026-06-19
feat(session): force-kill a stuck session (engine forceDestroy + POST /sessions/:id/force-kill)

rmyndharis:
Backend for the **"Kill Stuck"** dashboard button (#351). #351's button calls `POST /api/sessions/:id/force-kill`, but that endpoint did not exist on `main` (the original backend PR, #316/#329, was closed with an empty branch). This adds it the engine-neutral way.

## Why not `pkill chromium`
The earlier attempt used a process-level `pkill`, which is wrong for an engine-abstracted layer: it's Puppeteer-specific (meaningless for Baileys) and would kill **other** sessions' (or the system's) Chromium, not just the stuck one. Instead this puts the responsibility on the engine — each adapter kills only its **own** resources.

## What this adds
- **`IWhatsAppEngine.forceDestroy()`** — force-kill the engine's own resources, then best-effort graceful teardown.
  - **whatsapp-web.js**: SIGKILLs *this client's own* Chromium process via `client.pupBrowser.process()`, then best-effort `client.destroy()`. A missing process handle or a hung `destroy()` can't prevent teardown.
  - **Baileys**: ends its own socket (no separate browser process), so `forceDestroy()` is just a `destroy()`.
- **`SessionService.forceKill(id)`** — mirrors `stop()`'s lifecycle: sets the stop-mark (blocks an in-flight reconnect from resurrecting the session it just killed; a later `start()` clears it), cancels pending reconnects, and runs the **time-bounded, isolated** teardown (`teardownEngineSafely`, 10s cap) so a still-wedged engine can't hang the call. The Map is reconciled and the session is left `DISCONNECTED` and restartable.
- **`POST /sessions/:id/force-kill`** (OPERATOR, mirrors `stop`) → 200 with the session, 404 if unknown. Audited as `session_force_killed`.

No migration (the audit-action column is a free `varchar`), no new dependency, no change to existing routes.

## Tests
- whatsapp-web.js `forceDestroy`: SIGKILLs only its own browser process then destroys; still completes when the process handle is gone and `destroy()` rejects; no-op with no client.
- `forceKill`: force-destroys + reconciles the Map + keeps the stop-mark (so a late reconnect can't revive it); completes even when `forceDestroy()` rejects; 404 for an unknown session.

Full backend 734/734; dashboard build + lint clean.

## Merge note
Merge this **before** #351 so the dashboard button has a working endpoint. The two together close the stuck-session-recovery gap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Force-kill a stuck session.** `POST /sessions/:id/force-kill` (OPERATOR) recovers a session whose engine
+  is wedged and won't respond to a normal stop/delete: the whatsapp-web.js engine **SIGKILLs its own Chromium
+  process directly** (never a process-wide kill that could take down other sessions), then best-effort tears
+  the client down; the Baileys engine ends its socket. The teardown is time-bounded and isolated, the session
+  is left `DISCONNECTED`, and it can be started again. Backs the dashboard's "Kill Stuck" button.
+
 ## [0.4.2] - 2026-06-19
 
 Bug-fix and hardening release: access-control tightening, session-lifecycle resilience, data-migration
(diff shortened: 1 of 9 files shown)


rmyndharis:

PR #363, opened 2026-06-19
fix(baileys): allow sending captions with document messages

Fernixp:
The baileys adapter was ignoring the `caption` field when sending document messages. This commit passes the `caption`
  from the media payload to the `sendContent` method. Additionally, it updates `message.service.ts` to correctly store the document's caption in the database instead of defaulting to the filename.

## Description
This PR fixes an issue where the `caption` field was being completely ignored when sending document payloads (like PDFs) using the Baileys adapter.

- Updated `baileys.adapter.ts` to properly pass the `caption` property to the underlying `sendContent` method.
    - Updated `message.service.ts` so the database stores the `caption` as the message `body` for documents, falling back to the filename if a caption isn't provided.
    - Added test coverage in `baileys.adapter.spec.ts` to prevent future regressions.

## Type of Change
- [x ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [x] Tests added/updated
- [ ] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

N/A
## Related Issues
Closes #

### DIFF
--- a/src/engine/adapters/baileys.adapter.spec.ts
+++ b/src/engine/adapters/baileys.adapter.spec.ts
@@ -852,12 +852,14 @@ describe('BaileysAdapter media sends', () => {
       mimetype: 'application/pdf',
       data: Buffer.from('PDFDATA').toString('base64'),
       filename: 'doc.pdf',
+      caption: 'a document',
     });
     expect(loadRemoteMediaBuffer).not.toHaveBeenCalled();
     expect(fakeSock.sendMessage).toHaveBeenCalledWith('628111@s.whatsapp.net', {
       document: Buffer.from('PDFDATA'),
       mimetype: 'application/pdf',
       fileName: 'doc.pdf',
+      caption: 'a document',
     });
   });
 
--- a/src/engine/adapters/baileys.adapter.ts
+++ b/src/engine/adapters/baileys.adapter.ts
@@ -380,7 +380,12 @@ export class BaileysAdapter implements IWhatsAppEngine {
   async sendDocumentMessage(chatId: string, media: MediaInput): Promise<MessageResult> {
     this.ensureReady();
     const { data, mimetype } = await this.resolveMediaBuffer(media);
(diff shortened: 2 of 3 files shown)


rmyndharis (approved):
Nice catch and a clean fix, @Fernixp — approving.

I verified the gap: the whatsapp-web.js adapter already forwards the caption for documents (`sendMediaMessage` → `client.sendMessage(..., { caption: media.caption })`), so the Baileys `sendDocumentMessage` was the only path dropping it. This brings the two engines to parity.

The `message.service` change (`body: dto.caption || dto.filename || ''`) is a good engine-neutral improvement too: both engines now store the document's caption as the message body, falling back to the filename when there's no caption — and it's backward-compatible (the no-caption case is unchanged). The added spec assertion locks it in.

CI is green (Lint/Test/Build/Dashboard) and the change is minimal and well-scoped. 👍 Thanks for including the test.

rmyndharis:
Thanks for the clean, well-scoped fix, @Fernixp — and for including a regression test. I traced it through end to end and everything lines up:

- The caption now flows correctly: `dto.caption` → `MediaInput.caption` → `sendContent({ document, ..., caption })` → Baileys `sendMessage`. `MediaInput.caption` and the validated `caption` DTO field (max 1024) were already in place, so this was purely the adapter dropping it.
- This brings Baileys to parity with whatsapp-web.js, whose `sendDocumentMessage` already forwarded `caption: media.caption`. Baileys was the only path losing it — nice catch.
- The `message.service.ts` change (`body: dto.caption || dto.filename || ''`) is a good engine-neutral improvement and is backward-compatible: the no-caption case is unchanged.
- The added spec assertion (`sendMessage` called with `caption: 'a document'`) locks the behavior in nicely.

CI is green across Lint/Test/Build/Dashboard and the change is minimal. This is ready to merge — I'll fold a short CHANGELOG note in on my end so it shows up in the next release. Appreciate the contribution!

PR #374, opened 2026-06-20
feat(engine): typed WaId + persistent lid->phone resolution table + message from-filter

### LINKED ISSUE #349: [Feature]: typed `WaId` identity value object + a `lid <-> phone` resolution table
### Pre-flight

- [x] I searched existing issues and this isn't already requested

### Problem / motivation

Every WhatsApp identity in the app is a raw JID string today (`from`, `to`, `chatId`, contact/chat
`id`, stored values, send DTOs). Two problems fall out of that:

- The neutral `@c.us` convention is enforced only by discipline plus the `wa-id` helpers from #342 -
  nothing actually stops a string in the wrong dialect from slipping through.
- The hard case, a **lid whose phone we don't know**, isn't visible in the type. To the code, `123@lid`
  and `628...@c.us` are interchangeable strings, so a phone-based match against a lid contact quietly
  fails and nothing signals that resolution was even needed. Concretely: a webhook from-filter that
  matches a DM today **silently misses the same person's already-resolved group message**.

Note this is a **behavior** gap, not just a typing nicety: response shapes don't change, but filter and
matching behavior does (misses become hits once a lid is resolved).

### Proposed solution

A small value object that makes the kind, the (maybe-unknown) phone, and the lid explicit, so the type
carries the rules instead of every caller re-deriving them.

```ts
type WaIdKind = 'user' | 'group' | 'lid' | 'status' | 'newsletter' | 'broadcast';

class WaId {
  readonly kind: WaIdKind;
  readonly phone?: string;    // E.164 digits, when known
  readonly lid?: string;      // lid number, when this id is/carries a lid
  readonly groupId?: string;  // for groups
  readonly raw: string;       // original engine jid; debug-only, excluded from matching

  toNeutral(): string;
  refersToSamePerson(other: WaId): boolean;   // relational, NOT a hashable key

  static fromEngineJid(jid: string, resolvePhone?: (jid: string) => string | null): WaId;
  static fromUserInput(value: string): WaId;
}
```

Key properties:

- `WaId` is **in-memory only**. It is never persisted as rows and introduces no table of identities.
  The wire/storage format stays the neutral string. The *only* new persistent artifact is the
  resolution table below.
- Matching is relational, not structural equality: two `WaId`s refer to the same person if they share a
  known phone or share a lid. It's **not** a clean equivalence relation while lids are unresolved
  (`{phone:1, lid:X}` matches both `{phone:1}` and `{lid:X}`, but those two don't match each other), so
  there's a deliberate third outcome: matched / didn't match / **couldn't tell** (phone unknown). `raw`
  is excluded from matching so the same person seen via two engines doesn't split.
- Adapters build `WaId` at the engine boundary; the contract documented as prose on `IWhatsAppEngine`
  in #342 becomes type-enforced. The wire stays a neutral string in both directions, so REST/webhook
  output is unchanged.

### The `lid <-> phone` resolution table

`WaId` resolution only helps if `lid -> phone` lookup is synchronous (filters/dispatch can't await a
network call). Resolution reads from an **in-memory map loaded from a persisted table on boot and
written through on every new mapping**. Two write sources with opposite costs:

- **Passive (free):** mappings WhatsApp pushes to us - history sync (`lidPnMappings`),
  `contacts.upsert`, and message `senderPn`/`participantPn`. Captured eagerly, batched writes.
- **Active (rate-limited):** `getContactLidAndPhone` (whatsapp-web.js) / the Baileys lid-mapping store,
  only when something needs a phone we don't have. Done lazily and written through immediately
  (including negative results) so it's one network call per unknown lid, ever.

Today this is scattered and mostly ephemeral (Baileys' in-memory `lidToPn`, the session service's
per-session `lidPhoneCache` gated by `RESOLVE_LID_TO_PHONE`, whatsapp-web.js's inline lookup). The
proposal consolidates them into one persisted, cross-session table on the **data** connection (so a
TypeORM migration under `src/database/migrations/`). A stored mapping is "best known, not forever"
(numbers get recycled), so it's treated as a cache WhatsApp can correct.

## Open questions (where I'd like input before building)

1. **Wire/storage format:** keep the neutral JID string (`<phone>@c.us`) as the serialized form (least
   churn, human-readable, matches existing DTO examples and stored values), or store bare digits? I'm
   leaning toward keeping the neutral string - `WaId` is the in-memory type, the string is the boundary
   format.
2. **Active resolution on Baileys:** the eager-passive + lazy-active write-through model feels settled.
   The open bit is whether the installed Baileys exposes a lid-mapping lookup
   (`signalRepository.lidMapping`) to rely on for active resolution, since `resolveContactPhone` is
   cache-only today.
3. **`RESOLVE_LID_TO_PHONE` semantics:** today it gates `senderPhone`. Should `WaId` resolution respect
   it (privacy), or always resolve internally into the table and just gate what's *exposed*?

## Scope
Staged so nothing observable breaks; the wire stays a neutral string throughout.

1. Land `wa-id.ts` parse/format primitives - **done in #342**.
2. Add `WaId` as a thin wrapper over those primitives, keeping the public string fields.
3. Move consumers over one module at a time (id-comparison spots first: contact matching, then
   groups/messages).
4. Parse DTOs / API inputs into `WaId` at the edge, keeping the wire format neutral.
5. Switch `IncomingMessage` (and friends) to typed `WaId` fields internally; delete scattered string
   normalization. REST/webhook output keeps emitting `toNeutral()`, so nothing observable changes.
6. Fold `isLidSender` / `senderPhone` into `WaId` and deprecate them.

`WaId` type + tests is small (~1 file, builds on `wa-id.ts`). Consumer migration is medium and spread
out (~50 files mention a dialect literal, about half specs/fixtures), done module-by-module behind the
existing string fields so each step is reviewable and reversible. Worth starting only once the #342
contract has settled in `main`.

### Alternatives considered

1. **Status quo - raw strings + a mapping side-table + dual webhook fields (the WAHA/Evolution shape).**
   Rejected as the *primary* model: it pushes "is this the same person" onto every consumer and every
   internal call site, which is the discipline-dependent setup that produces the silent-miss bug here
   and the recurring lid-leak bugs in both peers. We reuse the parts that work (a persisted lid<->pn
   table, eager `senderPn`/`*Alt` capture) but put them behind one typed identity instead of exposing
   raw strings.
2. **Variant-set + substring matching (Evolution's `jidOptions`).** Rejected: brittle and only works for
   variants already expanded into the set.
3. **Lean entirely on the engine's mapping store, no neutral type.** Rejected: two different mechanisms
   across engines (Baileys' persistent store vs wwebjs' per-call query), the helpers won't unify forms
   anyway, and `raw` differs per engine - so identity would split when switching engines.
4. **Chat-level merge toggle (WAHA's `merge`).** Solves duplicate *threads*, not filter/matching
   correctness - orthogonal, and a symptom-level fix rather than an identity model.
5. **Keep raw strings + lean harder on the `wa-id` helpers.** Rejected: the unresolved-lid case stays
   invisible to the type, so the silent-miss bug remains a matter of caller discipline.

Net: `WaId` + a unified resolution table is the ecosystem's converged building blocks (a lid<->pn
table, capture of pushed mappings) exposed as **one neutral typed identity with an explicit "couldn't
tell" state**, rather than leaving "same person" to each consumer - which is where both peers keep
getting bitten.

(Two narrower in-design variants, separate from the above: a discriminated union on `kind` vs the bag of
optionals, and storing bare digits vs the neutral string - see open question 1.)

### Scope

- [x] I understand some features are limited by the underlying WhatsApp engine (e.g. interactive Buttons/List messages are not supported on whatsapp-web.js).

tobiasstrebitzer:
Closes #349.

Typed `WaId` value object + a persistent, cross-session `lid -> phone` resolution table + a `from` filter on message history. On top of `main` @ v0.4.5.

## Composes with the v0.4.5 Baileys work (not duplicates)

- **#372** added `recordKeyLidMappings(key)` (inbound `senderPn`/`participantPn`). This PR lets it feed the durable table: the method calls `addLidMappings`, which now writes through to the persistent `lid_mappings` table - so **#372's inbound capture now persists cross-session for free**.
- **#370**'s `resolveContactName` is merged with this PR's contact/chat id canonicalization in `toNeutralChat` (neutral `@c.us` id **and** resolved name).

## Motivation

Lid resolution was non-deterministic and ephemeral: a lid sender could be unresolved (`@lid`) early and resolved (`@c.us`) later once the `lid -> pn` map populated, and that map lived only in one Baileys session's memory (lost on restart, not shared across sessions). So a phone-based `from` filter matched a DM but **silently missed the same person's already-resolved group message**. A durable, shared table closes that; a typed `WaId` makes "a lid whose phone we don't know" first-class in the type.

## New

- `src/engine/identity/wa-id.value.ts` - the `WaId` value object over the `wa-id.ts` primitives. In-memory only; `toNeutral()/toString()/toJSON()` yield today's neutral string; three-valued `refersToSamePerson()` (matched / didn't match / **couldn't tell**).
- `src/engine/identity/lid-mapping.entity.ts` + `src/database/migrations/1781200000000-AddLidMappings.ts` - `LidMapping` on the **`data`** connection (portable Postgres + SQLite migration).
- `src/engine/identity/lid-mapping-store.service.ts` - boot-loads the table into an in-memory map, synchronous forward (`getCached`) + reverse (`lidsForPhone`) lookups, write-through `remember()`.

## Wired (no contract changes)

- The store threads `EngineFactory -> BaileysPlugin -> adapter config -> BaileysSessionStore`, mirroring the existing `messageStore` seam.
- `addLidMappings` / `upsertContacts` write through to the table; `resolvePhone` falls back to the table's in-memory cache (kept synchronous).
- `GET /api/sessions/:sessionId/messages?from=` resolves through the table - a phone matches `@c.us`/`@s.whatsapp.net` stored ids **and** any lid resolving to it.
- Baileys contact/chat **listing** ids now emit `@c.us` (the #342-domain conformance that was deferred), merged with #370's name resolution; read-back lookups fold a neutral id back to the engine dialect so the round-trip still hits.

## Acceptance criteria (from your review)

- [x] **Response shapes stay byte-identical.** `WaId` is internal; a spec asserts `WaId.toNeutral()` equals `toNeutralJid()` for every representative id.
- [x] **Resolution table persists on the `data` connection.** Entity on `data`, portable migration, boot-verified.
- [x] **From-filter test - a lid-resolved match becomes a hit.** With the table mapping lid `111 -> 628999`, filtering `from=628999` returns the lid-authored row; a control test shows the prior silent miss.

## Settled open questions

1. Wire/storage format = neutral string, `WaId` in-memory. ✅
2. `6.7.23` has no `signalRepository.lidMapping` lid->phone lookup, so unknown lids degrade to "couldn't tell" (`@lid`); resolution is fed by passive sources (inbound `senderPn`/`participantPn` via #372, `phoneNumberShare`, contacts, history sync). ✅
3. `RESOLVE_LID_TO_PHONE` resolves internally into the table and gates only what's exposed (privacy flag, not a correctness toggle). ✅

## Deferred (issue #349 scope steps 3-6)

Module-by-module migration of consumers onto `WaId` (behind the existing string fields); active, lazy per-miss `lid->phone` lookup. The lid key-flip limitation you're tracking under #349 is addressed at the foundation (durable, cross-session identity); re-keying the specific in-memory consumers is part of the deferred migration.

A follow-up PR (`feat/baileys-sync-hardening`) stacks Baileys sync hardening on top of this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Persistent, cross-session `lid -> phone` resolution + a `from` filter on message history.** A new
+  `lid_mappings` table (on the `data` connection) records the `lid -> phone` mappings WhatsApp pushes us
+  (history sync, contacts) so resolution is shared across sessions and survives restarts, instead of
+  living only in one Baileys session's in-memory map. `GET /api/sessions/:sessionId/messages` now accepts
+  a `from` query param that resolves through this table: filtering by a phone returns not just messages
+  stored as `<phone>@c.us` but also those whose sender was an unresolved `<lid>@lid` that has since
+  resolved to that phone - closing a gap where a phone-based filter silently missed the same person's
+  lid-addressed (e.g. group) messages. The table is populated at runtime from the lid<->phone pairs the
+  Baileys engine observes (inbound message `senderPn`/`participantPn`, the `chats.phoneNumberShare`
+  event, contacts, and history sync), so it fills continuously without re-auth. Internally these ids are
+  now carried by a typed `WaId` value object; it is in-memory only and serializes to the exact same
+  neutral string, so **no webhook / WebSocket / REST response shape changes**.
+
+### Fixed
+
+- **Baileys engine: contact and chat *listing* ids are now engine-neutral (`@c.us`).** `getContacts` /
+  `getChats` / `getContactById` previously returned the raw `<phone>@s.whatsapp.net` id (visible in the
+  dashboard, and mismatched against the `@c.us` chatId stored on messages). They now emit the neutral
+  `@c.us` dialect like the message payloads; the read-back paths (`sendSeen` / `deleteChat` / contact
+  lookup) accept the neutral id and fold it back internally, so sending and marking-read still round-trip.
+  **Consumer-visible:** Baileys contact/chat-list ids flip `@s.whatsapp.net` -> `@c.us` (whatsapp-web.js
+  already used `@c.us`).
+
 ## [0.4.5] - 2026-06-20
 
 A Baileys engine quality-and-correctness release, plus a chat-history enhancement. **Identity:** inbound
@@ -45,7 +70,6 @@ consumer that stored or compared the old ids will see the new value.
   `participantPn`), so the sender of an incoming message resolves to its number and later contact lookups
   succeed. Still best-effort by design — a number is only revealed once WhatsApp delivers the mapping
   (e.g. an inbound message from that contact). (#362)
-
 - **Baileys engine: inbound message ids are now engine-neutral (`@c.us`).** The Baileys adapter emitted
   its native `<phone>@s.whatsapp.net` / `<lid>@lid` ids in message payloads (`from` / `to` / `chatId` /
   `author`, plus revoked and reaction events), while the whatsapp-web.js engine and the rest of the
--- a/src/database/migrations/1781200000000-AddLidMappings.ts
+++ b/src/database/migrations/1781200000000-AddLidMappings.ts
@@ -1,0 +1,32 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+/**
+ * Creates `lid_mappings` - the persisted, cross-session `lid -> phone` resolution table on the `data`
+ * connection. Hand-authored because `synchronize` is off for `data` on Postgres (and optional on
+ * SQLite); the `hasTable` guard keeps it idempotent on a DB where synchronize already created it.
+ */
+export class AddLidMappings1781200000000 implements MigrationInterface {
+  name = 'AddLidMappings1781200000000';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    if (await queryRunner.hasTable('lid_mappings')) return;
+    const isPostgres = queryRunner.connection.options.type === 'postgres';
+
+    if (isPostgres) {
+      await queryRunner.query(
+        `CREATE TABLE "lid_mappings" ("lid" varchar PRIMARY KEY NOT NULL, "phone" varchar, "sessionId" varchar, "updatedAt" timestamp NOT NULL DEFAULT NOW())`,
+      );
+    } else {
+      await queryRunner.query(
+        `CREATE TABLE "lid_mappings" ("lid" varchar PRIMARY KEY NOT NULL, "phone" varchar, "sessionId" varchar, "updatedAt" datetime NOT NULL DEFAULT (datetime('now')))`,
+      );
+    }
+
+    await queryRunner.query(`CREATE INDEX "IDX_lid_mappings_phone" ON "lid_mappings" ("phone")`);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP INDEX "IDX_lid_mappings_phone"`);
+    await queryRunner.query(`DROP TABLE "lid_mappings"`);
+  }
+}
--- a/src/database/migrations/1781200000000-AddLidMappings.ts
+++ b/src/database/migrations/1781200000000-AddLidMappings.ts
@@ -26,7 +26,9 @@ export class AddLidMappings1781200000000 implements MigrationInterface {
   }
 
   public async down(queryRunner: QueryRunner): Promise<void> {
-    await queryRunner.query(`DROP INDEX "IDX_lid_mappings_phone"`);
-    await queryRunner.query(`DROP TABLE "lid_mappings"`);
+    // IF EXISTS so rollback is safe even when the table was created by the `synchronize` path (which
+    // auto-names the index differently) rather than by up()'s explicit `IDX_lid_mappings_phone`.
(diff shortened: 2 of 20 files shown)


rmyndharis:
Thank you for this — it's a genuinely thorough piece of work, and I appreciate how carefully you staged the persistence and kept `resolvePhone` synchronous. I reviewed it closely against the source (migration portability on both SQLite and Postgres, the fire-and-forget write-through with its dedup, the boot-loaded cache, and the `from`-filter), and it all holds up.

I hope you don't mind — I pushed three small maintainer touch-ups directly to the branch so we could land it without a round-trip:

- The `EngineFactory` specs were still constructing with 3 args after you added the `LidMappingStore` 4th parameter, so that new wiring was being exercised as `undefined` (it slips past CI because specs aren't type-checked). They now pass a `LidMappingStore` fake.
- `DROP INDEX/TABLE IF EXISTS` in the migration `down()`, so a rollback is safe even when the table was created by the `synchronize` path (auto-named index).
- A small migration spec mirroring the sibling `AddBaileysStoredMessages` test.

Everything else is yours, and the design decisions are unchanged. One note for release framing: the contact/chat **listing**-id flip to `@c.us` is the consumer-visible piece we deferred in v0.4.5 — you've already documented it correctly in the CHANGELOG; we'll just settle patch-vs-minor at tag time.

Merging now. Thanks again for the careful work — this is a strong foundation for the identity arc.

tobiasstrebitzer:
Thanks @rmyndharis appreciate your additional fixes and kind words

PR #375, opened 2026-06-20
fix(baileys): enable initial sync + dialect-agnostic chatId filter + debug logging

tobiasstrebitzer:
**Depends on #349** - stacks on `feat/typed-waid-349`; that PR's commit appears in this diff until #349 merges. Review/merge #349 first.

Three independent Baileys-engine fixes that surfaced while testing #349 on a real session.

## Initial sync was silently disabled

Baileys defaults `shouldSyncHistoryMessage` to `() => !!syncFullHistory`, so with `syncFullHistory` unset the **entire** initial sync (address-book / app-state) never ran - no contacts, chat list, recent messages, or `lid -> phone` mappings ever arrived (`Socket/index.js` + `Socket/chats.js` go straight to `Online`). The adapter now passes `shouldSyncHistoryMessage: () => true`, enabling the sync while keeping the full-archive download opt-in via `BAILEYS_SYNC_FULL_HISTORY` (WhatsApp then sends the recent window + contact/app-state snapshot, not the entire message history).

Verified against `@whiskeysockets/baileys@6.7.23` source and corroborated by upstream issues (e.g. WhiskeySockets/Baileys#2077, NousResearch/hermes-agent#11951).

## Dialect-agnostic `chatId` filter on message history

A chat addressed as `<phone>@c.us` (the neutral list id) now also returns messages stored under `<phone>@s.whatsapp.net` (e.g. an outbound send addressed by a raw engine id), so the conversation view is no longer empty when the stored and queried dialects differ. Reuses the #349 `from`-filter resolution.

## `BAILEYS_LOG_LEVEL` + receipt instrumentation

`BAILEYS_LOG_LEVEL` (trace|debug|info|warn|error, silent by default) surfaces the Baileys library's own diagnostics; `trace` dumps the decoded WhatsApp wire frames to stdout (context `baileys-wire`). Plus contacts/chats/history-set receipt logging, for diagnosing sync issues.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -70,6 +70,12 @@ PUPPETEER_ARGS=--no-sandbox,--disable-setuid-sandbox,--disable-dev-shm-usage,--d
 # global Node.js version floor. Node.js 22 LTS is recommended for all deployments.
 BAILEYS_AUTH_DIR=./data/baileys
 # NOTE: proxy (PROXY_URL/PROXY_TYPE) is not yet supported by the baileys engine; it is ignored.
+# Pull the FULL message-history archive on connect (large). Off by default: the engine still enables the
+# initial sync (contacts, chats, recent messages, lid mappings) without downloading the entire history.
+BAILEYS_SYNC_FULL_HISTORY=false
+# Surface the Baileys library's own logs for debugging (trace|debug|info|warn|error). Silent by default.
+# `trace` dumps the decoded WhatsApp wire frames to stdout (context "baileys-wire").
+# BAILEYS_LOG_LEVEL=debug
 
 # Humanise single sends: show a "typing…" indicator and pause briefly (length-scaled, jittered)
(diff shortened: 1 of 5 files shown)


rmyndharis:
Rebased onto `main` now that #374 has landed, so this is down to its own three-fix delta. I reviewed each against the source and they hold up nicely:

- `shouldSyncHistoryMessage: () => true` — confirmed against `@whiskeysockets/baileys@6.7.23` that the default (`() => !!syncFullHistory`) was silently skipping the entire initial sync; the full-archive download stays opt-in via `BAILEYS_SYNC_FULL_HISTORY`. Good catch, and an important one.
- Dialect-agnostic `chatId` filter cleanly reuses the candidate-expansion helper from #374.
- `BAILEYS_LOG_LEVEL` is silent by default; just flagging for anyone reading later that `trace` will surface decoded wire frames (incl. message content) by design — opt-in only.

Approving and merging. Thanks for chasing these down on a real session.

PR #409, opened 2026-06-22
chore(docker): fix dashboard build by using npm install instead of np…

JoseFEstevesP:
## Description

This PR fixes the Docker build failure occurring in the `dashboard` stage. 

**The problem:** 
Using `npm ci` in the Dockerfile for the dashboard dependencies was too strict. Since `npm ci` relies strictly on the lockfile and cannot reliably resolve or skip platform-specific optional native bindings (such as `@rolldown/binding-linux-x64-gnu`) during the Docker layer build, the installation was failing.

**The solution:**
- Replaced `npm run dashboard:ci` with `cd dashboard && npm install` inside the Dockerfile. This allows npm to dynamically resolve the correct native binary for the container's architecture (`linux-x64-gnu`) instead of being blocked by lockfile mismatches.
- Explicitly added `@rolldown/binding-linux-x64-gnu` to `optionalDependencies` in `dashboard/package.json` to ensure that if this binary fails to install on other platforms, it won't break the entire build process.

**Impact:** 
The Docker image now builds successfully without altering the actual runtime behavior of the dashboard.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

## Related Issues
Closes #

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,8 +24,9 @@ COPY . .
 
 # Build the API (dist/) and the dashboard SPA (dashboard/dist/). The root `npm ci` above
 # ran before the dashboard source was copied, so its postinstall hook skipped the dashboard
-# deps - install them explicitly here (npm ci, reproducible from dashboard/package-lock.json).
-RUN npm run build && npm run dashboard:ci && npm run dashboard:build
+# deps — install them explicitly here. We use `npm install` (not ci) so that npm correctly
+# resolves transitive native bindings (e.g. @rolldown/binding-linux-x64-gnu).
+RUN npm run build && cd dashboard && npm install && npm run build
 
 # ===== Stage 2: Production =====
 FROM docker.io/node:22-slim AS production
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -22,6 +22,9 @@
     "react-router-dom": "^7.17.0",
     "socket.io-client": "^4.8.3"
   },
+  "optionalDependencies": {
+    "@rolldown/binding-linux-x64-gnu": "^1.0.3"
+  },
   "devDependencies": {
     "@eslint/js": "^10.0.1",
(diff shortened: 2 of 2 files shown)


rmyndharis:
Thanks for digging into this, @JoseFEstevesP — and welcome! 🙏

I'm going to close this one, because the dashboard Docker build is actually passing on `main` as-is, so there isn't a failure to fix here:

- The multi-arch image build (`linux/amd64` + `linux/arm64`) is green on the latest `main`, and the builder stage installs the dashboard with `npm ci` (`npm run dashboard:ci`).
- `dashboard/package-lock.json` (lockfileVersion 3) already includes every `@rolldown/binding-*` platform package — `linux-x64-gnu`, `linux-arm64-gnu`, `linux-x64-musl`, darwin, win32, etc. — so `npm ci` resolves the correct native binary per platform.

On the specific changes:
- `npm ci` → `npm install` would cost us reproducible, lockfile-pinned builds (the rest of the Dockerfile stays on `npm ci`).
- Adding `@rolldown/binding-linux-x64-gnu` to `optionalDependencies` is already covered transitively via `vite`/`rolldown`, only pins the **x64** binding (so it wouldn't help the **arm64** image), and isn't reflected in `package-lock.json` — which would actually make `npm ci` fail on a lockfile mismatch.

If you're hitting a real failure locally, I'd genuinely like to fix the root cause — could you share the exact `docker build`/buildx command, your host architecture, and `node`/`npm` versions, plus the full error? Most local cases are building on Apple Silicon without buildx (defaults to arm64) or an older npm, and we can sort those out without dropping `npm ci`.

Thanks again for taking the time — please do send more PRs! 🚀

PR #414, opened 2026-06-22
[codex] Fix whatsapp-web.js authenticating hang

aeyeio:
## Summary

Fixes `whatsapp-web.js` sessions that remain stuck at `authenticating` after the phone has already linked the device and WhatsApp Web is active.

## Issue Encountered

A live Docker Compose session reached WhatsApp Web successfully, but the OpenWA dashboard and API stayed at `authenticating`. DevTools inspection showed the WhatsApp page was loaded and connected, with account info available, while `whatsapp-web.js` did not reliably finish its own `ready` event path. In the failing path, OpenWA only persisted `ready` from the library's `ready` callback, so the dashboard could remain stale indefinitely even after the underlying WhatsApp runtime was usable.

There was also an existing documented workaround to pin a known-good WhatsApp Web version, but the older documented cached version was no longer available upstream. Leaving the adapter on auto-selected WhatsApp Web builds made first-run sessions vulnerable to this recurring compatibility break.

## Why This Fix

This patch fixes the problem at the adapter boundary where the state mismatch is introduced:

- Defaults `whatsapp-web.js` to a validated cached WhatsApp Web build instead of relying on auto-selection.
- Keeps explicit opt-outs with `WWEBJS_WEB_VERSION=latest`, `auto`, or `off` for operators who want upstream auto-selection.
- Adds a bounded post-auth reconciliation loop: after `authenticated`, if the runtime reports `CONNECTED` and `window.WWebJS` is injected, the adapter promotes itself to `READY` even if the library missed the `ready` event.
- Clears the reconciliation timer on ready/disconnect/failure/teardown so it does not leak or mutate stopped sessions.
- Updates docs and compose comments so operator guidance matches runtime behavior.

This preserves the normal `ready` event path and only reconciles when the page proves it is actually connected and the injected helper layer exists.

## Validation

- `npm test -- --runInBand src/engine/adapters/whatsapp-web-js.adapter.spec.ts`
- `npm run build`
- `docker compose --profile full build openwa-api`
- Recreated the live `openwa-api` container from the patched image.
- Started the existing local session and verified it transitioned from `authenticating` to `ready` with phone/push name populated.
- Verified `GET /api/sessions/:id/chats?limit=1` returned `200`.
- Verified `/sessions` rendered the session as `Connected` in the dashboard.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -93,10 +93,9 @@ services:
       - SESSION_DATA_PATH=/app/data/sessions
       - PUPPETEER_HEADLESS=${PUPPETEER_HEADLESS:-true}
       - PUPPETEER_ARGS=${PUPPETEER_ARGS:---no-sandbox,--disable-setuid-sandbox,--disable-dev-shm-usage,--disable-gpu}
-      # Pin the WhatsApp Web version if a session hangs at "authenticating" after scanning the QR
-      # (whatsapp-web.js 1.34.x can stall on an incompatible auto-selected WA-Web version, #251/#273).
-      # Empty = auto-select (default). Set WWEBJS_WEB_VERSION in .env to a known-good version — see
-      # docs/12-troubleshooting-faq.md. (Without this line the var in .env never reaches the container.)
+      # Optional WhatsApp Web version override. The adapter defaults to a validated cached version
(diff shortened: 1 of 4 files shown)


aeyeio:
Closing in favor of a new ready-for-review PR that follows the repository's existing branch, commit, and PR title conventions.

PR #415, opened 2026-06-22
fix(engine): reconcile whatsapp-web.js ready after auth

aeyeio:
## Summary

Fix `whatsapp-web.js` sessions that remain stuck at `authenticating` after the phone has already linked the device and WhatsApp Web is active.

## Issue encountered

A live Docker Compose session reached WhatsApp Web successfully, but OpenWA kept persisting the session as `authenticating`. DevTools showed the WhatsApp page was loaded and connected, with account info available, while the `whatsapp-web.js` `ready` event did not reliably complete. Because OpenWA only persisted `ready` from that callback, the API and dashboard could remain stuck indefinitely even though the underlying WhatsApp runtime was connected.

The documented version-pin workaround also referenced an older cached WhatsApp Web build that is no longer available upstream. The version pin must remain opt-in so an unset `WWEBJS_WEB_VERSION` preserves the existing whatsapp-web.js auto-selection behavior.

## Why this fix

The mismatch is introduced at the engine adapter boundary, so the fix lives there:

- keep `WWEBJS_WEB_VERSION` opt-in; unset, `latest`, `auto`, and `off` all use whatsapp-web.js auto-selection;
- after `authenticated`, run a bounded reconciliation probe while the adapter is still `AUTHENTICATING`;
- promote the adapter to `READY` only when `getState()` returns `WAState.CONNECTED`, `client.info.wid.user` is populated, and `window.WWebJS` is injected;
- re-check liveness after the awaited probe so `disconnected` or `auth_failure` during a tick cannot promote a dead session;
- ignore repeated `authenticated` events after `READY` and keep repeated pre-ready events to one timer chain;
- clear the reconciliation timer on ready, disconnect, auth failure, timeout, and teardown;
- update tests, docs, and Compose guidance to match the opt-in pin behavior.

This keeps the normal `ready` callback path intact and only reconciles when the runtime proves it is actually usable.

## Validation

- `npm test -- --runInBand src/engine/adapters/whatsapp-web-js.adapter.spec.ts`
- `npm run lint`
- `npm run build`
- Previously validated the reconciliation path against the local Docker Compose session: the stuck session transitioned from `authenticating` to `ready`, chats returned `200`, and the dashboard rendered it as connected.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -57,13 +57,12 @@ PUPPETEER_ARGS=--no-sandbox,--disable-setuid-sandbox,--disable-dev-shm-usage,--d
 # Required in the Docker image and on hosts without a bundled browser (e.g. Alpine/ARM).
 # PUPPETEER_EXECUTABLE_PATH=/usr/bin/chromium
 
-# Pin the WhatsApp Web client version. If sessions get stuck at "authenticating" and never
-# reach "ready" (a whatsapp-web.js version-sync hang), set this to a known-good version from
-# https://github.com/wppconnect-team/wa-version (browse the html/ folder). Leave unset (or
-# "latest") to let whatsapp-web.js auto-select. WWEBJS_WEB_VERSION_REMOTE_PATH overrides the
-# URL template (use {version} as the placeholder) if you self-host the version HTML.
-# 2.3000.1023204257 is confirmed to reach "ready" (incl. ARM64 / Raspberry Pi 5) — uncomment it:
-# WWEBJS_WEB_VERSION=2.3000.1023204257
+# Optional WhatsApp Web client version pin. Leave unset (or use "latest", "auto", or "off") to
+# let whatsapp-web.js auto-select. If sessions get stuck at "authenticating" after scanning the QR,
+# set this to a known-good version from https://github.com/wppconnect-team/wa-version (browse the
+# html/ folder). WWEBJS_WEB_VERSION_REMOTE_PATH overrides the URL template (use {version} as the
+# placeholder) if you self-host the version HTML.
+# WWEBJS_WEB_VERSION=2.3000.1040641150-alpha
 
 # Extend the initial boot/inject wait (milliseconds) before QR generation. On slow first boots
 # (WSL2 or low-resource containers) the default 30000ms can expire before WhatsApp Web finishes
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -49,10 +49,9 @@ services:
       - SESSION_DATA_PATH=/app/data/sessions
       - PUPPETEER_HEADLESS=true
       - PUPPETEER_ARGS=--no-sandbox,--disable-setuid-sandbox,--disable-dev-shm-usage,--disable-gpu
-      # Pin the WhatsApp Web version if a session hangs at "authenticating" after scanning the QR
-      # (whatsapp-web.js 1.34.x can stall on an incompatible auto-selected WA-Web version, #251/#273).
-      # Empty = auto-select (default). Set WWEBJS_WEB_VERSION in .env to a known-good version — see
-      # docs/12-troubleshooting-faq.md.
+      # Optional WhatsApp Web version override. Leave empty for whatsapp-web.js auto-selection.
+      # If a session hangs at "authenticating", set WWEBJS_WEB_VERSION to a known-good cached build
+      # from wppconnect-team/wa-version; latest|auto|off also forces auto-selection.
       - WWEBJS_WEB_VERSION=${WWEBJS_WEB_VERSION:-}
       - WWEBJS_WEB_VERSION_REMOTE_PATH=${WWEBJS_WEB_VERSION_REMOTE_PATH:-}
       # Raise whatsapp-web.js's first-boot init wait (default 30000ms) on slow boots. Empty = default.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -93,10 +93,9 @@ services:
       - SESSION_DATA_PATH=/app/data/sessions
       - PUPPETEER_HEADLESS=${PUPPETEER_HEADLESS:-true}
       - PUPPETEER_ARGS=${PUPPETEER_ARGS:---no-sandbox,--disable-setuid-sandbox,--disable-dev-shm-usage,--disable-gpu}
-      # Pin the WhatsApp Web version if a session hangs at "authenticating" after scanning the QR
-      # (whatsapp-web.js 1.34.x can stall on an incompatible auto-selected WA-Web version, #251/#273).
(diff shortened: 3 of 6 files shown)


rmyndharis (changes requested):
Thanks for this — the runtime-readiness probe is a good idea, and it mirrors whatsapp-web.js's own internal readiness signal (`window.WWebJS`, `Client.js:307-335`), so the approach is sound. I traced the change against the engine source and our session lifecycle. A CI blocker and a few correctness issues to resolve before merge.

## Blocker (CI lint)

`eslint` fails on the adapter:

```
src/engine/adapters/whatsapp-web-js.adapter.ts:501  error  @typescript-eslint/no-unsafe-enum-comparison
```

`getState()` returns the `WAState` enum, not a string. Compare against the enum member:

```ts
import { Client, LocalAuth, MessageMedia, MessageTypes, WAState } from 'whatsapp-web.js';
// ...
if (state !== WAState.CONNECTED) return false;
```

## Correctness

### 1. An in-flight probe tick can revive a disconnected session to READY

`on('disconnected')` clears the timer and sets `DISCONNECTED`, but does **not** null `this.client`. If a tick is already awaiting `pupPage.evaluate()` when `disconnected` fires, the page is still open, the evaluate resolves `true`, and `markReadyFromClientInfo()` runs. Its only guard is `if (this.status === EngineStatus.READY) return`, which does not catch `DISCONNECTED` — so it flips the engine back to READY and fires `onReady`. Downstream that persists `status=READY` and dispatches `session.authenticated` on a session that is actually tearing down / reconnecting, after which every send throws `EngineNotReadyError`. `clearReadyReconcile()` only cancels the *pending* timer, not a tick already past the top guard.

### 2. The probe can promote READY with an empty phone/pushName, and it sticks

The gate is `getState()==='CONNECTED' && window.WWebJS!==undefined` — it does not require `client.info`. In whatsapp-web.js `window.WWebJS` is defined (`Client.js:327`) **before** `this.info` is assigned (`Client.js:349`, the only assignment). A tick landing in that window reads `info?.wid?.user` → `null` → `onReady('', '')`, which we persist as `phone=''` / `pushName=''`. The `status===READY` guard then suppresses the genuine `ready` event that would have corrected it, so the empty identity persists until a manual reconnect.

**Both #1 and #2** are addressed by requiring populated identity in the gate and re-asserting liveness *after* the await:

```ts
private async isClientRuntimeReady(): Promise<boolean> {
  if (!this.client) return false;
  if (await this.client.getState() !== WAState.CONNECTED) return false;
  if (!this.client.info?.wid?.user) return false; // info not populated yet -> keep waiting
  const page = (this.client as unknown as { pupPage?: { evaluate: <T>(fn: () => T) => Promise<T> } }).pupPage;
  return (await page?.evaluate(() => typeof (window as unknown as { WWebJS?: unknown }).WWebJS !== 'undefined')) === true;
}
```

```ts
// inside the tick, after the await:
if (await this.isClientRuntimeReady()) {
  if (!this.client || this.status !== EngineStatus.AUTHENTICATING) {
    this.clearReadyReconcile();
    return;
  }
  this.logger.warn('WhatsApp Web ready event was missed; reconciling from connected runtime state');
  this.markReadyFromClientInfo();
  return;
}
```

The probe is best-effort recovery, so demanding real `info` costs nothing on the happy path.

### 3. A repeated `authenticated` after READY causes a status flap + duplicate webhook

whatsapp-web.js re-emits `authenticated` on app-state re-sync. The handler unconditionally drops `READY → AUTHENTICATING` and re-arms the probe, producing a second `onReady` → a duplicate `session.authenticated` webhook + DB write + reconnect-counter reset (the `status===READY` dedup guard is defeated because the status was lowered first). Guard the handler:

```ts
this.client.on('authenticated', () => {
  if (this.status === EngineStatus.READY) return;
  this.setStatus(EngineStatus.AUTHENTICATING);
  this.qrCode = null;
  this.scheduleReadyReconcile();
});
```

## Tests

The new spec covers only the happy path. With the fixes above, please add coverage for:
- empty `client.info` while CONNECTED + WWebJS injected → must **not** promote with an empty phone;
- the `status===READY` dedup → emit `ready` after the reconcile promotes, assert `onReady` still fires exactly once;
- `disconnected` / `auth_failure` mid-probe → no promotion, timer cleared (`jest.getTimerCount() === 0`);
- double `authenticated` → a single timer chain (one `onReady`).

## Minor (docs drift)

`docker-compose.yml` and `docs/12-troubleshooting-faq.md` are updated, but `.env.example` (~lines 60-66) and `docker-compose.dev.yml` still say "unset = auto-select", recommend `2.3000.1023204257` (now returns 404), and omit the new `auto` opt-out keyword.

## Note on the default version pin (no action needed from you yet)

Flipping `WWEBJS_WEB_VERSION` from opt-in to default-on — pinning an `-alpha` build fetched from raw.githubusercontent.com at every boot — is a behavior change we're evaluating on our side: it adds a boot-time network dependency for every wwebjs session, and the pinned file rotates over time (the previous `2.3000.1023204257` already 404s). We'll follow up with a direction on that; the probe itself is the part we want to keep.

Thanks again for digging into this one — it's a real and long-standing pain point.

rmyndharis:
Following up on the version-pin note above, here's our direction:

We'd like to keep `WWEBJS_WEB_VERSION` **opt-in** (unset = whatsapp-web.js auto-select, as before) and rely on the new reconciliation probe as the actual fix. Reasoning: the probe addresses the reported symptom directly — a missed `ready` on a runtime that's already connected with account info — whereas a default-on remote pin makes every wwebjs session fetch from raw.githubusercontent.com at boot, and that pinned file rotates over time (the previous `2.3000.1023204257` already 404s). Probe + opt-in pin is strictly better than the pre-PR behavior, without the new boot-time egress dependency.

Concretely, that means reverting `resolveWebVersionPin()` to its opt-in shape (drop the default-on flip and `DEFAULT_WWEBJS_WEB_VERSION`) while keeping everything else — the probe, the `auto` opt-out keyword, and the docs/compose clarifications are all welcome to stay.

No problem either way on who does that slice: happy to take the version-pin revert on our side so you can keep this PR focused on the probe + the correctness fixes, or fold it in here if you prefer. Let us know which you'd rather.

rmyndharis (approved):
Deep source-level re-review (state machine + downstream callsites + 5-dimension adversarial pass, each finding verified). The revision resolves the prior CHANGES_REQUESTED, and I folded in fixes for the 4 edge cases surfaced (timeout-safe probe, single reconcile window, teardown-safe authenticated guard) with TDD tests. Verified: 66 adapter tests; full backend gate; clean merge onto current main (1229 jest); fold-in independently re-checked (0 new bugs). The only CI red is the pre-existing arm64 lightningcss Docker-build issue on main (push-only job, unrelated to this PR).

PR #433, opened 2026-06-22
feat(plugins): install from URL + remote catalog (marketplace MVP)

rmyndharis:
## What

Adds a VS Code-style **catalog + one-click install** to the plugin system, fed by a remote
`plugins.json` (the [OpenWA-plugins](https://github.com/rmyndharis/OpenWA-plugins) repo). Implements the
spec in [OPENWA-MARKETPLACE.md](https://github.com/rmyndharis/OpenWA-plugins/blob/main/OPENWA-MARKETPLACE.md).

Today the only way to install is uploading a `.zip`. This adds:

- **`POST /plugins/install-url`** (ADMIN) — downloads a plugin `.zip` from an HTTP(S) URL through the
  existing SSRF guard (`withSafeFetch`: host validated, connection pinned to the vetted IP, redirects
  refused, size-capped), then hands the buffer to the **existing** `parsePluginPackage` + write +
  `loadPlugin` pipeline. The download is treated exactly like an upload — same zip-slip/size/manifest/
  reserved-id validation, same sandbox.
- **`GET /plugins/catalog`** (ADMIN) — fetches a configured remote catalog (`PLUGIN_CATALOG_URL`,
  default the OpenWA-plugins `plugins.json`) and annotates each entry with `installed`,
  `installedVersion`, and `updateAvailable` (semver compare vs what's loaded).
- **Dashboard:** the install modal gains **Upload / Catalog** tabs; the Catalog tab lists entries with
  one-click Install and an "update available" badge.
- **Config:** `plugins.{dir,catalogUrl,downloadMaxBytes}`. Operators add a non-public catalog/release
  host to `SSRF_ALLOWED_HOSTS`.

## Security

- ADMIN-only, like every other plugin route.
- Every download goes through `withSafeFetch` (the same guard as media/webhook fetches) — no
  unguarded fetch of a user-supplied URL.
- The downloaded package still passes the full `parsePluginPackage` validation and runs sandboxed.

## Testing

- New unit tests for `compareSemver` + `annotateCatalog` (`catalog.spec.ts`).
- `plugins.service.spec.ts` updated for the new `ConfigService` dependency.
- `jest src/modules/plugins` → 37/37 pass; `nest build` clean; backend eslint clean.
- Dashboard `tsc -b`, `eslint`, and `i18n:check` (locale parity) all pass. New UI strings use inline
  English defaults (`t(key, 'default')`), so they render in every locale; proper translations can be
  added by the usual i18n pass.

## Deferred (follow-up)

One-click **in-place Update** is not included: `install()` rejects an already-installed id, and a safe
in-place update (preserving operator config across the new version) is a separate feature. The catalog
surfaces "update available" today; the action can land once update-preserving-config exists.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Install plugins from a URL / catalog.** `POST /plugins/install-url` downloads a plugin `.zip` from an HTTP(S) URL through the SSRF guard (host validated, connection pinned, redirects refused, size-capped) and runs the exact same validate-write-load pipeline as an uploaded package. `GET /plugins/catalog` fetches a configured remote catalog (`PLUGIN_CATALOG_URL`, default the OpenWA-plugins `plugins.json`) and annotates each entry with `installed` / `installedVersion` / `updateAvailable`. The dashboard install modal gains a **Catalog** tab to browse and one-click install. Add a non-public catalog/release host to `SSRF_ALLOWED_HOSTS`.
+
 ### Fixed
 
 - Sandboxed plugins now receive `onConfigChange` (config updates reach the worker instead of being silently ignored until disable + re-enable) and have their real `healthCheck` run — `GET /plugins/:id/health` previously always returned the default "healthy" for sandboxed plugins. (#430)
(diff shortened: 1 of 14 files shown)


rmyndharis:
Added one-click **in-place Update** (the path noted as deferred above): `POST /plugins/:id/update` downloads the new package via the same SSRF-guarded path and swaps it in while **preserving operator config and the enabled state** (unload keeps the registry entry → config survives → reload → re-enable). Package id must match; old version is backed up and restored on failure. The Catalog tab now shows an **Update** button for entries with a newer version. 40/40 plugin tests, `nest build`, and dashboard tsc/eslint/i18n all pass.

PR #450, opened 2026-06-23
fix(compose): make dev compose environment variables overridable

MS-Jahan:
## Summary

`docker-compose.dev.yml` hardcoded most environment variables to fixed literal values (e.g. `NODE_ENV=development`, `DATABASE_TYPE=sqlite`, `PUPPETEER_HEADLESS=true`, `WEBHOOK_TIMEOUT=10000`, ...). There was no way to override them from the shell environment or an `.env` file without editing the compose file.

This converts the **user-facing** settings to the `${VAR:-default}` pattern — same defaults as before, but now overridable — bringing the dev compose in line with the production `docker-compose.yml`.

### Made overridable (default unchanged)
`NODE_ENV`, `DATABASE_TYPE`, `DATABASE_NAME`, `DATABASE_SYNCHRONIZE`, `ENGINE_TYPE`, `PUPPETEER_HEADLESS`, `PUPPETEER_ARGS`, `STORAGE_TYPE`, `WEBHOOK_TIMEOUT`, `WEBHOOK_MAX_RETRIES`, `QUEUE_ENABLED`, plus the data-path settings documented in `.env.example`: `SESSION_DATA_PATH`, `STORAGE_LOCAL_PATH`. Also added `LOG_LEVEL=${LOG_LEVEL:-info}` for parity with production.

The data-path defaults stay under `/app/data` (the `./data` bind mount), so persistence is unchanged out of the box; override only if you mount that target elsewhere.

### Intentionally kept fixed
`PORT`, `HOME`, `XDG_CONFIG_HOME`, `XDG_CACHE_HOME` — container-internal values that must match the image/entrypoint. Production keeps these fixed for the same reason.

## Notes

- `WWEBJS_*` were already parameterized; unchanged.
- Defaults are identical to the previous literals, so existing `docker compose -f docker-compose.dev.yml up` behavior is unchanged when no env is set.

## Test plan

- [x] `docker compose -f docker-compose.dev.yml config` parses successfully
- [ ] `docker compose -f docker-compose.dev.yml up` with no env uses the prior defaults
- [ ] Setting e.g. `LOG_LEVEL=debug` / `PUPPETEER_HEADLESS=false` in `.env` overrides the defaults

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -34,21 +34,27 @@ services:
     ports:
       # Bind to localhost by default; set BIND_HOST=0.0.0.0 in .env to reach it from another host.
       - '${BIND_HOST:-127.0.0.1}:2785:2785'
+    # User-facing settings read from the environment (or .env) with a sane default fallback
+    # (${VAR:-default}); override any of them without editing this file. The data paths default
+    # under /app/data (the ./data bind mount) so they persist out of the box — override only if you
+    # also mount that target. Truly container-internal values (HOME, XDG_*, PORT) stay fixed: they
+    # must match the image/entrypoint. Same convention as the production docker-compose.yml.
     environment:
-      - NODE_ENV=development
+      - NODE_ENV=${NODE_ENV:-development}
       - PORT=2785
       - HOME=/tmp
       # Chromium reads its home from the passwd entry (no /home/openwa), so it needs writable, existing
       # config/cache dirs on the tmpfs or it hard-crashes at launch; the entrypoint pre-creates them. (#254)
       - XDG_CONFIG_HOME=/tmp/.config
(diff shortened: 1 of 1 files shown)


rmyndharis:
Incorporated into **v0.7.1** (commit 96b4998) with one reconciliation: the `ENGINE_TYPE` line was dropped. v0.7.1 moves engine selection to the dashboard (Infrastructure → Engine); since `main.ts` loads `.env.generated` with `override:false`, pinning `ENGINE_TYPE` as a container env would silently override the dashboard's choice. Your overridable pattern for every other var is in, and `PLUGINS_DIR` was preserved (+ made overridable). Thanks @MS-Jahan! 🙏

PR #453, opened 2026-06-23
fix(engine): clear Baileys auth state on logout so re-linking shows a fresh QR

ulises2k:
## Problem

When a Baileys session is terminated by WhatsApp (`DisconnectReason.loggedOut` — e.g. the device is unlinked from the phone) — or via the API `logout()` — `BaileysAdapter` leaves the multi-file auth directory (`authDir/sessionId`) on disk.

On the next `connect()`, `useMultiFileAuthState(this.authPath)` reloads those now-invalid credentials (`creds.json` with `registered: false`). Baileys then retries the dead identity instead of emitting a fresh QR, so the session gets stuck: the status churns but **no QR is ever produced**, and re-linking is impossible without manually deleting the auth folder.

This is exactly what the existing TODO in `logout()` anticipated:

> `// leaves the multi-file auth dir on disk; a fresh link overwrites it. Add fs cleanup if stale creds ever block re-linking.`

In practice a fresh link does **not** overwrite stale creds in the logged-out case — they are reloaded and reused.

## Fix

A small `clearAuthState()` helper removes `authDir/sessionId` (`fs.promises.rm`, `recursive: true, force: true`), called in the two terminal paths:

1. `handleConnectionUpdate` — on `DisconnectReason.loggedOut` (and null out the now-dead socket).
2. `logout()` — replacing the TODO.

`force: true` makes a missing dir a no-op; failures are logged and swallowed (best-effort cleanup must not throw on a teardown path). `disconnect()` / `destroy()` are intentionally untouched — those are non-terminal and must keep the session for reconnect. Only the session's own dir (`authDir/sessionId`) is removed, never the shared `authDir`.

## Tests

Two new cases in `baileys.adapter.spec.ts`:
- a `loggedOut` close wipes `authDir/sessionId`;
- `logout()` wipes `authDir/sessionId`.

`baileys.adapter.spec.ts` green (90/90); `npm run build` clean.

### DIFF
--- a/src/engine/adapters/baileys.adapter.spec.ts
+++ b/src/engine/adapters/baileys.adapter.spec.ts
@@ -1,4 +1,6 @@
 import { EventEmitter } from 'events';
+import * as fs from 'fs';
+import * as path from 'path';
 
 jest.mock('../../common/media/load-remote-media', () => ({
   loadRemoteMediaBuffer: jest.fn(),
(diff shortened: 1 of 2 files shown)


ulises2k:
Thanks so much for merging this so quickly, @rmyndharis — much appreciated! 🙏

One question that came up while upgrading to v0.7.1: the bundled compose files no longer pin `ENGINE_TYPE`, and the active engine is now selected via the dashboard / `data/.env.generated`, which **defaults to `whatsapp-web.js`**. Combined with v0.7.1 dropping `env_file`, a Baileys deployment that relied on `ENGINE_TYPE=baileys` in its `.env` (feeding the old `${ENGINE_TYPE:-whatsapp-web.js}` interpolation) silently boots `whatsapp-web.js` after upgrading, unless a *real* container/host `ENGINE_TYPE=baileys` env is set. It caught me on my Raspberry Pi — I had to add `ENGINE_TYPE=baileys` back into my compose override.

Was that change intentional, and is there a reason the selection defaults to `whatsapp-web.js` rather than Baileys (which is lighter / needs no Chromium)? Mostly want to understand the reasoning so I can document the upgrade path correctly for Baileys users. Thanks again!

PR #461, opened 2026-06-24
feat(mcp): expose the REST API as an opt-in MCP server

tobiasstrebitzer:
## Summary

Adds an **opt-in MCP (Model Context Protocol) server** so AI agents (Claude, Cursor, etc.) can drive WhatsApp as first-class tools, reusing the existing REST controllers. It is **off by default** and fully **additive**: every existing route keeps serving REST unchanged.

Set `MCP_ENABLED=true` to mount a stateless Streamable-HTTP transport at `POST /mcp` on the existing server (same port, no extra process). When `MCP_ENABLED` is unset, none of the MCP code loads.

## How it works

- A small in-house adapter under `src/modules/mcp/`, built directly on the official [`@modelcontextprotocol/sdk`](https://github.com/modelcontextprotocol/typescript-sdk) (no third-party MCP framework dependency).
- A route opts in by adding a single `@Mcp()` decorator next to its existing `@Get`/`@Post`. Around 126 routes are exposed as tools; tool name is `{Controller}{Method}` (e.g. `SessionFindOne`). Only applied if MCP_ENABLED=true, otherwise ignored as pure metadata.
- Tool input schemas are reflected from `@Param`/`@Query`/`@Body` + `class-validator` + `@nestjs/swagger`, so the schema stays in sync with the REST contract.
- On a call, the validated input is re-bound to the controller method and it runs directly.

## Auth (fully enforced over MCP)

Every tool call goes through the same `ApiKeyGuard` + `@RequireRole` + per-key `allowedSessions` scoping as REST, so a key scoped to one session cannot act on another. The throttler is deliberately excluded (it needs a writable response the MCP transport does not provide). There is no client IP over MCP, so IP-allowlisted keys will not authorize MCP calls. Covered by unit tests and a dedicated `test/mcp-auth.e2e-spec.ts`.

## Also in this PR

Sharpened the session-id parameter descriptions across all controllers from `Session ID` to `Session UUID (the session id, not the name)`. Sessions have both an `id` (UUID) and a human `name`; the vague "Session ID" led agents to pass the name, which silently does not resolve. This also improves the REST Swagger docs.

## How to test

1. Build and run with MCP enabled:

   ```bash
   npm run build
   MCP_ENABLED=true PORT=2785 node dist/main
   ```

2. Point an MCP client at it. For Claude Code, drop a `.mcp.json` at the repo root (it is gitignored; replace the key with a real API key from `data/.api-key`, default `dev-admin-key`):

   ```json
   {
     "mcpServers": {
       "openwa": {
         "type": "http",
         "url": "http://localhost:2785/mcp",
         "headers": {
           "Authorization": "Bearer dev-admin-key"
         }
       }
     }
   }
   ```

   The same key can be passed as `X-API-Key` instead of `Authorization: Bearer`.

3. From the client, list tools (you should see ~126, e.g. `SessionFindAll`, `MessageSendText`) and call one, for example `SessionFindAll`, to confirm auth and reflection work end to end.

4. Sanity-check that REST is untouched and MCP is inert when disabled: start without `MCP_ENABLED` and confirm `POST /mcp` is not mounted while `/api/...` behaves as before.

## Notes

- Unit + e2e suites pass; `npm run build` resolves the reflected `@Mcp` types.
- No new runtime behavior when `MCP_ENABLED` is unset (the package is only required on boot when MCP is enabled).


tobiasstrebitzer:
@rmyndharis following up on the discussion in #256, this PR is the in-house-SDK rewrite we agreed on there. Rather than re-explain from scratch, here is a precise mapping of each piece of feedback from that thread to how this PR addresses it.

## The core reason #256 was declined

> It adds `@silkweave/*` to **production** dependencies (~37 transitive packages incl. a second HTTP framework and a second logger), and because every controller imports `@Mcp()` at module load, **the framework loads on every boot regardless of `MCP_ENABLED`**.

Both halves are resolved:

- **No `@silkweave/*` dependency.** This PR depends directly on the official `@modelcontextprotocol/sdk` behind a thin in-house adapter (`src/modules/mcp/`, ~14 small files using Nest's `DiscoveryService`). The only new dependencies are `@modelcontextprotocol/sdk` and `zod`. `pino` and the silkweave CLI extras are gone. (`hono`/`express` arrive from the SDK itself either way)
- **Nothing MCP-related loads on a default boot.** `@Mcp()` is a pure `SetMetadata` marker that imports only `@nestjs/common` (plus a type-only `zod` import that erases at compile). The SDK is imported in exactly one file (`mcp.server.ts`), which is only reachable through `mcp.module.ts`, which `app.module.ts` `require()`s **only when `MCP_ENABLED=true`** (same lazy pattern as `QUEUE_ENABLED`). With the flag unset there is no `/mcp` route and the SDK is never loaded.

## The "clean merge" checklist

| Requirement | How this PR meets it |
|---|---|
| **1. `@Mcp()` stays a pure `SetMetadata` marker, discovery via `DiscoveryService`** | `mcp.decorator.ts` is `SetMetadata(MCP_METADATA, options)` and nothing else. Tools are discovered at boot via `DiscoveryService` in `discovery.ts`. |
| **2. Truly off by default - gate the module, not just the route** | `app.module.ts` env-gates the entire `require()` + `forRoot()`. A default boot instantiates no module, no transport, no SDK server code. |
| **3. Mount on the existing server / single port** | The transport attaches to the existing Nest/Express adapter via `adapter.post('/mcp', ...)` in `mcp.server.ts`. No second port, no `@hono/node-server`. |
| **4. Auth is the acceptance bar - every call through `ApiKeyGuard` + `@RequireRole` + per-key session scoping, test-covered** | Every tool call runs the guards before the controller method (`discovery.ts`). Covered by unit specs plus an e2e (`test/mcp-auth.e2e-spec.ts`) that proves the cross-session-denial case. Details below. |
| **5. `src/modules/mcp/` core module, REST untouched, README/CHANGELOG note** | Module lives at `src/modules/mcp/`. Controllers only gain a `@Mcp()` decorator (additive); no route, verb, path, DTO, or service changes. README has an "MCP" section and CHANGELOG has an `[Unreleased]` entry. |

## The two follow-up flags

**Stateless, single port.** The transport is a stateless Streamable-HTTP handler: `new StreamableHTTPServerTransport({ sessionIdGenerator: undefined })`, built fresh per request and torn down, with no SSE/session affinity. This lines up with the 2026-07-28 stateless spec direction and removes the single-port friction.

**The auth crux: the path-scoped `sessionId`.** This is the part flagged as most sensitive, so it is worth being explicit. On a tool call the adapter routes the validated input into `request.params`/`query`/`body` (only filling absent keys) **before** the guards run, so `ApiKeyGuard` reads `request.params.sessionId` (or `request.params.id` on `@SessionScoped` controllers) exactly as it does over REST. The guard makes the same decision it would on the equivalent REST route, so per-key `allowedSessions` scoping holds identically.

`test/mcp-auth.e2e-spec.ts` drives the real `ApiKeyGuard` + `AuthService` + sqlite and asserts:

- no key and invalid key are both rejected
- a viewer key scoped to `session-a` is allowed on `session-a`
- **the same scoped key is denied on `session-b`** (the cross-session-leak path we wanted nailed down)
- `@RequireRole` is enforced (a viewer key is blocked on an operator tool)
- an unscoped admin key reaches any session

The fixture intentionally models the harder `@SessionScoped` + `:id` shape, since that is the one that depends on the reflector reading through the MCP execution context.

## Two intentional, documented caveats

- The throttler is **not** run over MCP (it needs a writable HTTP response the transport does not provide). The auth guards are unaffected.
- IP-allowlisted keys cannot authorize MCP calls, because there is no client IP over the MCP transport. Keys without an IP restriction are unaffected.

## Scope

~126 tool-shaped routes opt in across the existing controllers. Deliberately not exposed: the Prometheus `metrics` text endpoint, the login-only `POST /api/auth/validate`, and file-upload / HTML routes that are not tool-shaped. REST is byte-for-byte unchanged.

Happy to adjust the tool surface, the auth-scoping shape, or anything else. Thanks again for the detailed direction in #256.

rmyndharis:
Wow — thank you for this, @tobiasstrebitzer. 🙌 This is a seriously impressive piece of work. Reflecting the whole REST surface into MCP tools, keeping it off-by-default on a single port, and wiring the auth replay is genuinely clever engineering, and the polish shows: the result-shaping in `tool-result.ts`, the lazy gating/mount, and the "Session UUID (not the name)" clarification are all excellent. Honestly, this PR is what made us take MCP seriously and sit down to design it properly — that's a real contribution on its own, and it's appreciated.

Because it pushed us to think it through end to end, we've landed on an architectural direction we'd love to build *with* you, and I want to be open about it:

1. **A curated, semantic set of tools rather than a 1:1 reflection of all ~126 routes.** This is where the ecosystem has converged too (GitHub's MCP server, FastMCP, Speakeasy all curate + tier instead of auto-exposing everything) — a focused, well-described, read/write-tiered surface keeps agents from getting overwhelmed and keeps the riskier operations off the default path.
2. **Running each tool through the existing service + auth/validation layer**, so the agent path stays perfectly in lockstep with REST as the app evolves, with nothing to keep manually in sync.

On the practical side, it also sits on a base from before #458–#460 so it'd need a rebase regardless — which gave us a natural moment to align on direction before anyone redoes that work.

Crucially, **none of your work is going to waste**: we want to carry over your `tool-result.ts` and the gating/mount approach, with credit to you in the commits. And I'd genuinely love to have you involved in the curated version — I'm happy to share the registry/adapter shape we're planning and would welcome your contributions on it. Would you rather we keep this PR open as the discussion home, or spin up a fresh tracking issue for the direction?

Either way — thank you for pushing this forward. 🙏

rmyndharis:
Honest update, @tobiasstrebitzer — and thank you again. To move the architecture we aligned on forward, we've implemented the curated, service-layer version in-house, and it's now on `main` (behind `MCP_ENABLED`, off by default, in `[Unreleased]`). The full design is written up publicly here: [docs/24-mcp-integration.md](https://github.com/rmyndharis/OpenWA/blob/main/docs/24-mcp-integration.md).

Concrete credit: your result-shaping (`tool-result.ts`) is carried over with attribution in the commit and CHANGELOG, and the off-by-default single-port gating/mount shape follows what you established here. This PR is genuinely what got MCP designed properly — that's a real contribution.

What it means: the reflection/re-bind approach in this PR won't be merged as-is — the shipped version is the curated tool surface (~39 tools that run through the existing services + auth, rather than reflecting all routes). That's a direction decision, not a reflection on your work, which was strong.

I'd love to keep you involved — a review of the merged code is very welcome, and two deferred pieces are great to own if you're interested: **agent-action audit provenance** and **env-tunable rate limiting**. I'll leave this PR open for now so you can weigh in; I'm happy to close it as superseded (with thanks) once you've had a chance to respond, or keep it open if you'd rather adapt it.

Thanks again for pushing this forward. 🙏

tobiasstrebitzer:
Hi @rmyndharis, thanks for all the effort on this. I've had a look at the implementation and fully agree with the direction you've taken. Your work also gave me some immediate ideas and considerations I can bring back into silkweave.

Specifically:
- Clear separation of read/write permissions is a great idea with obvious use-cases.
- While curated tools require more maintenance, they clearly benefit agent interaction.
- Rate limiting also helps keep the integration secure and avoid trouble with WhatsApp.

I'll be using the new MCP server over the next few weeks to get some field experience.

Happy to keep contributing on this. Will review the merged code more thoroughly, and aim to take a stab at the expansion-pack tool domains and OAuth 2.1 design/scoping soon.

And yes, please go ahead and close this PR as superseded.

rmyndharis:
Thanks so much, @tobiasstrebitzer — really glad the direction resonates, and thank you for being gracious about it. 🙏

As promised, your work is credited where it landed: the result-shaping helper (`tool-result.ts`) and the off-by-default, single-port gating/mount shape both carried over, with attribution in the commit (`643e76b`) and the CHANGELOG — which just shipped in **v0.7.3**.

Closing this as **superseded** per your note. The door's wide open for the pieces you mentioned wanting to take on — the **expansion-pack tool domains** and the **OAuth 2.1 design/scoping**, plus the two deferred items (**agent-action audit provenance** and **env-tunable rate limiting**). Happy to share the registry/adapter shape whenever you want to dig in, and keen to hear field notes once you've run the MCP server for a bit.

Thanks again for pushing MCP forward here. 🙌

PR #470, opened 2026-06-25
feat(logger): add human-readable logging format and enhance LoggerSer…

m7fz7:
## Description

Application logs from the project's `LoggerService` were always emitted as raw JSON, so they looked nothing like NestJS's own framework lines (`[Nest] … LOG [DockerService]`) printed right next to them — the console was a jarring mix of one-line JSON blobs and pretty Nest output.

This adds a human-readable, colorized formatter to `LoggerService` that mirrors NestJS's layout (`[OpenWA] <pid> - <timestamp> <LEVEL> [Context] <message>`), with structured fields shown as dimmed `key=value` trailers and error stack traces on their own line.

- **Default format:** pretty everywhere, **structured JSON only when `NODE_ENV=production`** (containers, log aggregators). This mirrors NestJS's logger and — unlike a TTY check — stays correct when stdout is piped through a dev runner (e.g. `concurrently` in `npm run dev`).
- **Override:** `LOG_FORMAT=pretty|json` always wins.
- **Colors:** gated on `FORCE_COLOR` / `NO_COLOR` / TTY.
- JSON output is byte-for-byte unchanged when selected, so existing log-aggregation consumers are unaffected.

The prefix is `[OpenWA]` (not `[Nest]`) so the lines are honestly attributed rather than impersonating the framework.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist
- [x] Tests added/updated
- [x] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

## Example

**Before** — raw JSON mixed with NestJS lines:

{"timestamp":"2026-06-25T14:00:25.465Z","level":"info","context":"AuthService","message":"Welcome to OpenWA - WhatsApp API Gateway"}
[Nest] 10888  - 25/06/2026, 6:00:25 PM     WARN [DockerService] Docker not available. Container orchestration disabled.

**After** — uniform, colorized output:

[OpenWA] 12216  - 25/06/2026, 6:13:25 PM     LOG [AuthService] Welcome to OpenWA - WhatsApp API Gateway
[OpenWA] 12216  - 25/06/2026, 6:13:25 PM     LOG [PluginLoaderService] WhatsApp-web.js engine plugin loaded pluginId=whatsapp-web.js
[OpenWA] 12216  - 25/06/2026, 6:13:25 PM    WARN [DockerService] Docker not available. Container orchestration disabled.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,6 +10,9 @@
 NODE_ENV=production
 API_PORT=2785
 LOG_LEVEL=info                      # error | warn | info | debug
+# Console output format. Default: json in production (containers, log aggregators),
+# human-readable pretty otherwise. Force one with: json | pretty
+# LOG_FORMAT=pretty
 
 # Auto-start previously authenticated sessions on server boot
 AUTO_START_SESSIONS=false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Human-readable console logs: the `LoggerService` now renders a colorized, NestJS-style line (`[OpenWA] <pid> - <timestamp> <LEVEL> [Context] <message>` with dimmed `key=value` metadata and stack traces on their own line) instead of always emitting raw JSON, so application logs line up visually with NestJS's own framework logs. The format defaults to structured JSON in production (`NODE_ENV=production`, for containers and log aggregators) and human-readable pretty everywhere else, and can be pinned with `LOG_FORMAT=pretty|json`. `NO_COLOR` / `FORCE_COLOR` are honored. JSON output is byte-for-byte unchanged when selected. (#469)
+
 ## [0.7.4] - 2026-06-25
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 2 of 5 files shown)


rmyndharis (commented):
Thanks for this — it's a clean, well-tested improvement, and I appreciate the care in the details: native ANSI (no new dependency), a `NODE_ENV`-based default so production stays JSON for aggregators, the `LOG_FORMAT` override, `NO_COLOR`/`FORCE_COLOR` honored, stack traces on their own line, and the honest `[OpenWA]` prefix rather than impersonating `[Nest]`. Tests, `.env.example`, and CHANGELOG are all updated. 👍

Two things before it can merge:

1. **Revert the `package-lock.json` change.** The feature adds no dependency (you used native ANSI codes), so the lockfile churn (+33 −45) looks like incidental `npm install` drift. Please restore it to `main`'s version so the diff is just the logger change — unrelated lockfile shifts are a conflict and regression source.
2. **Rebase onto `main`.** It's currently CONFLICTING — `CHANGELOG.md`'s `[Unreleased]` has moved (v0.7.4 shipped and a couple of entries landed). A quick rebase and re-placing your `### Added` note under the new `[Unreleased]` clears it.

Optional nit (not blocking): `colorEnabled()` checks `process.stdout.isTTY` even for the `console.error`/`warn` paths, which write to stderr — negligible in practice, fine to leave.

Once the lockfile is reverted and it's rebased, this is good to merge. Thanks again for a thoughtful PR!

m7fz7:
Thanks for the thorough review and the kind words! Both items are addressed:

**1. Lockfile reverted.** You're right, it was incidental `npm install` drift, no dependency was added (native ANSI throughout). I've restored `package-lock.json` to match main exactly, so it no longer appears in the diff. The change is now just the four files that belong to the feature:

- `.env.example`
- `CHANGELOG.md`
- `src/common/services/logger.service.ts`
- `src/common/services/logger.service.spec.ts`

**2. Rebased onto main.** The branch had picked up a merge commit, which is what made it conflict. I dropped the merge and rebased the feature cleanly onto the v0.7.4 tip, so history is linear again. The CHANGELOG conflict is resolved: the `### Added` logger note now sits under the new (post-v0.7.4) `[Unreleased]`, with the shipped 0.7.4 entries intact below it.

Logger spec still green (10/10).

On the optional nit: good catch. `colorEnabled()` keys off `process.stdout.isTTY` even for the `error`/`warn` paths that write to stderr. Agreed it's negligible in practice, so I've left it as-is for this PR; happy to follow up separately if you'd prefer it check the matching stream.

Thanks again!

PR #482, opened 2026-06-26
update

Offjedd:
## Description
Brief description of changes

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] Lint passes
- [ ] Self-reviewed

## Related Issues
Closes #


Offjedd:
a

rmyndharis:
Thank you for sharing this — there's clearly a lot of real work here, and the Agency OS + customer portal is an impressive build.

I want to be honest about fit, and explain the reasoning so it's useful rather than just a "no."

OpenWA is intentionally a **single-tenant, self-hosted gateway** with one backend (NestJS + TypeORM) and a single API-key auth model. This PR is effectively a different product built **on top of** that idea:

- It introduces a second backend — Supabase (Postgres + RLS + Deno edge functions) — as the actual persistence/compute layer. The new `src/modules/customer/*` writes exclusively to Supabase, so it doesn't run inside OpenWA's own deployment model.
- It adds multi-tenancy (agencies / sub-accounts / per-customer signup) and a parallel customer-JWT auth system alongside the API-key model.
- It pulls Supabase / bcrypt / JWT into core's dependency tree, plus Netlify deploy config and a couple of committed image assets — ~17k lines, no tests.

The natural home for this is **your own product/fork that consumes OpenWA through its public REST + webhook API**. That keeps you free to move fast on the SaaS side while OpenWA stays a small, focused gateway, and you depend on a released OpenWA rather than vendoring and maintaining a divergence.

One thing I'd genuinely urge regardless of upstreaming: before running this anywhere near production, please get a security review of the Supabase **RLS policies** and the **edge-function authentication** — a few of the policies and functions are more permissive than I think you intend, and that's worth locking down early. Happy to point you at general guidance if it helps.

I'm going to close this PR for the reasons above, but thank you again for building on OpenWA — I'm glad it's a useful base. If you'd like to talk through the API-consumer approach, open a Discussion and I'm happy to help.

PR #484, opened 2026-06-27
Dashboard /chats: clickable links, WhatsApp formatting, photo lightbox, scroll memory

softronicve:
## Summary

Three UX polishes to the dashboard's `/chats` page that together make it feel modern:

1. **Clickable links + WhatsApp text formatting in message bodies.** URLs (including bare domains like `github.com/x`) become `<a target="_blank" rel="noopener noreferrer">`. `*bold*`, `_italic_`, `~strike~`, `` `inline code` ``, and `` ```code blocks``` `` render with the corresponding semantics. Nesting (`*_bold italic_*`) supported. URLs inside code segments stay literal.
2. **Photo lightbox** powered by [`yet-another-react-lightbox`](https://yet-another-react-lightbox.com): pinch-to-zoom on touch, wheel-zoom on desktop, pan when zoomed, swipe carousel between all photos in the current chat, ESC / click backdrop to close, download button per image.
3. **Per-chat state cache + scroll-position memory.** Messages are wrapped in a React Query `useQuery` (per `(sessionId, chatId)`); revisiting a chat renders cached data instantly (no spinner). The smooth-scroll-to-bottom animation is gone; instead, `useChatScrollPosition` saves each chat's `scrollTop` on the way out and restores it on the way back in via `useLayoutEffect` (before paint, no visible jump). Auto-scroll on incoming messages only fires when the user is already near the bottom (threshold 100px); outgoing always scrolls.

## Architecture

The maintainer's "pure logic in utils, thin glue in React, test the utils" pattern is preserved — no Vitest / Jest / RTL / jsdom added. Three new tested utilities (`messageFormatter`, `scrollDecision`, `decideRestoreTarget` + the existing `chatMessages` extended), two hooks (`useChatMessages`, `useChatScrollPosition`), two components (`MessageBody`, `MediaLightbox`).

- WebSocket cache writes use `event.sessionId` (not `selectedSessionId`) so messages arriving for a non-active chat correctly update its cache slice.
- The send-success branch in `Chats.tsx` uses `queryClient.setQueryData` directly with a read-before-write race-guard so that an echo from the WS arriving before the optimistic `temp` swap is handled correctly.
- yarl theme overrides live in `App.css` (global) because the portal-rendered overlay can't be scoped under `.chats-page`.

## Dependencies added (3, ~25 KB gzip combined)

- `linkifyjs@^4` + `linkify-react@^4` — the standard React linkifier; handles trailing punctuation and TLD validation correctly.
- `yet-another-react-lightbox@^3` with plugins `zoom`, `download`, `counter`, `captions`.

No backend, env, or migration changes.

## Files

- New utils: `dashboard/src/utils/{messageFormatter,scrollDecision}.ts` (+ tests)
- New hooks: `dashboard/src/hooks/{useChatMessages,useChatScrollPosition}.ts` (+ `useChatScrollPosition.test.ts`)
- New components: `dashboard/src/components/chats/{MessageBody,MediaLightbox}.tsx`
- Extended: `dashboard/src/utils/chatMessages.ts` (exports `ChatMessageView` + mutation helpers, mostly extracted from the now-thinner `Chats.tsx`)
- Rewired: `dashboard/src/pages/Chats.tsx` (+171 / −128)
- CSS: `dashboard/src/pages/Chats.css` (code/pre rules, zoom cursor) + `dashboard/src/App.css` (yarl theme)

## Tests

```
npm --prefix dashboard run test:unit   → 69 pass, 0 fail
npm --prefix dashboard run build       → tsc -b && vite build clean
npm --prefix dashboard run lint        → 0 errors (3 pre-existing warnings, unchanged)
```

Tests cover: formatter parser including the WhatsApp boundary rules and nesting (14 cases), scroll decision geometry (6 cases), restore decision matrix incl. cold-open transitions (8 cases), and message mutation helpers (9 cases).

## Test plan (manual smoke recommended before merge)

- [ ] Open chat A → spinner → messages render at bottom (no smooth animation).
- [ ] Switch to chat B → back to A → A renders instantly from cache, at the position you left it.
- [ ] Scroll up in A; switch to B; switch back to A → scroll position preserved.
- [ ] Receive a message in A while scrolled up → no auto-scroll. While near bottom → auto-scroll.
- [ ] Send a message → always scrolls to bottom.
- [ ] Send a message with `mira *bold* y _italic_ y github.com/x` → formatting applied; `github.com/x` becomes a clickable link in a new tab.
- [ ] Receive a photo; click it → lightbox opens. Wheel-zoom, pan, ESC, arrow keys all work. Download button downloads the image.

## Non-goals (deferred)

- Email / phone / mention / hashtag linkification (only URLs).
- "↓ N new messages" badge when scrolled up (the hook leaves room for it).
- Videos in lightbox (`type === 'image'` only).
- WS reconnect → cache invalidation strategy.

## Design + plan docs

`docs/superpowers/specs/2026-06-26-dashboard-chats-ux-improvements-design.md`
`docs/superpowers/plans/2026-06-26-dashboard-chats-ux-improvements.md`


dashboard/src/hooks/useChatMessages.ts
```diff
@@ -0,0 +1,74 @@
...
+        sessionApi.getChatMessages(sessionId, chatId!, 100),
+        sessionApi.getChatHistory(sessionId, chatId!, 100, true),
+      ]);
+      if (dbRes.status === 'rejected' && historyRes.status === 'rejected') throw dbRes.reason;
+      const dbMessages = dbRes.status === 'fulfilled' ? dbRes.value.messages : [];
+      const history = historyRes.status === 'fulfilled' ? historyRes.value.map(mapEngineHistoryMessage) : [];
+      return mergeChatMessages(dbMessages, history);
+    },
+    enabled: Boolean(sessionId && chatId),
+    staleTime: Infinity,
```
rmyndharis, line 41: **Cache-seed truncation (blocking).** With `staleTime: Infinity`, once a slice exists for a `(sessionId, chatId)` it's never stale, so opening that chat won't run `queryFn`. But `appendMessage` (`Chats.tsx:231`) seeds a slice for *any* chat in the active session — including ones never opened — from a single live WS message. Result: opening that chat shows only the messages that arrived live, not its DB/engine history, until gcTime (30 min) evicts it.

Cleanest fix: only append when a slice already exists, e.g. guard with `queryClient.getQueryData(messagesQueryKey(sessionId, chatId))` and skip when `undefined`. Alternatively mark the seeded slice stale so the next open refetches.

dashboard/src/pages/Chats.tsx
```diff
@@ -199,34 +212,32 @@ export function Chats() {
...
+        type: asMessageType(newMsg.type),
+        direction: newMsg.fromMe ? 'outgoing' : 'incoming',
+        status: 'sent',
+        timestamp: newMsg.timestamp,
+        createdAt: new Date(newMsg.timestamp * 1000).toISOString(),
+        metadata: newMsg.metadata || {
+          media: newMsg.media,
+          quotedMessage: newMsg.quotedMessage,
+        },
+      };
```
rmyndharis, line 232: This writes a cache slice for every chat in the session, even unopened ones — paired with `staleTime: Infinity` (`useChatMessages.ts:41`) it truncates the chat's history on first open. See the note there; a `getQueryData` existence guard here is the minimal fix.

dashboard/src/pages/Chats.tsx
```diff
@@ -111,11 +111,18 @@ export function Chats() {
 
   // Selected chat & message history
   const [activeChat, setActiveChat] = useState<Chat | null>(null);
-  const [messages, setMessages] = useState<ChatMessageView[]>([]);
-  const [loadingMessages, setLoadingMessages] = useState<boolean>(false);
+  const {
+    data: messages = [],
+    isLoading: loadingMessages,
```
rmyndharis, line 116: **Silent load failures (blocking).** Only `data`/`isLoading` are consumed — `isError`/`error` from `useChatMessages` are dropped. When `queryFn` rejects (both the DB *and* engine-history calls fail), the UI falls through to the `messages.length === 0` → "No messages in chat" branch, indistinguishable from a real empty chat, and the previous `showErrorToast(t('chats.errors.loadMessages'))` is gone (its i18n key is now orphaned). Please surface the error — e.g. a `useEffect` on `isError` that toasts, or a distinct error state.

dashboard/src/utils/chatMessages.ts
```diff
@@ -35,3 +35,71 @@ export function mergeChatMessages(db: ChatMessage[], history: ChatMessage[]): Ch
...
+
+/**
+ * Append `incoming` to `list`. If an entry with the same id exists, replace it in place.
+ * Returns a new array — does not mutate the input.
+ */
+export function mergeOrAppend(
+  list: ChatMessageView[],
+  incoming: ChatMessageView,
+): ChatMessageView[] {
+  const idx = list.findIndex(m => m.id === incoming.id);
```
rmyndharis: Minor: dedups on `id` only, but the WS handler this PR removed also matched `waMessageId` (`prev.some(m => m.id === x.id || m.waMessageId === x.id)`). DB-persisted messages have `id` = UUID and `waMessageId` = the WA id, while a WS `mappedMessage` sets `id` = the WA id — so a WS replay of an already-cached DB message won't match and appends a duplicate bubble. Restoring the `waMessageId` arm (plus a test) closes it. Non-blocking.

dashboard/src/utils/messageFormatter.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  ]);
+});
+
+test('marker without outside boundary stays literal (no over-formatting)', () => {
+  // 'word*bold*end' has no boundary char before the opening '*' nor after the closer.
+  // Per WhatsApp rules this is literal text — the boundary guard prevents over-formatting.
+  assert.deepEqual(parseMessageBody('word*bold*end'), [
+    { type: 'text', value: 'word*bold*end' },
+  ]);
+});
```
rmyndharis, line 94: The formatter is safe by construction today (it only emits `{type:'text', value}` / code values that React escapes, and links go through `linkify-react`, never `dangerouslySetInnerHTML`). Since this is the one component rendering untrusted, attacker-controlled message bodies, a few assertions pinning the invariant would be valuable insurance against a future refactor:
- `parseMessageBody('<script>alert(1)</script>')` → a single literal text node
- `*<b onmouseover=x>*` → a bold node wrapping the literal `<b onmouseover=x>` text
- a `javascript:alert(1)` body stays plain text (no link)

Cheap, no jsdom needed, runs in the existing `test:unit` runner.

rmyndharis (commented):
Thanks for this — it's a genuinely strong PR. The architecture follows the project's "pure logic in utils, thin glue in React, test the utils" pattern, the new utilities are well covered (69 tests), and the formatting/link rendering is safe by construction (AST → React text nodes, no `dangerouslySetInnerHTML`, `linkify-react` with `rel="noopener noreferrer"` and default-only URL schemes). CI is green across the board. Really nice work.

I traced the cache redesign carefully — there are three things I'd like fixed before merge, plus a few minor notes inline. None are hard; all dashboard-only and non-breaking.

### Blocking before merge
1. **Cache-seed truncation** — `appendMessage` runs for *every* chat in the active session, even ones never opened, seeding a 1-message cache slice. With `staleTime: Infinity`, opening that chat later finds "fresh" data and never fetches its history, so you see only the live messages until gcTime (30 min) evicts it. (inline on `useChatMessages.ts:41` / `Chats.tsx:231`)
2. **Silent load failures** — `useChatMessages` exposes `isError`/`error`, but they're never read; a failed fetch falls through to the same "No messages in chat" empty state and the old error toast is gone. (inline on `Chats.tsx:116`)
3. **Add one XSS-inertness test** — the formatter is the one place untrusted message bodies are rendered. It's safe today, but a 2–3 line test pinning that invariant is cheap insurance against a future refactor. (inline on `messageFormatter.test.ts`)

### Minor (non-blocking, your call)
- `mergeOrAppend` dedups on `id` only; the removed WS handler also matched `waMessageId`, so a WS replay of a DB-cached message can duplicate the bubble. (inline on `chatMessages.ts:59`)
- `file://` URLs become clickable — consider narrowing linkify to http/https.
- Stickers get a `zoom-in` cursor + click handler but aren't in `imageMedia`, so the click is a no-op.
- The inline send-echo race-guard is the trickiest concurrency bit and isn't extracted/tested — worth a pure `reconcileSentEcho` helper if you touch this again.
- The ~1.8k lines of plan/spec docs under `docs/superpowers/` are fine by me (there's precedent), just flagging.

Happy to merge once 1–3 are in. Thanks again — high-quality work.

rmyndharis:
Thanks again for this PR @softronicve — it's a strong, well-tested contribution. To get it landed without keeping it waiting, I pushed the fixes for the two blocking review items (and the dedup nit) directly to the branch:

- **History truncation** — `appendMessage` no longer seeds a React Query slice for a never-opened chat. With `staleTime: Infinity` that phantom slice was "fresh", so opening the chat skipped the full-history `queryFn` and showed only the single live message. It now appends only to an already-loaded slice; never-opened chats fetch their full history on first open.
- **Silent load failures** — the thread now renders an error state when the message load fails (both the DB *and* engine-history calls reject); `isError` was being dropped, so a failure showed a misleading empty chat. Added `chats.loadMessagesError` across all 9 locales.
- **Dedup** — `mergeOrAppend` now keys on `waMessageId ?? id` (same as `mergeChatMessages`), so a live WS message (`id` = WA id) and its DB copy (`id` = UUID, `waMessageId` = WA id) no longer double-add. Added a regression test.

All green locally (70 dashboard unit tests, lint, build, i18n parity). Merging once CI passes.

softronicve:
I realized that when I have too much messages the loading is very slow, so
I'm pushing a new PR to put a better pagination, when you scroll up load
more using the infinite scroll

PR #486, opened 2026-06-27
fix(dashboard): unique color per message type in 'by type' chart

### LINKED ISSUE #485: Dashboard: "Messages by type" chart reuses the same green for audio and text (palette only has 8 colors for 11 types)
## Summary

In the dashboard **"Messages by type"** donut chart, the legend assigns the **same green (`#25d366`)** to both **audio** and **text**. Other types can also collide. Each message type should get a distinct color.

## Root cause

The slice palette has only **8 colors** but there are **11 message types**, so the modulo index wraps around and reuses colors.

`dashboard/src/components/DashboardCharts.tsx:26`
```ts
const TYPE_COLORS = ['#25d366', '#3b82f6', '#a855f7', '#f59e0b', '#ef4444', '#06b6d4', '#ec4899', '#64748b'];
```

`dashboard/src/components/DashboardCharts.tsx:129`
```tsx
<Cell key={entry.name} fill={TYPE_COLORS[i % TYPE_COLORS.length]} />
```

Message types produced by `src/engine/adapters/message-mapper.ts` plus `text`/`image`/`video` are **11**:
`audio, voice, document, sticker, location, contact, revoked, unknown, image, video, text`.

With `i % 8`, the 9th/10th/11th types fall back to colors `0/1/2`:
- index 8 → `#25d366` (green) → collides with **audio** (index 0)
- index 9 → `#3b82f6` (blue)
- index 10 → `#a855f7` (purple)

That is why **audio** and **text** render with identical green in the legend and donut.

## Expected

Every message type has its own distinct color in the legend and donut.

## Suggested fix

Expand `TYPE_COLORS` to cover all message types with visually distinct hues (≥11 entries), or map colors by type name instead of array index so a stable, unique color is guaranteed per type. Mapping by name is preferable: it keeps colors stable even when the set of present types changes between requests.

## Environment

- OpenWA v0.7.6
- Engine: Baileys
- Affected file: `dashboard/src/components/DashboardCharts.tsx`

ulises2k:
## Summary

The dashboard **"Messages by type"** donut assigned the **same green (`#25d366`)** to both **audio** and **text** (and could collide on other types too). This keys the slice color by **type name** instead of array index, so every type gets a stable, unique color.

Fixes #485

## Root cause

`TYPE_COLORS` had only **8 colors**, but `mapMessageType()` emits **11 types** (`audio, voice, document, sticker, location, contact, revoked, unknown` + `text, image, video`). The slice fill used `TYPE_COLORS[i % TYPE_COLORS.length]`, so the 9th/10th/11th types wrapped around to colors `0/1/2`:

- index 8 → `#25d366` (green) → collided with **audio** (index 0)
- index 9 → `#3b82f6` (blue)
- index 10 → `#a855f7` (purple)

## Change

- `TYPE_COLORS` is now a `Record<string, string>` mapping each known type to a distinct hue.
- New `colorForType(name)` returns the mapped color, with a **deterministic hash-based fallback** (`FALLBACK_COLORS`) for any unexpected/unmapped type so its color stays stable across renders.
- `<Cell>` now uses `colorForType(entry.name)` rather than the modulo index.

Because color is keyed by name (not position), two types can never share a color, and a slice keeps its color even when the set of present types changes between requests.

## Testing

- `npx tsc -b` — passes
- `npx eslint src/components/DashboardCharts.tsx` — clean

Only `dashboard/src/components/DashboardCharts.tsx` is touched.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,6 +10,9 @@
 NODE_ENV=production
 API_PORT=2785
 LOG_LEVEL=info                      # error | warn | info | debug
+# Console output format. Default: json in production (containers, log aggregators),
+# human-readable pretty otherwise. Force one with: json | pretty
+# LOG_FORMAT=pretty
 
 # Auto-start previously authenticated sessions on server boot
 AUTO_START_SESSIONS=false
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,4 @@ docs/plans/
 CLAUDE.md
 _docs/
 .superpowers/
+/graphify-out
--- a/.gitignore
+++ b/.gitignore
@@ -77,4 +77,3 @@ docs/plans/
 CLAUDE.md
 _docs/
 .superpowers/
-/graphify-out
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Human-readable console logs: the `LoggerService` now renders a colorized, NestJS-style line (`[OpenWA] <pid> - <timestamp> <LEVEL> [Context] <message>` with dimmed `key=value` metadata and stack traces on their own line) instead of always emitting raw JSON, so application logs line up visually with NestJS's own framework logs. The format defaults to structured JSON in production (`NODE_ENV=production`, for containers and log aggregators) and human-readable pretty everywhere else, and can be pinned with `LOG_FORMAT=pretty|json`. `NO_COLOR` / `FORCE_COLOR` are honored. JSON output is byte-for-byte unchanged when selected. (#469)
+
 ## [0.7.6] - 2026-06-26
 
 ### Changed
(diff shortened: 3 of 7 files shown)


rmyndharis (approved):
Nice fix — keying the donut color by message-type name instead of array index is the right call. The index-based version let a slice change color whenever the set/order of present types shifted between requests (and wrapped past 8 types); keying by name fixes both, and the deterministic hash fallback keeps unmapped types stable too.

I checked the keys against the source: all 11 match the engine-neutral `MessageType` union in `src/engine/interfaces/whatsapp-engine.interface.ts`, which both adapters normalize to before persistence — so they line up with what the stats endpoint emits. 👍

Approving. A couple of small things you can take or leave:

- **Scope:** the PR also bundles an unrelated `ENGINE_TYPE` env-precedence test and a `/graphify-out` `.gitignore` entry. `graphify-out` looks like a local-tool artifact — better in your personal/global gitignore than the repo's. The env test is valid but overlaps an existing one; ideally these would be a separate PR so this one stays "chart colors" end to end.
- `forward` is a real persisted outgoing type that isn't in `TYPE_COLORS` (it falls to the stable fallback — not a bug). Either add a curated hue or soften the "covers every type" comment.
- That comment references a `mapMessageType()` that doesn't exist — the real sources are `mapWwebjsMessageType` / `mapBaileysMessageType`.

None blocking. Thanks!

PR #492, opened 2026-06-27
feat: add MEDIA_DOWNLOAD_ENABLED env var to skip inbound media download

spidgrou:
## Summary

Adds a new environment variable MEDIA_DOWNLOAD_ENABLED (default: true) that lets operators disable inbound media download entirely.

When set to false, the engine skips downloading, decrypting, and processing media from incoming messages — no memory allocation, no storage, no base64 encoding, no Puppeteer/Chromium overhead. The message is emitted without a media field, which callers already handle correctly (they check for undefined).

## Motivation

Not every deployment needs to process inbound media. For setups that only care about text messages — alerts, notifications, simple chatbots — downloading every photo, video, and document that comes in is wasted CPU, memory, and bandwidth. On low-resource VPS or container environments, this can make a real difference.

The existing MEDIA_DOWNLOAD_MAX_BYTES cap only limits size; it doesn't let you turn media processing off at the source. This PR adds that toggle.

## Changes

| File | Change |
|------|--------|
| src/engine/adapters/inbound-media-cap.ts | Add isMediaDownloadEnabled() — reads MEDIA_DOWNLOAD_ENABLED env var, accepts false, 0, or no to disable |
| src/engine/adapters/whatsapp-web-js.adapter.ts | Guard in capInboundMediaFor() — returns undefined immediately when disabled |
| src/engine/adapters/baileys.adapter.ts | Guard in the inbound media handler — same approach, skips the entire media download path |
| .env.example | Document MEDIA_DOWNLOAD_ENABLED |
| src/engine/adapters/inbound-media-cap.spec.ts | Tests for the new function |

## Backward Compatibility

Fully backward compatible. Default is true — existing deployments keep downloading media as before. No code changes needed unless you want to disable it.

## Testing

- isMediaDownloadEnabled() defaults to true when MEDIA_DOWNLOAD_ENABLED is unset
- Returns false for false, 0, no
- Returns true for any other value (true, 1, yes, garbage)
- Both engine adapters handle undefined media correctly (verified in existing code paths)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -192,6 +192,8 @@ WEBHOOK_SSRF_PROTECT=true
 # store or a sidecar webhook receiver).
 # SSRF_ALLOWED_HOSTS=localhost,minio
 # Server-side media size/time limits:
+# MEDIA_DOWNLOAD_ENABLED=true          # Set to false to skip downloading inbound media entirely
+#                                     # (no decryption, no memory allocation, no storage).
 # MEDIA_DOWNLOAD_MAX_BYTES=52428800   # cap remote-URL sends, inbound media, AND outbound base64 sends (default 50 MiB; oversized inbound media is dropped, message kept; oversized base64 is rejected with 400)
 # MEDIA_DOWNLOAD_TIMEOUT_MS=30000     # abort a slow media download (default 30s)
 # Storage import/export limits (ADMIN /infra/storage/* endpoints):
--- a/.env.example
+++ b/.env.example
@@ -194,6 +194,8 @@ WEBHOOK_SSRF_PROTECT=true
 # Server-side media size/time limits:
 # MEDIA_DOWNLOAD_ENABLED=true          # Set to false to skip downloading inbound media entirely
 #                                     # (no decryption, no memory allocation, no storage).
+#                                     # NOTE: disabling also makes hasMedia report false and removes the
+#                                     # media field from webhooks and the dashboard.
 # MEDIA_DOWNLOAD_MAX_BYTES=52428800   # cap remote-URL sends, inbound media, AND outbound base64 sends (default 50 MiB; oversized inbound media is dropped, message kept; oversized base64 is rejected with 400)
 # MEDIA_DOWNLOAD_TIMEOUT_MS=30000     # abort a slow media download (default 30s)
 # Storage import/export limits (ADMIN /infra/storage/* endpoints):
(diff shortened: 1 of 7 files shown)


rmyndharis:
Thanks for this, @spidgrou — it's a clean, well-scoped addition. 👍 Letting operators skip inbound media entirely is genuinely useful for text-only/low-resource deployments, and routing it through the shared `inbound-media-cap.ts` so both the whatsapp-web.js and Baileys engines are covered is exactly the right place for it. Default `true` keeps it fully backward-compatible, and emitting the message without a `media` field is safe since downstream already treats `media` as optional.

A few small things before merge — none are blockers:

**Please address**
- **CI hasn't run on the branch** (no checks reported). Could you re-trigger it (a rebase/empty push usually does it) so we can confirm `lint` / `build` / `test` are green?
- **A couple of focused tests for the toggle itself.** Right now only the pure parser is covered. One assertion per engine would lock the behavior in: with `MEDIA_DOWNLOAD_ENABLED=false`, deliver a media message and assert (a) the download is **not** called and (b) the emitted message has no `media`.

**Nice to have**
- A one-line note in `.env.example` that disabling inbound media also makes `hasMedia` report `false` (and removes the `media` field) in webhooks and the dashboard — just so the side effect isn't surprising.
- Minor: the flag parse is case/whitespace-sensitive (`FALSE` / `' false '` won't disable). Normalizing with `?.trim().toLowerCase()` would be a touch friendlier, though it's entirely optional.

Once CI is green and the two engine tests are in, this is good to merge. Really appreciate the contribution! 🙏

spidgrou:
Hi @rmyndharis, all feedback addressed:

**Required:**
- ✅ Case/whitespace-insensitive flag parsing (`.trim().toLowerCase()` — `FALSE`, `' false '` etc. all work now)
- ✅ Engine test for whatsapp-web.js: `MEDIA_DOWNLOAD_ENABLED=false` → `downloadMedia` not called, no `media` field on the emitted message
- ✅ Engine test for Baileys: same assertions

**Nice to have:**
- ✅ `.env.example` note: disabling also makes `hasMedia` report `false` and removes the `media` field from webhooks and dashboard
- ✅ Fixed an unused constant (`DEFAULT_MEDIA_DOWNLOAD_ENABLED`) and a lint warning in the new wwebjs test

Full test suite passes locally: 1476 tests, 113 suites, 0 failures. Lint clean.

Regarding CI: the workflow runs are showing `action_required` on your end — I believe that's the standard fork approval gate. Happy to rebase or adjust anything else before you approve the run.

PR #494, opened 2026-06-28
feat(dashboard): chats UX — reliable scroll, lazy pagination, on-demand media, call/location rendering

softronicve:
## Description

Follow-up to #484, hardening the `/chats` experience after testing on live sessions. Grouped:

**Scroll & layout**
- Fix reopened chats jumping to the top — scroll position is recorded continuously via `onScroll` (not read after the DOM already swapped to the next chat), so returning restores where you were.
- Reliable open-at-bottom (re-pin on the next frame for late media layout).
- WhatsApp-style jump-to-bottom FAB.

**Loading & pagination**
- Render windowing: render the recent ~30 messages and reveal more on scroll-up, keeping the DOM light on large chats.
- Lazy "load older" via the engine deep-history (both directions, up to 2000) once the window reaches the top — older messages, including the user's own, are reachable again.
- Drop the incoming-only band at the top of the initial thread (the DB stores only incoming and its page reached further back than the both-directions history page, so the top looked one-sided).

**Media**
- On-demand media for older (metadata-only) messages with **server-side persistence**: new `GET /sessions/:id/messages/:chatId/:messageId/media` resolves the message via the engine, downloads its media, and caches it through `StorageService` — so a media fetched once is instant next time. Client downloads are concurrency-limited with retry/backoff (no dropped media under a 429 burst).
- Payload-less media renders a labeled placeholder; the lazy loader (IntersectionObserver) swaps in the real photo/audio as the bubble nears the viewport.

**Message types**
- Render location messages (the base64 map thumbnail) instead of dumping the base64 as text.
- Map WhatsApp `call_log` → a new `call` MessageType with missed/video detection (from the raw `_data`), rendering "📞 Missed call" / "📹 Video call" etc. instead of a misleading "attachment".
- Precise non-media handling: only real media types attempt a download; contacts and unsupported types get neutral labels.

**Hardening**
- linkify blocks non-http(s) schemes (e.g. `file://`); stickers open in the lightbox; formatter XSS-inertness tests; dedup test for optimistic temps.

Aligned with `main` (#484 already merged): the cache-seed-truncation, dedup, and load-error handling are main's — this PR builds on top.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [x] Tests added/updated (dashboard pure logic: scroll/window/pagination, formatter XSS, dedup; backend: `call_log`→`call` mapper case)
- [ ] Documentation updated
- [x] Lint passes (backend + dashboard)
- [x] Self-reviewed

N/A — behavioral; verified against live sessions (scroll restore, old-message pagination, on-demand media, location/call rendering).

## Related Issues
Follow-up to #484.

### DIFF
--- a/dashboard/src/components/chats/LazyMediaPlaceholder.tsx
+++ b/dashboard/src/components/chats/LazyMediaPlaceholder.tsx
@@ -1,0 +1,49 @@
+import { useEffect, useRef } from 'react';
+import { Loader2 } from 'lucide-react';
+
+interface Props {
+  /** Label shown while the media loads (e.g. "📷 Photo"). */
+  label: string;
+  /** Called once when the placeholder approaches the viewport, to start downloading the media. */
+  onVisible: () => void;
+}
+
+/**
+ * Placeholder for an older media message whose payload isn't loaded yet. When it nears the viewport it
+ * fires `onVisible` once to lazily download the media. The IntersectionObserver `rootMargin` doubles as
+ * a soft prefetch — it starts the fetch a bit before the bubble is actually on screen.
+ */
+export default function LazyMediaPlaceholder({ label, onVisible }: Props) {
+  const ref = useRef<HTMLDivElement>(null);
+  const firedRef = useRef(false);
+  // Keep the latest onVisible in a ref so the observer can mount once (the prop is a fresh closure each
+  // render; without this the observer would tear down and re-create on every parent re-render).
+  const onVisibleRef = useRef(onVisible);
+  useEffect(() => {
+    onVisibleRef.current = onVisible;
+  }, [onVisible]);
+
+  useEffect(() => {
+    const el = ref.current;
+    if (!el) return;
+    const observer = new IntersectionObserver(
+      entries => {
+        if (entries[0]?.isIntersecting && !firedRef.current) {
+          firedRef.current = true;
+          onVisibleRef.current();
+          observer.disconnect();
+        }
+      },
+      { rootMargin: '300px' },
+    );
+    observer.observe(el);
+    return () => observer.disconnect();
+  }, []);
+
+  return (
+    <div ref={ref} className="message-media-placeholder message-media-placeholder--loading">
+      <Loader2 className="animate-spin" size={14} />
+      <span>{label}</span>
+    </div>
+  );
+}
--- a/dashboard/src/components/chats/MessageBody.tsx
+++ b/dashboard/src/components/chats/MessageBody.tsx
@@ -13,6 +13,12 @@ const linkifyOptions = {
   rel: 'noopener noreferrer',
   defaultProtocol: 'https',
   ignoreTags: ['code', 'pre'],
+  validate: {
+    // Only linkify web URLs. A scheme'd match that isn't http(s) — e.g. file://… — is left as inert
+    // text. Bare domains and www.* have no '://' here and still resolve to https via defaultProtocol;
+    // emails are a separate token type and remain linkified.
+    url: (value: string) => !value.includes('://') || /^https?:\/\//i.test(value),
+  },
   attributes: {
     onClick: (e: React.MouseEvent) => e.stopPropagation(),
   },
--- a/dashboard/src/hooks/useChatMessages.ts
+++ b/dashboard/src/hooks/useChatMessages.ts
@@ -73,5 +73,21 @@ export function useChatMessagesActions() {
         (old = []) => removeMessageById(old, id),
       );
     },
+    /**
+     * Load older messages by fetching a deeper slice of the engine history (both directions, metadata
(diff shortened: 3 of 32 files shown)


rmyndharis:
Thanks for the thorough follow-up, @softronicve — this is a substantial quality pass on `/chats`. 🙌 The render windowing, lazy deep-history, on-demand media, and call/location bubbles are all great additions, and I appreciate that the engine-interface change is additive (both engines implement `downloadMessageMedia`, so no breaking contract) and that i18n parity holds across all 9 locales.

I did a detailed pass and there are a few things I'd like to sort out before we merge. Grouping by priority:

### Please address (higher priority)
1. **Server-side media cache is unbounded.** `getMessageMedia` persists `media-cache/<sessionId>/<id>.json` with no TTL/eviction, and that path is also counted in storage stats and swept into export tarballs. Could we bound it (size/age cap or simple LRU), exclude it from export + the file-count stats, and clean it up on session delete?
2. **Pagination caps prematurely when reopening a deep-loaded chat.** `historyLimitRef` hard-resets to the initial limit on chat switch, so "load older" stops early on a chat you'd already paged into. Deriving the starting ceiling from the cached thread length (instead of resetting) should fix it.
3. **Concurrent deep-history fetch guard.** `loadingOlder` is React state, so a fast scroll near the top can fire duplicate fetches before it updates — a `useRef` guard would make it synchronous.
4. **Windowing vs. the #484 scroll-memory.** On reopen the window resets to ~30, so a previously-saved scroll position no longer matches the rendered layout. Persisting `visibleCount` per chat alongside the saved scrollTop would let them cooperate.

### Merge-order note ⚠️
There's an open hotfix, **#495**, that touches the same chat files (`utils/chatMessages.ts`, `pages/Chats.tsx`, `hooks/useChatMessages.ts`) to fix a delivery-status regression. This PR currently carries the older `mergeOrAppend` (full-replace), which both downgrades the delivery tick on a replayed `message.sent` echo **and** can clobber the lazily-fetched media you add here. Cleanest path: **land #495 first, then rebase this on top** — after that, please drop the inline `mergeDeliveryStatus`/`DELIVERY_RANK` here in favor of the shared helper #495 moves into `utils/chatMessages.ts`, so there's a single forward-only merge used by both the append and ack paths.

### Nice to have
- On Baileys (no on-demand media), the placeholder retries a terminal 501/404 four times with backoff — reading the status and stopping immediately on non-retryable codes would avoid the spinner.
- Live/webhook `call` messages don't carry the `{ video, missed }` detail (only history does), and the "missed" heuristic can mislabel outgoing/group calls — worth factoring the detection into `buildIncomingMessageBase` so live + history + webhook agree.
- A couple of backend tests (`getMessageMedia` cache hit/miss, `downloadMessageMedia`, call-log extraction) and a `CHANGELOG` `[Unreleased]` entry for the new media endpoint + call rendering.
- Minor: the `:chatId` route segment on `GET :chatId/:messageId/media` is unused by the handler, and the Baileys 501 isn't documented in Swagger.

None of this takes away from a really solid PR — happy to re-review quickly once #495 is in and the cache/pagination items are addressed. Thank you! 🙏

rmyndharis:
Hi @softronicve — first, a genuine thank you. This was a thoughtful, substantial PR, and the care you put into it really showed: the render windowing, the lazy deep-history, the call/location bubbles, the i18n parity across every locale. None of the work below takes anything away from that. 🙏

I owe you an honest explanation, and an apology for the timing.

While this PR was open, a couple of changes landed on `main` that ended up overlapping with large parts of it:

- **#484** added the chat scroll-memory + clickable links / lightbox.
- **#524 / #525** tackled the media-footprint / OOM problem — but with a deliberately different approach than this PR's server-side media cache: instead of caching and lazily re-fetching media, they stop caching base64 and show a lightweight placeholder for older media.

Because of that last point especially, merging this PR as-is would actually re-introduce the unbounded media cache that #525 set out to remove — so the two directions now pull against each other through no fault of yours. It's simply a case of the codebase moving while the review was in flight, and I'm sorry it landed that way after the effort you put in.

Rather than let the genuinely new and still-valuable part go to waste, I've **rebased your call & location rendering onto current `main`** and opened **#541** to carry it forward. Your authorship is preserved on the commit (`Co-authored-by`), the CHANGELOG credits you, and #541 references this PR. So this work *is* shipping — just trimmed to the slice that doesn't collide with what's already on `main`.

I'm closing this one as superseded, but please don't read that as your contribution being unwelcome — quite the opposite. If you'd like to pick up any of the other ideas here (e.g. the live/webhook call detail, or scroll-window cooperation) against the current code, I'd be glad to review, and thank you again for making OpenWA better. 🙏

softronicve:
Thanks you for create this tools, I’m using at openwa.softronic.dev

Issue #473, opened 2026-06-25
Baileys engine: recipients see "This message won't disappear — the sender may be using an older version of WhatsApp"


ulises2k:
## Question / Issue

When running OpenWA with the **Baileys** engine (`ENGINE_TYPE=baileys`), some chats show the inline WhatsApp system notice:

> **This message won't disappear**
> The sender may be using an older version of WhatsApp.

(Original, in Spanish: *"Este mensaje no desaparecerá. Es posible que el remitente esté usando una versión anterior de WhatsApp."*)

This appears even though the device is fully linked and messages are sent/received normally. I'd like to understand **why this happens specifically with the Baileys engine**, and whether there's anything that can be configured on the OpenWA side to avoid it.

## What I think is going on (not confirmed)

This WhatsApp notice typically shows up in chats that have **disappearing messages** turned on, when an incoming/relayed message does **not** carry the expected ephemeral metadata (e.g. `disappearingMode` / `ephemeralExpiration`). Because Baileys is a protocol-level client, my guess is that messages sent through it don't include the disappearing-messages settings that the official client expects, so WhatsApp flags the sender as "an older version."

If that's correct, the question is whether OpenWA could propagate/honor the chat's ephemeral expiration on outgoing messages when using Baileys.

## Questions

1. Why does this notice appear with the Baileys engine but (as far as I've seen) not with `whatsapp-web.js`?
2. Is OpenWA expected to set/propagate the disappearing-messages (`ephemeralExpiration` / `disappearingMode`) metadata on outgoing Baileys messages?
3. Is there any config flag or workaround to prevent recipients from seeing this "older version of WhatsApp" warning?

## Environment

- OpenWA version: **0.7.3**
- Engine: **Baileys** (`ENGINE_TYPE=baileys`)
- `@whiskeysockets/baileys`: **6.7.23**
- Storage: SQLite
- Deployment: Docker (single container)

## Steps to reproduce

1. Run OpenWA with `ENGINE_TYPE=baileys` and link a device.
2. Use a chat that has **disappearing messages** enabled (or where the contact has them on).
3. Send/receive a message through the Baileys session.
4. The chat shows the "This message won't disappear — the sender may be using an older version of WhatsApp" notice.

Thanks!

rmyndharis:
Hi @ulises2k — great question, and your instinct is exactly right. 🙏

[...]

PR #506, opened 2026-06-29
feat: add STORE_EPHEMERAL_MESSAGES to skip persisting disappearing messages

spidgrou:
### What

A new env var `STORE_EPHEMERAL_MESSAGES` (default `true`, backward compatible). Set it to `false` and OpenWA stops saving WhatsApp disappearing messages to the local database.

WhatsApp calls them "disappearing messages." Internally they carry an `ephemeralDuration` in the raw payload — 86400 for 24 hours, 604800 for 7 days, 7776000 for 90 days. The client deletes them locally after the timer runs out. But OpenWA saves everything it receives to its SQLite database, so those messages never actually disappear from the dashboard.

This PR gives operators a choice. Keep the current behaviour (store everything, default) or tell OpenWA to treat ephemeral messages the same way WhatsApp does.

### Why

Same pattern as `MEDIA_DOWNLOAD_ENABLED` from PR #492 — a simple on/off toggle proved useful there. Some use cases here:

- **Privacy.** A group chat has disappearing messages on for a reason. OpenWA should not silently archive everything.
- **Storage.** Ephemeral chats tend to be high-volume. Skipping them keeps the DB lean.
- **Legal / compliance.** Storing messages the sender expects to disappear is a liability in some contexts.

Opt-in (`false`). Default behaviour does not change.

### How

Three changes to the backend, no dashboard work.

**1. `IncomingMessage` interface** — added `ephemeralDuration?: number`. Raw WhatsApp value in seconds, or undefined when the message is not ephemeral.

**2. `buildIncomingMessageBase`** — reads `_data.ephemeralDuration` from the wwebjs message. It is already on the raw payload; we just were not looking at it.

**3. Session service guard** — in `onMessage`, right after `isStatusBroadcast` and before the hook/persist pipeline. If the env var is `false` and `ephemeralDuration > 0`, skip everything. No DB insert, no webhook, no websocket event. Same early-return pattern as `isStatusBroadcast` a few lines above.

No new tables, no cron jobs, no auto-delete logic, no dashboard changes.

### Verification

```
npm run build   -> 0 errors
npm test        -> 1536 passed, 116 suites (no regressions)
npm run lint    -> 0 warnings
npm run format  -> clean
```

### Files

```
src/engine/interfaces/whatsapp-engine.interface.ts  | +5
src/engine/adapters/message-mapper.ts               | +5
src/engine/adapters/message-mapper.spec.ts           | +20 (3 tests)
src/modules/session/session.service.ts               | +15
src/modules/session/session.service.spec.ts           | +20 (2 tests)
CHANGELOG.md                                         | +3
.env.example                                         | +3
```

### Open questions

**Should this also affect `onHistoryMessages`?** When a session syncs its history at connect, ephemeral messages predating the sync would still be persisted. I left it out because history sync is a one-time event. Happy to add it.

**Webhook consumers.** Setting this toggle means webhook listeners will not receive ephemeral messages. That is the feature, but it is worth documenting. Already in `.env.example`.


rmyndharis (changes requested):
Thanks for this — a clean opt-in toggle in the spirit of `MEDIA_DOWNLOAD_ENABLED`, and the early-return next to `isStatusBroadcast` is the right place for it. A few things to address before merge:

**1. The Baileys engine isn't covered — the toggle silently no-ops there.** The guard keys off `message.ephemeralDuration`, but you populate that field only on the wwebjs path: the read was added to `buildIncomingMessageBase` (`message-mapper.ts`), which the Baileys adapter doesn't use. Baileys builds its `IncomingMessage` via `buildIncomingMessageFromBaileys` (`baileys-message-mapper.ts`), which never sets `ephemeralDuration`. So on the Baileys engine, `STORE_EPHEMERAL_MESSAGES=false` will still store and dispatch everything.

One subtlety worth knowing for the Baileys fix: the live inbound `WebMessageInfo.ephemeralDuration` is typically empty there — the per-chat timer lands on the message's `contextInfo.expiration` instead (confirmed against the `baileys@6.7.23` source; #513 surfaces this on the Baileys session store). For the toggle to behave the same on both engines, the Baileys mapper needs to populate `ephemeralDuration` from `contextInfo.expiration`.

**2. Duplicate block in `.env.example`.** The `STORE_EPHEMERAL_MESSAGES` comment block is added twice — looks like a copy-paste slip. Please drop one copy.

**3. CHANGELOG will conflict on rebase.** `[Unreleased]` has moved since you branched, and your entry lands at the same spot as a recent one, so a rebase will need a small reconcile. The credit line is correct — please keep the "Thanks @spidgrou."

On your open questions:

- **`onHistoryMessages`** — I'd leave it out as you have. Keeping this a forward-looking ingestion filter is the cleaner scope; folding in history sync can be a follow-up if anyone needs it.
- **Webhook consumers not receiving ephemeral messages** — agreed that's the intended behavior, and yes, it should be called out beyond `.env.example`. Please also note it where the message/webhook flow is documented (the env table in the README and the webhook docs), so an operator enabling the toggle understands that downstream listeners and the dashboard both stop seeing those messages.

The wwebjs path, the interface addition, and the tests all look good. Once the Baileys path is covered, the duplicate line is removed, and the behavior is documented, I'm happy to take another pass.

spidgrou:
Thanks for the thorough review! Addressed everything:

1. **Baileys covered** — `ephemeralDuration` now populated from `contextInfo.expiration` in `baileys-message-mapper.ts` + extracted in the adapter
2. **Duplicate removed** from `.env.example`
3. **Documentation** added under `message.received` in the webhook events spec (§6.6)
4. **3 unit tests** added for Baileys ephemeralDuration

All checks pass: 1539 tests, build clean, format/lint clean.

Ready for another pass 👀

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **Disappearing-messages support (Baileys engine).** Outbound messages now honor a chat's disappearing-messages timer: the Baileys adapter reads the chat's cached `ephemeralExpiration` and sets it on each send (text, media, and replies), so recipients no longer see _"This message won't disappear — the sender may be using an older version of WhatsApp."_ The timer is applied only when a positive value is known for the chat; when it's unknown, disabled, or not yet synced, the per-message expiration is omitted, exactly as before. Reactions, deletes/revokes, and status posts are unaffected. Thanks @ulises2k for the report. (#473)
 
+- **Selective skip for disappearing messages.** New `STORE_EPHEMERAL_MESSAGES` env var (default `true`). Set to `false` to skip persisting and dispatching incoming disappearing messages (those with `ephemeralDuration > 0`) — no DB insert, no webhook dispatch, no websocket event. Backward compatible; existing deployments are unaffected. The `ephemeralDuration` field is also surfaced on `IncomingMessage` for consumers that want to handle it themselves. Thanks @spidgrou. (#506)
+
 ## [0.7.10] - 2026-06-28
 
 ### Added
```
tlhcelik, line 18: 👏

rmyndharis:
Thanks for turning this around so quickly, @spidgrou! The dup `.env.example`, the `IncomingMessage` field, the session-service guard, and the docs note are all good.

I pushed a small follow-up directly to your branch (maintainer edit) to close out the Baileys path — hope that's alright; happy to walk through any of it:

1. **Disappearing timer now unwraps the `ephemeralMessage` envelope.** A live disappearing message arrives wrapped (the same wrapper the media path a few lines up already normalizes), so reading `contextInfo.expiration` off the **raw** `content` returned `undefined`, and `STORE_EPHEMERAL_MESSAGES=false` silently no-op'd on the actual disappearing messages. The context extraction now reads off `normalizeMessageContent(content)`, so both the quote context and the timer resolve for wrapped messages — consistent with how #473/#513 source the timer from inbound messages.
2. **Regression test** for an `ephemeralMessage`-wrapped inbound message → `ephemeralDuration`, plus a per-test reset of the `normalizeMessageContent` mock (it was leaking between cases once the context path started reading through it).
3. **Events-table doc fix** — the added note had introduced a stray `||` and a blockquote in the middle of the table, which split it; moved the note just below the table.

Full backend suite green (1540) + lint/build clean. The feature and the credit stay yours — thanks for the fast iteration. Letting CI settle, then this is good to merge from my side. 🙏

spidgrou:
Thanks Yudhi for the fix and the patience — you are absolutely right, I should have dug deeper into the Baileys side instead of going with the surface-level approach. Lesson learned: when a message arrives wrapped, raw content is not enough.

It has been a pleasure to contribute, even if small. Looking forward to the next one! 🙏

Issue #319, opened 2026-06-19
Baileys engine: inbound/sent messages fail to persist — SQLITE_CONSTRAINT: FOREIGN KEY constraint failed (baileys_stored_messages.sessionId → sessions.id)


ulises2k:
## Summary

On the Baileys engine (`ENGINE_TYPE=baileys`), **inbound (and sent) messages repeatedly fail to persist** to the Baileys message store, logging:

```json
{"level":"warn","context":"BaileysAdapter","message":"Failed to persist message to store","error":"SQLITE_CONSTRAINT: FOREIGN KEY constraint failed"}
```

The message is still received and the `message.received` webhook still fires, but nothing is written to `baileys_stored_messages`. Because that store backs **reply / forward / react / delete by id**, those operations can later fail with `Message <id> not found`.

## Environment

- OpenWA **v0.4.1** (`5dd8c28`)
- Engine: **Baileys** (`@whiskeysockets/baileys` 6.7.23)
- Data DB: **SQLite**, `DATABASE_SYNCHRONIZE=false` (migration-managed)
- Single-container deployment, Docker, Raspberry Pi 5 (ARM64)
- DB predates 0.4.0 (upgraded across versions; migrations adopted over a previously `synchronize`-created DB)

## Root cause analysis

`baileys_stored_messages.sessionId` is a FK to `sessions(id)` with `ON DELETE CASCADE` — `src/database/migrations/1781000000000-AddBaileysStoredMessages.ts`:

```
CONSTRAINT "FK_baileys_stored_messages_sessionId" FOREIGN KEY ("sessionId") REFERENCES "sessions" ("id") ON DELETE CASCADE
```

The adapter persists every message with `this.config.sessionId` — `src/engine/adapters/baileys.adapter.ts`:
- line **714** (inbound `messages.upsert`) → logs `Failed to persist message to store`
- lines **325 / 909** (sent) → log `Failed to persist sent message to store`

`BaileysMessageStoreService.put()` then does an `upsert`, which violates the FK when the `config.sessionId` passed in **is not present in `sessions` at insert time**. In practice this happens when the adapter receiving the message belongs to a session row that no longer exists (or isn't committed yet) — e.g. after a session is **recreated with a new UUID** or during **reconnect churn**, an adapter keeps receiving `messages.upsert` events and tries to store them under a stale/absent `sessionId`.

[...]

PR #513, opened 2026-06-29
fix(baileys): source disappearing-messages timer from inbound messages (#473)

ulises2k:
> **Draft.** This builds on #503 and is shared mainly as a **diagnosis + proposed direction**. One step (the exact field the live 1:1 timer lives in) I could not confirm end‑to‑end on my box — details below. Happy to adjust or hand it over.

## TL;DR

After deploying #503 (the disappearing‑messages fix for the Baileys engine) to production, the *"This message won't disappear — the sender may be using an older version of WhatsApp"* notice **still appears** on a real 1:1 chat with a 90‑day timer. I instrumented the running adapter and traced it: **#503 no‑ops in practice**, for two independent reasons.

## Root cause (instrumented on a live session, `v0.7.10 + #503`)

**1. `Chat.ephemeralExpiration` is never populated.** `getEphemeralExpiration` reads `chat?.ephemeralExpiration`, but a debug dump of the in‑memory `chats` map showed **0 of 159 chats** (across two sessions) carrying any `ephemeralExpiration`. Baileys 6.7.23 only emits `EPHEMERAL_SETTING` → `chats.update` on a *fresh* toggle; for a long‑standing timer the post‑reconnect `messaging-history.set` snapshot doesn't carry it, and the in‑memory store is wiped on every restart. So the value the fix depends on is simply not there.

**2. `getEphemeralExpiration` doesn't bridge `@lid` ↔ phone.** Under WhatsApp's LID migration the 1:1 chat is keyed in the store under the contact's `@lid` JID (live `contacts.update` showed the peer as `…@lid`, with `contacts.upsert` carrying a separate `@s.whatsapp.net` id). The lookup does `chats.get(chatId) ?? chats.get(toEngineJid(chatId))`, and `toEngineJid` only folds `@c.us`↔`@s.whatsapp.net` — a `@lid` id passes through unchanged. The outbound send targets the **phone** JID (the neutral `from` OpenWA delivered on the webhook), so both probes miss the `@lid`‑keyed chat → `undefined` → 2‑arg `sendMessage` → notice persists.

(The send wiring itself is correct — a plain‑text reply does route through `withEphemeral`. The gap is entirely on the read/population side.)

## What this PR changes

- **Source the timer from inbound messages**, which every message in a disappearing chat carries, rather than from `Chat.ephemeralExpiration`. Cached in a new `ephemeralByChat` map keyed by **both the raw and neutral JID**, so a send addressed in either dialect (phone or `@lid`) resolves to the same entry — this also closes the `@lid` gap.
- **`getEphemeralExpiration`** now tries the message‑learned timer across raw/engine/neutral keys first, then falls back to `Chat.ephemeralExpiration` for engines/sessions that do surface it (so existing behavior is preserved).
- **+6 unit tests** (inbound‑sourced timer, `@lid`→phone resolution, `contextInfo.expiration` incl. an `ephemeralMessage` envelope, non‑positive‑doesn't‑clear, precedence). Full suite green; lint/prettier clean.

## ⚠️ The one thing I couldn't verify live (why this is a draft)

My first attempt read `WebMessageInfo.ephemeralDuration` — but instrumentation showed it **empty on live 1:1 `messages.upsert`** (it appears to be a group/history‑sync field). So `extractEphemeralDuration` now also reads the per‑message **`contextInfo.expiration`** after unwrapping the ephemeral/view‑once/document envelope, which is my best guess for the live 1:1 source.

I could **not** confirm that extraction end‑to‑end, because on the test box `baileys_stored_messages` was empty — the session was orphaned (`No parent session row … skipping Baileys message store`, the #319 FK‑skip path), so I had no stored raw message to inspect, and I didn't want to keep guessing against production. **Could you sanity‑check where the 1:1 disappearing timer actually lands on an inbound `messages.upsert` in 6.7.23** (`contextInfo.expiration` vs `disappearingMode` vs the `ephemeralMessage` wrapper)? If it's elsewhere, it's a one‑line tweak to `extractEphemeralDuration`.

## Repro

Real LID‑migrated 1:1 chat, disappearing timer = 90 days. An OpenWA‑driven session auto‑replies via the send‑text endpoint; the reply still shows the recipient‑side "older version" notice. Won't reproduce on a contact that resolves purely as `@c.us` (no `@lid`), which is likely why it passed review.

## Test plan

- [x] `baileys-session-store.spec.ts` — 36/36 (6 new)
- [x] `baileys.adapter.spec.ts` — 99/99
- [x] eslint/prettier clean on touched files
- [ ] **Live confirmation of the `contextInfo.expiration` source** (needs a session whose message store isn't orphaned) — the open question above

Thanks for #503 and the quick turnaround on the original report 🙏

### DIFF
--- a/src/engine/adapters/baileys-session-store.spec.ts
+++ b/src/engine/adapters/baileys-session-store.spec.ts
@@ -96,6 +96,82 @@ describe('BaileysSessionStore', () => {
       expect(store.getEphemeralExpiration('628444@s.whatsapp.net')).toBeUndefined();
       expect(store.getEphemeralExpiration('nope@s.whatsapp.net')).toBeUndefined();
     });
+
+    it('learns the timer from an inbound message ephemeralDuration without any chats.* upsert', () => {
+      // The real-world case: Chat.ephemeralExpiration is never populated, but every inbound message in
+      // a disappearing chat carries the live timer. A neutral @c.us send target must resolve it too.
+      store.recordMessage({
+        key: { remoteJid: '628111@s.whatsapp.net', fromMe: false, id: 'M1' },
+        message: { conversation: 'hi' },
+        messageTimestamp: 100,
+        ephemeralDuration: 7776000,
+      });
+      expect(store.getEphemeralExpiration('628111@s.whatsapp.net')).toBe(7776000);
+      expect(store.getEphemeralExpiration('628111@c.us')).toBe(7776000);
+    });
+
+    it('resolves a timer learned under @lid when the send targets the phone jid (#473 LID migration)', () => {
+      store.addLidMappings([{ lid: '41562515988583@lid', pn: '5491169954736@s.whatsapp.net' }]);
+      store.recordMessage({
+        key: { remoteJid: '41562515988583@lid', fromMe: false, id: 'M2' },
+        message: { conversation: 'hola' },
+        messageTimestamp: 100,
+        ephemeralDuration: 7776000,
+      });
+      // matchwa replies to the neutral phone jid OpenWA delivered on the webhook — the prior no-op case.
+      expect(store.getEphemeralExpiration('5491169954736@c.us')).toBe(7776000);
+      expect(store.getEphemeralExpiration('5491169954736@s.whatsapp.net')).toBe(7776000);
+      expect(store.getEphemeralExpiration('41562515988583@lid')).toBe(7776000);
+    });
+
+    it('ignores a non-positive ephemeralDuration and never clears a known timer', () => {
+      store.recordMessage({
+        key: { remoteJid: '628111@s.whatsapp.net', fromMe: false, id: 'M1' },
+        message: { conversation: 'hi' },
+        messageTimestamp: 100,
+        ephemeralDuration: 86400,
+      });
+      store.recordMessage({
+        key: { remoteJid: '628111@s.whatsapp.net', fromMe: false, id: 'M2' },
+        message: { conversation: 'later' },
+        messageTimestamp: 200,
+        ephemeralDuration: 0,
+      });
+      expect(store.getEphemeralExpiration('628111@s.whatsapp.net')).toBe(86400);
+    });
+
+    it('prefers the message-learned timer over a stale Chat.ephemeralExpiration', () => {
+      store.upsertChats([{ id: '628111@s.whatsapp.net', ephemeralExpiration: 604800 }]);
+      store.recordMessage({
+        key: { remoteJid: '628111@s.whatsapp.net', fromMe: false, id: 'M1' },
+        message: { conversation: 'hi' },
+        messageTimestamp: 100,
+        ephemeralDuration: 7776000,
+      });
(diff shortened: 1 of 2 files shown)


rmyndharis (commented):
Excellent diagnosis — thank you for instrumenting this on a live session rather than guessing. Both root causes hold up:

1. **`Chat.ephemeralExpiration` not populated for a long-standing timer.** Consistent with Baileys 6.7.23 only emitting the setting on a fresh toggle; the post-reconnect history snapshot doesn't carry it and the in-memory store is wiped on restart. Sourcing the timer from the messages themselves is the right call.
2. **The `@lid`↔phone gap in `getEphemeralExpiration`.** Real — `toEngineJid` only folds `@c.us`↔`@s.whatsapp.net`, so an `@lid`-keyed chat is missed when the send targets the phone JID. Caching under both the raw and neutral keys closes it cleanly.

**On your open question — `contextInfo.expiration` is the correct field.** I confirmed it against the pinned `baileys@6.7.23` source. The outbound generator stamps the timer there directly:

```js
// node_modules/@whiskeysockets/baileys/lib/Utils/messages.js
innerMessage[key].contextInfo = {
  ...(innerMessage[key].contextInfo || {}),
  expiration: options.ephemeralExpiration || WA_DEFAULT_EPHEMERAL,
};
```

Inbound messages in a disappearing chat carry the same `contextInfo.expiration`, so reading it after unwrapping the `ephemeralMessage` envelope (as you do) is exactly right. `WebMessageInfo.ephemeralDuration` is the history-sync field and is empty on a live `messages.upsert` — your instrumentation matches the source. `disappearingMode` only carries `{ initiator, trigger }`, not the duration, so `expiration` is the one to read.

The fail-safe shape is good too: a non-positive/absent value leaves a known timer untouched, and a single non-ephemeral message never clears it.

This is solid in my view. Please mark it ready for review and I'll do a full pass and merge. One thing to check on the rebase: it stacks on #503, which is now on `main`, so a rebase should be clean — just confirm the `getEphemeralExpiration` fallback still lines up after it.

ulises2k:
Thanks for the thorough review and for confirming `contextInfo.expiration` against the 6.7.23 source — good to know the extraction is reading the right field.

Done and marked ready:
- **Rebased onto `main`** (now `56b518f`). Clean — no new commit since #503 touched `baileys-session-store.ts`, `baileys.adapter.ts`, or `wa-id.ts`, so the `getEphemeralExpiration` fallback to `Chat.ephemeralExpiration` lines up unchanged.
- **Tests green post-rebase:** `baileys-session-store.spec.ts` 36/36, `baileys.adapter.spec.ts` 99/99; eslint/prettier clean on touched files.

The change is isolated to `baileys-session-store.ts` (+ its spec). Over to you for the final pass 🙏

rmyndharis (approved):
Confirmed end-to-end: `contextInfo.expiration` is the right field (verified against the 6.7.23 source), the rebase is clean, and CI is green. Sourcing the timer from inbound messages with the raw/neutral-key caching closes the LID lookup gap that left it unresolved on migrated 1:1 chats. Thanks for the careful instrumentation and the clean follow-through — merging.

Issue #557, opened 2026-06-30
[Feature]: Install On VPS


krishnasharma1337:
### Pre-flight

- [x] I searched existing issues and this isn't already requested

### Problem / motivation

Hello, can you make a .sh file where whatsapp gateway with all setup auto install on fresh new vps, when we run commond.

### Scope

- [x] I understand some features are limited by the underlying WhatsApp engine (e.g. interactive Buttons/List messages are not supported on whatsapp-web.js).

rmyndharis:
Thanks for the request, and thanks @DanielOberlechner for the ready-to-use compose snippet 🙏

This is effectively already covered. The repo ships an official `docker-compose.yml`, and a fresh VPS only needs Docker + this one-liner from the [Quick Start → Docker](https://github.com/rmyndharis/OpenWA#-quick-start) section:

```sh
docker compose up -d
```

That brings up the gateway (SQLite + local storage, zero-config schema on first boot) with the dashboard and API on port 2785. See [Production Deployment](https://github.com/rmyndharis/OpenWA#-production-deployment) in the README for hardening (CORS, reverse proxy, API key). @DanielOberlechner's snippet above is a great minimal starting `.env`/compose pair.

A bespoke auto-install `.sh` would just wrap these same steps, so I'll close this as covered by the existing Docker path. If you hit a specific gap in the compose flow, feel free to open a focused issue.

PR #558, opened 2026-06-30
feat(stats): add chat names to top chats and stats endpoints

buluma:
## Description

Store the human-readable chat name (contact pushName/saved name) on each message at save time, and surface it as `chatName` in the `GET /stats/messages` and `GET /sessions/:id/stats` topChats response. The dashboard uses it to display readable names instead of raw JIDs.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [x] Tests added/updated (1682/1682 pass)
- [x] Documentation updated (CHANGELOG)
- [x] Lint passes
- [x] Self-reviewed

**Before:** Top Chats showed raw JIDs like `254797747792@c.us`, `131207224393887@lid`
**After:** Shows readable names like contact names and group names

## Related Issues
None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Stats endpoint now returns chat names in top chats.** The `GET /stats/messages` and `GET /sessions/:id/stats` endpoints include a `chatName` field on each top-chat entry, populated from the contact's pushName or saved name at message time. The dashboard uses it to show readable names instead of raw JIDs. A backfill script updates existing rows from the current contact/chat list. (#557) Thanks @shadowwalker.
+
 ### Fixed
 
 - **Sending to — or operating on — a WhatsApp Channel (newsletter) no longer logs internal errors.** On the whatsapp-web.js engine a channel JID (`…@newsletter`) resolves to a `Channel`, which has none of the per-chat operations; the gateway now skips those for channels instead of throwing. The typing indicator that precedes a send, the typing/recording presence endpoint and its MCP tool, mark-unread, and delete-chat now cleanly no-op for a channel (presence does nothing; mark-unread and delete-chat report no change) rather than emitting an internal `TypeError`. Fetching chat labels for a channel previously failed with HTTP 500 — it now returns an empty list. Direct chats, groups, and broadcast lists are unaffected. (#554) Thanks @DanielOberlechner.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- **Stats endpoint now returns chat names in top chats.** The `GET /stats/messages` and `GET /sessions/:id/stats` endpoints include a `chatName` field on each top-chat entry, populated from the contact's pushName or saved name at message time. The dashboard uses it to show readable names instead of raw JIDs. A backfill script updates existing rows from the current contact/chat list. (#557) Thanks @shadowwalker.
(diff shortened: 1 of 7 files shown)


rmyndharis (changes requested):
Thanks for this, @buluma — readable names in the top-chats view is a genuinely useful addition, and the migration / entity / dashboard wiring is all in the right places. Before it can land there's one correctness issue in the aggregation that needs fixing, plus a few CHANGELOG/doc corrections.

## Blocking

**1. `GROUP BY m.chatName` fragments the top-chats aggregation.**
Adding `.addGroupBy('m.chatName')` to `getMessageStats` and `getSessionStats` changes the grouping grain from one row per chat to one row per `(chatId, chatName)`. Because `chatName` is not constant for a given chat, a single chat splits into several rows:
- the `COUNT(*)` total is divided across those rows, so the counts are wrong;
- `.limit(10)` then counts rows rather than chats, so the same chat can appear several times and busier chats drop off the list.

This is the normal path, not an edge case:
- outbound messages are persisted by `saveOutgoingMessage` (message.service.ts), which does not set `chatName`, so every 1:1 chat ends up with named inbound rows and `NULL` outbound rows → two buckets;
- legacy rows are `NULL` → another bucket;
- in groups `chatName` is the sending participant, so each member becomes its own bucket.

It reproduces on the default SQLite dev DB. Suggested fix — keep the grain at one row per chat and pick a representative name with an aggregate instead of grouping by it:

```ts
.addSelect('MAX(m.chatName)', 'chatName')
.groupBy('m.chatId')            // drop both .addGroupBy('m.chatName')
```

`MAX()` ignores NULLs and works on both SQLite and Postgres.

## Worth addressing here

**2. For group chats `chatName` is the sender's name, not the group's.** `incoming.contact` is the message sender (the whatsapp-web.js adapter notes *"getContact() gives the real sender (author in groups)"*), so a group's stored name is whoever last spoke. Real group names would need the group subject resolved for `@g.us` JIDs before persisting; otherwise it's worth scoping the feature to 1:1 chats.

**3. Outbound-only chats stay unnamed.** Only the inbound path sets `chatName`, so a chat you only ever send to keeps `NULL` and renders as a raw JID. Setting it in `saveOutgoingMessage` (from the recipient contact/group) would close that gap.

## CHANGELOG / docs

**4.** The entry says *"A backfill script updates existing rows from the current contact/chat list"* — there's no backfill script in the PR, and the migration only adds a NULL column, so existing rows stay unnamed. Please either add the backfill or drop that sentence.

**5.** The credit reads `(#557) Thanks @shadowwalker.` — #557 is an unrelated issue and this PR is #558; it should be `(#558) Thanks @buluma.`

**6.** `docs/06-api-specification.md` still shows the old `topChats` shape — please add the `chatName` field to both response examples.

**7. (nit)** In the migration, `const colType = isPostgres ? 'varchar' : 'varchar'` has two identical branches; you can drop `isPostgres`/`colType` and inline `varchar`.

The build/tests are green, so once the `GROUP BY` is fixed the rest is quick. Happy to help if any of this is unclear.

buluma:
Thanks for the thorough review @rmyndharis — really appreciate you catching the GROUP BY issue.

Here's what I fixed in [9c07b13](https://github.com/buluma/OpenWA/commit/9c07b13):

**Stats aggregation (blocking)** — dropped both `addGroupBy('m.chatName')` and switched to `MAX(m.chatName)` in `getMessageStats` and `getSessionStats`. Keeps the grain at one row per chat, and `MAX()` picks a representative name ignoring NULLs so outbound-only chats still get a name when inbound messages exist.

**Migration** — removed the dead `isPostgres ? 'varchar' : 'varchar'` ternary, just inlined `varchar`.

**CHANGELOG** — fixed the PR ref (`#557` → `#558`), credit (`@shadowwalker` → `@buluma`), and dropped the backfill script line since there isn't one in this PR.

Didn't touch the outbound path or group name resolution — those need contact lookups I don't have the plumbing for here. Happy to tackle them separately if useful.

rmyndharis:
Thanks for this, @buluma — nice feature. Landing it via #572, which carries your commits (authorship preserved) and adds two things needed for current `main`: your migration's timestamp (`1781900000000`) collided with a migration that merged after you opened this (`AddIntegrationFabric`), so it's renamed to `1782000000000`; and a stats-service test covering the `MAX(chatName)` aggregate. Full credit stays yours in the CHANGELOG. Closing this in favor of #572.

PR #622, opened 2026-07-04
feat(message): native WhatsApp polls via send-poll on both engines

alejo117:
Add POST /api/sessions/:sessionId/messages/send-poll (OPERATOR role): question (max 255), 2-12 options (each non-empty, max 100 chars) and an optional allowMultipleAnswers flag, validated in SendPollDto so an out-of-range poll fails as a clean 400 instead of an engine error.

Engine support:
- whatsapp-web.js: dynamic Poll import with the .default fallback (same pattern as Location); the send goes through sendResolved so @c.us ids get the usual lid resolution. wwebjs's typings mark messageSecret as required but at runtime it is optional, so the options object is cast to the constructor's own options type.
- Baileys: poll content with selectableCount 1 (single choice) or 0 (no limit = WhatsApp's 'allow multiple answers'); Baileys generates the poll messageSecret itself.

The message history row stores the poll question as the body (a poll has no plain-text body), keeping the log readable.

Covered by DTO validation specs, MessageService specs and adapter specs for both engines; API reference, collection examples and CHANGELOG updated.

## Description
Brief description of changes

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] Lint passes
- [ ] Self-reviewed

## Related Issues
Closes #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Native WhatsApp polls** via `POST /api/sessions/:sessionId/messages/send-poll`: question, 2–12 options and an optional `allowMultipleAnswers` flag (default single choice), implemented on both engines (whatsapp-web.js `Poll`, Baileys `poll` content with `selectableCount` 1/0). The message history stores the poll question as the body so the log stays readable.
+
 ## [0.8.7] - 2026-07-03
 
 ### Added
--- a/docs/06-api-specification.md
+++ b/docs/06-api-specification.md
@@ -1138,6 +1138,39 @@ Send a sticker (by URL or base64; typically webp). Reuses `SendMediaMessageDto`.
 
 **Errors:** `400` media validation failure / session not active / unknown body field · `401` missing/invalid API key · `403` key role below OPERATOR · `500` engine error
 
+#### POST /api/sessions/:sessionId/messages/send-poll
+
+Send a native WhatsApp poll.
+
+**Auth:** API key (OPERATOR)
+
+**Path parameters**
+
+| Name | Type | Description |
+| --- | --- | --- |
+| sessionId | string | Session ID |
+
+**Request body** — `SendPollDto`
+
+| Field | Type | Required | Constraints | Description |
+| --- | --- | --- | --- | --- |
+| chatId | string | Yes | non-empty | Target chat |
+| name | string | Yes | max 255 | Poll question / title |
+| options | string[] | Yes | 2–12 items, each non-empty, max 100 chars | Options to vote on |
+| allowMultipleAnswers | boolean | No | — | Allow picking several options (default single choice) |
+
+```json
+{ "chatId": "1203630000@g.us", "name": "Where should we meet?", "options": ["Park", "Beach", "Downtown"], "allowMultipleAnswers": false }
+```
+
+**Response** `201`
+
+```json
+{ "messageId": "true_1203630000@g.us_3EB0ABCD", "timestamp": 1719312000 }
+```
+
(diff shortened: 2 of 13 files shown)


Magnarks:
A very interesting addition.

rmyndharis (commented):
Solid work — the DTO validation is thorough, the implementation matches the existing `sendLocation`/`sendSticker` paths, both engines are covered, and the docs/CHANGELOG are updated. I verified the branch locally: lint, build, and the four targeted Jest suites are all green (315 tests). Two small follow-ups I'd like addressed:

**1. `'poll'` isn't a first-class `MessageType`.**
`sendPoll` persists the outgoing row as `type: 'poll'`, but `'poll'` isn't a member of the backend `MessageType` union (`src/engine/interfaces/whatsapp-engine.interface.ts`) or the dashboard's `MESSAGE_TYPES` (`dashboard/src/services/api.ts`), and neither adapter maps it on the receive side. The result is a split representation of one message:

- `GET /messages` and chat history return `type: 'poll'` (the persisted value).
- The live `message.sent` websocket/webhook event is built from `onMessageCreate → mapWwebjsMessageType`, which has no poll case, so it reports `type: 'unknown'`.
- The dashboard coerces the persisted `'poll'` to `'unknown'` via `asMessageType()`.

It's not a crash or data loss (the column is unconstrained `varchar` and the `📊 <question>` body still renders), but it breaks an invariant the codebase deliberately maintains — `api.ts` documents that persisted rows, the `message.sent` payload, and the websocket all share these type values. Please add `'poll'` to the backend union and the dashboard `MESSAGE_TYPES` (plus a label/color), and ideally map the inbound `poll_creation` / `pollCreationMessage` token in both adapters for full parity.

**2. Test the headline behavior.**
The new `sendPoll` service specs assert only the `sendPollMessage` call args and the returned `messageId` — they never assert the persisted row. The distinctive behavior this PR introduces (persisting `type: 'poll'` and rewriting the body to `📊 ${name}`, which the CHANGELOG highlights) has no test guarding it. The sibling send paths establish the convention — e.g. the voice/audio/forward tests assert `expect(repository.create).toHaveBeenCalledWith(expect.objectContaining({ type: 'voice' }))`. Please add the equivalent for the poll:

```ts
expect(repository.create).toHaveBeenCalledWith(
  expect.objectContaining({ type: 'poll', body: '📊 Where should we meet?' }),
);
```

Neither is a blocker and the build is green, so this is close. Thanks for the contribution!

alejo117:
Sorry for the late reply — a few things came up and I hadn't been able to get back to GitHub until now.

Thank you so much for the thorough review and for merging the PR! Both follow-up points are completely fair: I focused on the send path and missed that `'poll'` wasn't part of the `MessageType` union on the receive side, and the persistence test is a convention I should have followed from the sibling send paths. I really appreciate you addressing both yourself in #626 — I just read through it and it's exactly what was needed.

While working on this I noticed a couple of other gaps, and I checked against current main (v0.8.12) to make sure they're still relevant — would either of these be welcome as a follow-up PR?

1. **`quotedMessageId` on send-media** — the reply endpoint currently only supports text, so there's no way to send an image/media as a reply to an existing message. I'd add an optional `quotedMessageId` to `SendMediaMessageDto` and wire it through both engines, following the same pattern as the text reply path. Here's what it looks like in practice (I already run this in my fork, so the PR would be porting a tested feature):

image

2. A **`POST /sessions/:id/logout` endpoint** — I saw #656 now fully purges the auth directory on session delete, which covers re-linking. But would a logout that unlinks the phone without deleting the session (keeping its config and webhooks) be useful? The engine interface already defines `logout()`, it's just not exposed via the API. If you feel delete-and-recreate already covers it, no worries.

Happy to start with whichever (or neither!) makes sense for the roadmap. Thanks again!

PR #624, opened 2026-07-04
feat/custom-select-dropdowns

haseeblodhi1899:
## Description
Replaces native HTML `<select>` elements in the dashboard with a reusable
`CustomSelect` component that matches the existing design system (colors,
typography, spacing, border radius, animations, dark/light themes).

### Changes
- **New:** `CustomSelect` component with keyboard navigation, ARIA
  accessibility, theme support, responsive design, RTL support
- **Sessions:** "All Status" filter → CustomSelect
- **Logs:** "All Severities" filter → CustomSelect  
- **Login:** Language picker → CustomSelect
- No changes to sidebar (already uses custom menus)

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor (UI consistency, no breaking changes)

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] Lint passes (0 errors, 2 pre-existing warnings)
- [x] Build passes (tsc + vite)
- [x] Self-reviewed

image
image
image

## Related Issues
Closes #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Dashboard `<select>` elements replaced with a custom dropdown component.** The "All Status" filter (Sessions), "All Severities" filter (Logs), and language picker (Login) now use a reusable `CustomSelect` component that matches the dashboard design system with proper dark/light theming, keyboard navigation (arrows, Home/End, type-ahead, Escape), and responsive behavior. Focus returns to the trigger on close, matching native `<select>` semantics. Thanks @haseeblodhi1899.
 - The **"Install a plugin" modal is wider on desktop** (480px → 680px) to give the plugin catalog list more room, while still collapsing to a full-width bottom sheet on small screens.
 
 ### Fixed
--- a/dashboard/src/components/CustomSelect.css
+++ b/dashboard/src/components/CustomSelect.css
@@ -1,0 +1,128 @@
+.custom-select {
+  position: relative;
+  display: flex;
+  min-width: 0;
+}
+
+.custom-select-trigger {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  gap: 0.5rem;
+  width: 100%;
+  padding: 0.75rem 0;
+  border: none;
+  background: transparent;
+  color: var(--text-primary);
+  font-size: 0.9375rem;
+  font-family: inherit;
+  cursor: pointer;
+  white-space: nowrap;
+  overflow: hidden;
+}
+
+.custom-select-trigger:hover {
+  color: var(--text-primary);
+}
+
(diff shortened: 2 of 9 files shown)


rmyndharis (changes requested):
Thanks for this — the component matches the design system nicely and I verified the branch locally (tsc + vite build, lint, and i18n:check all pass). One accessibility regression needs fixing before merge, since this replaces a native `<select>`:

**Focus is dropped to `<body>` on every selection and on Escape.**

While the dropdown is open, real DOM focus is moved onto an option `<button>` via the effect:

```tsx
useEffect(() => {
  if (isOpen && focusedIndex >= 0 && listRef.current) {
    const items = listRef.current.querySelectorAll<HTMLButtonElement>('.custom-select-option');
    items[focusedIndex]?.focus();
  }
}, [isOpen, focusedIndex]);
```

On `close()` (selecting an option, Escape, or Tab) `isOpen` becomes `false`, which unmounts the entire `{isOpen && (...)}` dropdown subtree — including the option button that currently holds focus. Nothing returns focus to the trigger (there's no `triggerRef`), so the browser resets focus to `document.body`.

Impact: a keyboard/screen-reader user opens the filter, arrows to an option, presses Enter — and is silently dumped to the top of the page. The next Tab starts from the first focusable element and the screen reader loses its reading position. A native `<select>` returns focus to the control after selection/Escape; this is a WCAG 2.4.3 (Focus Order) regression, and it happens on all three call sites (Sessions, Logs, Login) on every selection.

Fix (~5 lines): add a ref to the trigger button and call `triggerRef.current?.focus()` inside `close()`.

Non-blocking nice-to-haves:
- Restore native type-ahead (first-letter jump) and Home/End key support — the native `<select>` this replaces had them.
- The trigger sets `aria-activedescendant` while real DOM focus roves to the option buttons. Those are two different, mutually-exclusive ARIA patterns — pick the roving-focus one you already use and drop `aria-activedescendant` when you touch this.

Once focus returns to the trigger on close, this is good to go.

haseeblodhi1899:
Thanks for the thorough review!

I added a `triggerRef` and called `.focus()` on it inside `close()`, so focus returns to the trigger button instead of falling back to `<body>` on close.

Also addressed both suggestions:
- Dropped `aria-activedescendant` since real DOM focus is already the source of truth
- Added type-ahead (first-letter jump) plus `Home`/`End` support to match native `<select>` behavior

Also resolved the CHANGELOG conflict.

rmyndharis (approved):
Verified the rebased branch locally — dashboard build, unit tests (84/84), and lint all pass, and focus now returns to the trigger on close/select/Escape (WCAG 2.4.3 resolved), with the type-ahead and Home/End additions in place. Rebased onto main (CHANGELOG merged) and CI is green. Thanks for the thorough follow-up, @haseeblodhi1899 — merging.

PR #643, opened 2026-07-06
Wa users/wa usernames and phonenames

alexander-pyrforos:
## Description
Added phone, lid/jid, name of user in message header and API responces
## Type of Change
New feature

### DIFF
--- a/dashboard/src/pages/Chats.css
+++ b/dashboard/src/pages/Chats.css
@@ -412,6 +412,17 @@
   white-space: pre-wrap;
 }
 
+.chats-page .message-sender-name{
+  font-size: 0.75rem;
+  font-weight: 600;
+  margin-bottom: 0.2rem;
+  color: var(--primary, #25d366);
+}
+
+.chats-page .message-bubble-wrapper.incoming .message-sender-name{
+  color: var(--text-secondary, #8696a0);
+}
+
 .chats-page .message-meta{
   display: flex;
   align-items: center;
--- a/dashboard/src/pages/Chats.css
+++ b/dashboard/src/pages/Chats.css
@@ -417,6 +417,7 @@
   font-weight: 600;
   margin-bottom: 0.2rem;
   color: var(--primary, #25d366);
+  word-break: break-word;
 }
 
 .chats-page .message-bubble-wrapper.incoming .message-sender-name{
--- a/dashboard/src/pages/Chats.tsx
+++ b/dashboard/src/pages/Chats.tsx
@@ -915,18 +915,46 @@ export function Chats() {
                       const isRevoked = msg.type === 'revoked';
                       const isMasked = msg.type === 'masked';
 
-                      // Resolve sender display name for incoming group messages.
+                      // Resolve sender info for incoming group messages.
                       // For 1:1 chats the chat header already identifies the contact.
-                      // Priority: contact.name > contact.number (E.164 phone) > pushName > senderPhone > author/from
-                      const senderLabel =
-                        !isMe && (msg.isGroup || activeChat?.isGroup)
-                          ? msg.contact?.name ||
-                            msg.contact?.number ||
-                            msg.contact?.pushName ||
-                            msg.senderPhone ||
-                            msg.author?.split('@')[0] ||
-                            msg.from?.split('@')[0]
-                          : undefined;
+                      // Always show something: name · phone, or just phone, or LID/JID as last resort.
+                      const senderInfo = !isMe && (msg.isGroup || activeChat?.isGroup || msg.author)
+                        ? (() => {
+                            const c = msg.contact;
+                            const phone = msg.senderPhone || c?.number;
+                            const rawJid = msg.author || msg.from;
+                            const rawId = rawJid?.split('@')[0] ?? '';
+                            const jidSuffix = rawJid?.split('@')[1] ?? '';
+                            const pushName = c?.pushName;
+                            const name = c?.name;
+                            const isLid = jidSuffix === 'lid';
+                            const isGroupJid = jidSuffix === 'g.us';
+                            // Is phone a real phone number (not just the LID digits)?
+                            const realPhone = phone && phone.replace(/^\+/, '') !== rawId
+                              ? (phone.startsWith('+') ? phone : `+${phone}`)
+                              : null;
+                            const parts: string[] = [];
+                            // Push name is the WhatsApp display name — always useful and distinct from phone
+                            if (pushName) {
+                              parts.push(pushName);
+                            }
+                            // contact.name is often the phone number with + — skip if it matches realPhone
+                            if (name && name !== realPhone && name !== `+${rawId}` && name !== pushName) {
+                              parts.push(name);
+                            }
+                            if (realPhone) {
+                              parts.push(realPhone);
+                            } else if (isLid) {
+                              // LID sender whose phone couldn't be resolved — show LID tag
+                              parts.push(`LID:${rawId}`);
+                            } else if (!isGroupJid && rawId) {
+                              // @c.us sender — rawId IS the phone number
+                              parts.push(`+${rawId}`);
+                            }
+                            const result = parts.length > 0 ? parts.join(' · ') : undefined;
+                            return result;
+                          })()
+                        : undefined;
 
                       return (
                         <div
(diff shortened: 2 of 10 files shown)


rmyndharis (changes requested):
Thanks for this, @alexander-pyrforos — the core idea here is spot-on, and there are a couple of genuinely good fixes in here. Wanted to call those out before the change requests.

**What's right:**
- **Resolving the sender via `author` instead of the group JID** is fundamentally correct. The previous behavior (enriching against the group contact) was a real bug, and the new test (`resolves group message sender contact via author, not group JID`) nails the intent.
- The LID-number detection (`contactId.endsWith('@lid') && merged.number === contactId.replace(/@lid$/,'')`) is precise — it correctly distinguishes "number is empty" from "number got filled with LID digits". Good eye.
- The dashboard `senderInfo` resolver is thorough (pushName · name · phone, with sensible LID/JID fallbacks).

Requesting changes for the reasons below — happy to help work through any of them.

---

**🛑 The `+` prefix breaks the `lid_mappings` reverse lookup (silent regression)**

`resolveContactPhone` now returns `+628999` (with `+`), but that value flows into the persisted lid→phone table:

```ts
// session.service.ts — resolveSenderPhone()
incoming.senderPhone = await this.resolveSenderPhone(id, incoming.author ?? incoming.from);
// ...
void this.lidMappingStore?.remember(userPart(contactId), phone, sessionId) // phone = "+628999"
```

But a *different* writer of the same table writes bare digits:

```ts
// baileys-session-store.ts — when a contacts.upsert carries a lid→pn mapping
void this.lidStore?.remember(userPart(lidJid), userPart(pnJid), this.sessionId) // "628999"
```

And the reverse-lookup reader (the message `from` filter) keys on bare digits:

```ts
// message.service.ts — resolveJidCandidates()
const phone = userPart(value);                       // "628999"
for (const lid of this.lidMappingStore.lidsForPhone(phone)) { ... }  // looks up key "628999"
```

So after this change, lid→phone rows written via the inline path get stored under `phoneToLids["+628999"]`, while every reader looks up `"628999"` — a silent miss. The `from` filter for LID senders quietly stops matching.

The `remember(...)` test update in `session.service.spec.ts` (`'628111222333'` → `'+628111222333'`) makes the test pass but doesn't fix the downstream reader.

**Suggested direction:** pick one canonical format for `lid_mappings.phone` and hold it everywhere. The column is documented as *"E.164 phone digits"* and the Baileys session store writes bare digits, so the cleanest path is to keep the engine-boundary contract as *bare digits* (matching the existing `IncomingMessage.senderPhone` / `MessageContact.number` docs) and add the `+` only at the presentation layer (dashboard display / an API DTO field) if you want E.164 there. If you do want `+` to be the new standard, it needs a coordinated migration: `baileys-session-store.resolvePhone`, `userPart` consumers, the from-filter reader, a CHANGELOG `BREAKING` entry, and tests for the reverse-lookup path — otherwise it's a partial migration.

**🛑 `IncomingMessage.senderPhone` contract change is undocumented**

The interface documents `senderPhone` as *"Best-effort phone number (MSISDN digits of the sender)"* (`whatsapp-engine.interface.ts:98`), and `MessageContact.number` as *"Phone digits, best-effort"`. Webhook consumers and SDK users persist/compare these values. Changing the format to include `+` is a silent breaking change for them — no CHANGELOG note, no major bump, no flag.

If `+` is the intended new shape, let's make that explicit in the interface docs, the CHANGELOG, and ideally gate it or call it out in the migration guide.

**🔁 Contact enrichment is duplicated between the live and history handlers**

The LID-number-fix + `resolveContactPhone` fallback block (~40 lines) is pasted near-identically in the live `message` handler (`whatsapp-web-js.adapter.ts:~466`) and the `getChatHistory` handler (`:~1618`). Worth extracting into a small helper like `resolveSenderContact(msg)` so the two paths can't drift apart.

**🎨 Formatting nits (will trip the Prettier/format CI gate)**
- `this.client.on('message', ...)` indentation drifted from 4→5 spaces (`adapter.ts:458`) — looks like an accidental re-indent.
- The `className={\`message-bubble ...\`}` block in `Chats.tsx` lost its alignment.
- New `.css` files are missing a trailing newline.

Minor, but worth a `npm run format` pass before re-push.

---

Thanks again for tackling the group-sender resolution — that part is a real improvement. Once the phone-format consistency is sorted, this is in good shape.

rmyndharis:
Following up on my earlier review — I owe you a correction before anything else.

In that review I called the author-resolution "fundamentally correct" and "a real bug fix," and treated the `+` prefix as a fixable side issue with the PR otherwise in good shape. On a closer read of the library source and the history, I had that backwards, and it changes the recommendation. Let me separate what I got wrong from what still stands.

**Correction: the author-resolution isn't fixing a bug.**

I said the old behavior "enriched against the group contact." It didn't. In whatsapp-web.js, `getContact()` is `getContactById(this.author || this.from)` — for an incoming group message `this.author` is the participant, so it already returned the **participant's** contact. The live handler on `main` already calls `msg.getContact()` (there's a comment right there saying so), and the participant `author` JID + `contact{name,pushName}` were added to the incoming/webhook payload in PR #223, which closed the original "missing fields" issue #146 about three weeks before this branch. So this looks like the fork is simply behind `main`; both adapters already resolve group senders correctly, and there's no parity gap. My earlier praise for that part was misplaced — apologies for the noise.

(The `getChatHistory` path is a slightly different story: on `main` it deliberately does *not* call `getContact` per message, for perf — so history rows aren't contact-enriched today. But that's "no enrichment by design," not "enrichment against the wrong contact," and re-adding it is a perf-sensitive decision rather than a bug fix — see #3.)

**What still stands from my earlier review (and is actually sharper than I first put it).**

The `+`-prefix concerns haven't changed, and digging in they're worse than I stated:

1. **It silently breaks the @lid ↔ @c.us reverse lookup.** `resolveSenderPhone` persists the resolved phone straight into the lid-mapping table — `remember(userPart(contactId), phone, …)` in `session.service.ts` — and the store keys its reverse map by that phone **verbatim** (`lid-mapping-store.service.ts`, `index()` / `lidsForPhone()`). The reader looks it up via `userPart(filter)`, and `userPart` (`wa-id.ts`) strips only the `@domain` and `:device` — never a leading `+`. So a phone stored as `+628999` is looked up as `628999` and silently misses (`message.service.ts`, `resolveJidCandidates`), breaking `GET /messages?from=<phone>` for LID senders learned this way. Worse than I said before: the store is last-write-wins **with eviction**, so this is flaky by write-order rather than a clean break — and the specs hide it from CI (the writer-spec asserts `+…`, the reader-spec mocks the return value; nothing wires the two together end-to-end).

2. **It's an undocumented breaking change for consumers.** `senderPhone` / `contact.number` reach webhook consumers verbatim (`webhook.service.ts`) and the REST `GET /:contactId/phone` returns the value directly (`contact.controller.ts`), where the contract — interface docstrings, the Swagger description, the published spec example — is "MSISDN digits," no `+`. Concrete breakage: the send `phoneNumber` input is `@Matches(/^[0-9]{6,15}$/)`, so a consumer that round-trips a `+`-prefixed `senderPhone` into a send gets rejected. No CHANGELOG entry accompanies the change.

3. **One thing I didn't flag before: history-path perf.** As above, `getChatHistory` skips per-message `getContact` on purpose; this PR adds a `getContactById` per history row — a round-trip per message on a path that can return hundreds or thousands.

4. **Duplication (flagged before, still applies).** The ~30-line LID-number-fix + `resolveContactPhone`-fallback block is pasted near-identically in the live handler and the history handler (indentation already drifting between them). If history enrichment is wanted, extract a shared `resolveSenderContact(msg)` helper so the two paths can't diverge.

**Where this leaves us — close and resubmit the dashboard piece.**

The genuinely valuable part is the one I underweighted last time: the dashboard sender-name display. `Chats.tsx` on `main` renders incoming group messages with no sender label at all, and your `senderInfo` resolver (pushName · name · phone, with LID/JID fallbacks) is the right fix for that. But it's bundled with an unmergeable format change and a redundant backend rewrite, so I'd rather close this and see that piece land on its own.

- Resubmit **just the dashboard piece**: `senderInfo` + CSS + the new `ChatMessage` / `EngineHistoryMessage` fields + the `mergeChatMessages` field-preservation. The dashboard can prepend `+` itself for display (your `senderInfo` already normalizes `phone.startsWith('+')`), so no engine change is needed.
- Drop the `resolveContactPhone` `+` change. If E.164 is genuinely the canonical shape you want, that's a coordinated migration (store reader, `userPart` consumers, the from-filter, a `BREAKING` CHANGELOG note, a major bump) — not a side change tucked into a feature PR.

Thanks for your patience with the back-and-forth — happy to review the focused version.

PR #650, opened 2026-07-06
feat(search): global message search across sessions via Meilisearch

alexander-pyrforos:
Summary

Adds an optional global message-search feature: search messages across all sessions through Meilisearch, with real-time sync and a search panel in the /chats sidebar. Meilisearch is an opt-in external service — when it isn't configured, the search endpoints return 501 and the rest of the application is entirely unaffected (no new hard dependencies, no SQL fallback, no queue).

Design choice: Approach A — Meilisearch optional with fire-and-forget sync hooks. Sync never blocks the message pipeline (all indexing is best-effort with .catch(() => {})).

Backend

- New src/modules/search/ module: SearchModule, SearchService, SearchController, DTOs, and a MeilisearchClient wrapper.
- Config (src/config/): MEILISEARCH_URL, MEILISEARCH_API_KEY, MEILISEARCH_INDEX_PREFIX (default openwa_), with env validation. Absent MEILISEARCH_URL → search disabled, logged once, no crash.
- Index openwa_messages: searchable body only; filterable sessionId, chatId, from, to, type, direction, status, hasMedia; sortable timestamp, createdAt.
- Real-time sync (fire-and-forget, never breaks sends/receives):
  - MessageService: index on send, delete on delete.
  - SessionService: index incoming messages on the live onMessage handler (whatsapp-web.js persists these via .insert(), so the hook lives here, not in MessageService); re-index on revoke so revoked messages stop matching their old body.
  - indexMessageByWaId(sessionId, waMessageId) re-fetches the row after .insert() (which doesn't populate DB-generated id/createdAt) before indexing.
- Endpoints:
  - GET /messages/search?q=&sessionId=&chatId=&from=&type=&direction=&hasMedia=&limit=&offset= → { hits, total, limit, offset } with _formatted.body carrying <mark> highlights; sorted createdAt:desc. 501 when Meilisearch isn't configured.
  - POST /messages/search/reindex (ADMIN role only) → backfills/re-indexes all messages in batches of 1000. 501 if not configured, 403 if the key role is below ADMIN.

JavaScript SDK

- messages.search(params) and messages.search.reindex() added (sdk/javascript/src/resources/messages.ts), with SearchHit / SearchResult / SearchParams types in types.ts, exposed on the client.

Dashboard (/chats)

- New GlobalSearch panel in the chats sidebar (toggled via a Chats / Messages tab), styled to match the existing sidebar patterns (.chat-search-input for the bar, .chat-item-card for result rows, .chats-list scroll region).
- Snippet: shows the matched line + up to 2 lines of context above and 2 below, with the found line highlighted (colored bar + tinted background) and the matched term wrapped in <mark>. Match-line detection is robust — it uses Meilisearch's <mark> when present, otherwise locates the query term in the body itself (case-insensitive) and self-highlights, so the snippet always centers on the real match.
- Thumbnails: real image/video/sticker previews are lazy-loaded via chat history (includeMedia), cached per (session, chat) so each chat fetches at most once; audio/voice/document and pending states use a styled type-tinted tile.
- Click-to-jump: clicking a result switches to the hit's session + chat and scrolls the specific message into view (centered) with a brief highlight flash. A pendingNav resolver effect walks the async stages (session switch → chat activation → thread render); falls back to a toast if the session isn't available or the message is outside the loaded history window.
- i18n keys added to en.json (other locales fall back to en, matching the existing search keys).

Tests

- search.service.spec.ts — unit tests covering isAvailable, indexMessage, deleteMessage, indexMessageByWaId, search (filter mapping, defaults, highlight passthrough, 501), and reindexAll (batching, empty, 501).
- test/search.e2e-spec.ts — e2e covering 501/401 when disabled and 200/400/403/200 when enabled (Meilisearch client mocked per app boot, ValidationPipe mirrored with enableImplicitConversion).
- test/__mocks__/meilisearch.ts — ESM stub for ts-jest (the meilisearch package is ESM-only); wired via jest moduleNameMapper.
- Existing message.service.spec.ts / session.service.spec.ts updated with a SearchService mock to satisfy the new @Optional dependency.

How to enable / verify

1. Run Meilisearch, e.g. docker run -d --name meilisearch -p 7700:7700 getmeili/meilisearch:latest.
2. Set MEILISEARCH_URL=http://localhost:7700 (and MEILISEARCH_API_KEY if your instance requires it). Restart the API.
3. Backfill existing messages once: POST /api/messages/search/reindex with an ADMIN key (the seeded data/.api-key is ADMIN). Incoming/sent messages index in real time thereafter.
4. Dashboard: open /chats, switch to the Messages tab, search. Click a result to jump to the message.
5. Unit + e2e: npx jest src/modules/search and npx jest test/search.e2e-spec.ts. Dashboard: npx tsc --noEmit && npx eslint src/components/GlobalSearch.tsx src/pages/Chats.tsx.

Notes / out of scope

- Typo tolerance uses Meilisearch defaults. Intentionally left as-is; the snippet highlights the verbatim match when present.
- No design docs are included in this PR (kept out of the repo).
- Only new optional dependency: meilisearch (server-side). No BullMQ, no SQL search fallback.


rmyndharis (changes requested):
Really nice work on the UX side here, @alexander-pyrforos — the snippet rendering with matched-line highlighting, the click-to-jump navigation through the async session→chat→render stages, and the graceful 501 fallback are all genuinely well-built. The fire-and-forget indexing principle (never block the message pipeline) is the right instinct too. Thank you for putting this together.

I'm requesting changes for two reasons: a couple of concrete bugs, and a larger architectural question I'd like us to settle before this lands. Happy to pair on any of it.

---

## Architectural question: should the indexing be a plugin?

I want to raise this gently because it shapes the rest of the review. OpenWA already ships a hook bus and a plugin system that were built *precisely* for this kind of cross-cutting extension. The hooks this PR needs already exist:

- `message:sending` / `message:sent` → index on outbound
- `message:received` → index on inbound

…and the plugin SDK exposes everything else the indexer needs (`ctx.net.fetch` for Meilisearch with a built-in SSRF guard, `ctx.storage` for index state, `ctx.engine.getChatHistory` for the thumbnail backfill). `PluginType.EXTENSION` is documented for exactly this category ("general extensions"), and `sessionScoped: false` is designed for global plugins like metrics/search.

What this PR does instead is wire the indexing directly into the core message and session services:

```ts
// message.service.ts (core) — now imports/depends on SearchService
constructor(..., @Optional() private readonly searchService?: SearchService) {}

// session.service.ts (core) — indexes inbound in the live onMessage handler
```

…and adds `meilisearch` to `dependencies` for every install, even though the feature is opt-in.

That said — I'll be the first to acknowledge the nuance: **the query side** (the `GET /messages/search` controller, the dashboard panel) can't fully live in a plugin today, because plugins can't register HTTP routes yet (the capability surface is hooks + `ctx.*` methods, no `registerRoute`). So realistically this splits into two:

- **Indexing sync** → hook-based plugin (`EXTENSION`, `sessionScoped: false`, `net.fetch` to Meilisearch). No core edits, no `@Optional` dependency leaking into `MessageService`/`SessionService`, and every existing test stays un-coupled.
- **Query API + dashboard panel** → legitimately wants to stay in core (needs a controller + UI).

I'd love to hear your thoughts on whether you'd be open to splitting it that way. If there's appetite, I can sketch the plugin manifest and hook registration so the indexer side is a clean drop-in. This isn't a hard blocker on the idea — it's more that this codebase deliberately built the hook system so core services wouldn't accumulate indexing/concern-of-the-week branches, and I'd like to honor that if we can.

---

## Concrete bugs (these need fixing either way)

**🛑 Duplicate imports in the JS SDK — this will fail `tsc`**

In `sdk/javascript/src/resources/messages.ts`, the `import type { ... }` block now declares several types twice — `SendBulkRequest`, `SendContactRequest`, `SendLocationRequest`, `SendMediaRequest`, `SendPollRequest`, `SendTemplateRequest`, `SendTextRequest`, `SuccessResult`, `ForwardMessageRequest`, `ListMessagesQuery`, `MessageHistoryQuery`, `MessageListResponse`, `MessageResponse`, `ReactionRecord`, `ReactMessageRequest`, `ReplyMessageRequest` are added as new `+` lines on top of an existing block that already imports them a few lines down. TypeScript will error with `Duplicate identifier`, and the `sdk-ci` workflow (which keys off `src/**/dto/**` + the engine interface) will go red.

Looks like the new `SearchQuery` / `SearchResult` lines were meant to slot into the existing import, but the rest got duplicated. Trimming to just the two new types should fix it.

**🛑 `meilisearch` is in `dependencies`, not `optionalDependencies`**

`package.json` adds `"meilisearch": "^0.59.0"` to `dependencies`, which contradicts the PR description's *"no new hard dependencies"*. Every OpenWA install now resolves and downloads the package even when search is never configured.

Two clean options:
- Move it to `optionalDependencies`, or
- Dynamic-import it inside `SearchService` (`const { MeiliSearch } = await import('meilisearch')`) when the feature is enabled, so it's never loaded (or required to be installed) for instances that don't use search. The `test/__mocks__/meilisearch.ts` wiring you already added stays as-is.

---

## Smaller things

- **Merge conflict.** `mergeable: CONFLICTING` — needs a rebase against `main`.
- **`GlobalSearch.tsx` uses `dangerouslySetInnerHTML`** for the highlighted snippet. You *did* handle this correctly — `escapeHtml()` runs before any `<mark>` injection, so it's XSS-safe today. My only ask: since it's a regression-prone spot (a future edit to `highlightQueryTerm` that forgets to escape is an XSS), it'd be good to factor the escape+highlight into one small unit-tested helper with a clear "order matters" comment, so the invariant is hard to break.
- **Thumbnail backfill cost.** `useSearchThumbnails` calls `sessionApi.getChatHistory(sessionId, chatId, 50, /*includeMedia*/ true)` per distinct chat — for a query that spans many chats, that's N full-with-media history fetches (base64 payloads) firing from the browser. Worth a cap on concurrent/different-chat fetches, or lazy-loading thumbs only on scroll into view, so a broad search doesn't saturate the client's bandwidth.
- **i18n.** New keys are only in `en.json` — that's consistent with how you've handled the existing search keys, so fine for this PR, but the `i18n:check` script will flag the drift against the other locales when they're not backfilled.

---

The search UX itself is genuinely lovely and I don't want to lose it. The main thing I'm pushing back on is *where* the indexing lives — the rest is fixable mechanics. Let me know if you'd like to talk through the plugin split; I'm happy to help scaffold it.

alexander-pyrforos:
> The search UX itself is genuinely lovely and I don't want to lose it. The main thing I'm pushing back on is _where_ the indexing lives — the rest is fixable mechanics. Let me know if you'd like to talk through the plugin split; I'm happy to help scaffold it.

ok, I'll change the code to correspond described requirements and advices. please wait for new PR) (that's first time I've ever collaborated on public repo, tbh)

*most of given changes were made for my own use, since this project helped a little, but didn't exactly cover the whole probem

alexander-pyrforos:
try this version please

rmyndharis:
Hey @alexander-pyrforos — thanks for the rework, this is a really thorough response to the review. I've gone through the new commits and the plugin split is exactly the shape I was hoping for: the `message:persisted` hook + sandboxed indexer cleanly pulls indexing out of the core services, the REST client rewrite drops the SDK dependency entirely (better than I'd asked for), and the escape/highlight helper with its XSS-guard tests, the thumbnail throttling, and the i18n backfill all land right. Keeping the query API in core is the correct call. Genuinely nice work.

One sequencing heads-up before this merges. I've got an in-flight refactor on `main` that touches the same files — `message.service.ts` and `session.service.ts` in particular (centralising feature flags + some metrics). It doesn't change the persistence path or the hook mechanism your indexer hangs off, so your design is unaffected, but it moves enough lines that this branch won't auto-merge once it lands. Expect conflicts in those two files; they'll be mechanical (re-inserting your `message:persisted` emits into the new layout), not a redesign.

I'd like to land that refactor first, then rebase this branch onto it — that way we review your hook sites once, against the final structure, instead of twice. No action needed on your end right now. Happy to do the rebase myself so you're not chasing a moving target, but if you'd rather take it, I'll flag the exact conflict points when the refactor lands. Either way I'll ping you here.

Thanks again — for a first open-source contribution this is a genuinely strong PR.

rmyndharis:
Hey @alexander-pyrforos — circling back on this. The refactor I flagged yesterday has landed (v0.8.11 is out), so the rebase is unblocked. But before you do that work, I want to revisit the architecture with you, because it changes what the rebase even looks like — and I owe you a straight explanation rather than a quiet stall.

When I last wrote here I said *keeping the query API in core is the correct call*, and I want to be honest: I've reconsidered part of that. The *route* and the *contract* still belong in core — that stands. What I've changed my mind on is having the **Meilisearch client itself** in core. I sat with it and it doesn't fit the way this project tries to stay vendor-neutral, and I think there's a better shape that keeps everything good about your work without core picking a search winner.

The principle, with a precedent so it's not just me hand-waving: we made exactly this call recently with MCP. PR #256 proposed adopting `@silkweave/*` — a single-maintainer framework that brings its own HTTP runtime and logger — as a core dependency. We declined it and built MCP in-house on the official `@modelcontextprotocol/sdk` instead (`docs/24-mcp-integration.md`). Not because Silkweave was bad; because core shouldn't load-bear a specific third-party service or library, so the gateway stays usable by everyone regardless of their stack. Meilisearch-in-core would be the search version of that — it'd make one search vendor load-bearing for every deployment, including the ones who'd rather run Elasticsearch, Typesense, or nothing at all.

The shape I'd like to commit core to, and where your work fits:

1. **An open `SearchProvider` contract in core** (query + health) with a small provider registry — the same abstraction we already use for engines (`whatsapp-web.js` / `baileys`) and storage (`local` / `s3`).
2. **A built-in, DB-native full-text provider as the default** — Postgres `tsvector`/`GIN` and SQLite `FTS5` — so search works out of the box with zero external services, on the database the operator already runs.
3. **Meilisearch as a plugin** — one `SearchProvider` among many, living in the OpenWA-plugins repo. Its indexing stays exactly as you designed it (`message:persisted` hook + sandboxed indexer), and it registers as a provider on load.

What this means concretely for the PR, and why I don't think it's bad news for you:

- I'll land the core foundation (interface + registry + built-in FTS + the `/search` route + the `message:persisted` hook) first. Several pieces of *this* PR are genuinely reusable there and I'd carry them through with you credited — the `message:persisted` hook event + emit sites, the `GlobalSearch` panel structure, `search-highlight`, the i18n keys, and the JS SDK search resource (adapted to the generic shape).
- This PR then becomes the **Meilisearch provider plugin** in OpenWA-plugins — and you're the natural author of it. I'm glad to co-author and help scaffold it onto the new interface once the foundation lands.

I know this is a big ask, especially after I just told you the query-in-core shape was the right one — that whiplash is on me, not you, and I'm sorry for it. I wouldn't reopen it if I didn't think both the feature and your contribution deserve to land in a shape that fits where the project is going, and honestly the plugin outcome is a better home for the Meilisearch work than core would have been. Happy to share the provider-interface design in advance if you want to weigh in before the foundation lands — and no rush; I know "first PR" is already a lot.

alexander-pyrforos:
Hey @rmyndharis, thanks for walking through your reasoning and being upfront about the reversal — I appreciate that.
I'm on board with the direction overall, the vendor-neutral core argument makes sense, especially given the Silkweave precedent.
That said, I don't have time right now for another round of rewriting blind before there's a solid SearchProvider interface shape. So here's what I'd propose:

Could you share a draft of the interface and provider registry once it's ready? I'd like to look at it before I start adapting the code, so I don't end up redoing it twice.
I won't be touching the current branch for now — it's working and covers my own use case locally, so I'm not in much of a rush.
Once the foundation (core + route + hook) lands, I'm happy to take on the Meilisearch provider in OpenWA-plugins, if the offer to help scaffold it is still on the table.

One more thing — could you clarify which parts of this PR (if any) are likely to land on main, and on what rough timeline? I'll keep using my own fork for my day-to-day work in the meantime, since the current Meilisearch implementation already fits my needs as-is.
Let me know once the interface is ready — I'll plan around that.

rmyndharis:
Hey @alexander-pyrforos — you asked me to share the `SearchProvider` interface once it was ready and to call out which parts of this PR were landing, so I'm circling back to do exactly that. Short version: the foundation is on `main`, the interface is ready for you to build against, and the scaffold offer is absolutely still on the table.

**Where it landed.** The vendor-neutral shape we talked through shipped across a few PRs:
- **#664** (v0.8.12) — the open-core foundation: the `SearchProvider` interface + provider registry + the built-in database full-text provider (Postgres `tsvector`/SQLite FTS5) + `GET /api/search` + the `message:persisted` hook.
- **#668** (v0.8.13) — the dashboard search panel + the JS/Python/PHP/Java SDK search resources + i18n.
- **#674** — the piece that wasn't there when you last looked: the **host→plugin search RPC**. A sandboxed plugin now registers as a `SearchProvider` by calling `ctx.registerSearchProvider(handler)` from its worker, and the host routes `/search` queries to it over a correlated `search`/`search-result` wire protocol. That's what makes the plugin query path real — the core never talks to the search backend directly, exactly the separation you were arguing for.

**What carried over from this PR.** The parts I said I'd carry through did land, adapted to the generic shape: the `message:persisted` hook event + emit sites, the `GlobalSearch` panel structure, the `search-highlight` escape-then-highlight helper, the i18n keys, and the SDK search resource. They're referenced back to #650 in #664's description. That said — I owe you a clearer attribution than a PR cross-reference, and I'll make sure your handle is on the next CHANGELOG entry that touches the search-indexing origin. The ideas were yours and you should be credited for them.

**The interface you wanted to see.** It's in [`src/modules/search/search.types.ts`](src/modules/search/search.types.ts) — `SearchProvider` (id/label/search/health), `SearchQuery` (q + the filters + limit/offset), `SearchResults`/`SearchHit`. And I just opened **[#682](https://github.com/rmyndharis/OpenWA/pull/682)** — `docs/27-plugin-search-providers.md` — a full author guide: the `ctx.registerSearchProvider` handler signature, the `SearchHit` shape (incl. the `<mark>` snippet convention + epoch-seconds timestamp), indexing via `message:persisted` (including the backfill-is-your-responsibility note), the host-side guarantees (limit cap, session-scope re-filter, 10s timeout, 503 on failure, auto-supersede), and a complete minimal working example. That's the starting point you asked for — no more building blind.

**The Meilisearch plugin.** The offer to co-author / scaffold it in OpenWA-plugins stands exactly as before — you're the natural author of it, and your indexing + query logic from this branch carries over. Now that the RPC + the guide exist, the path is: a sandboxed plugin that registers on `message:persisted` to index and calls `ctx.registerSearchProvider` to answer queries, living in the plugins repo as one provider among many. Whenever you've got the bandwidth, ping me and I'll help land it.

I'm going to close this PR now — its core-side work has been absorbed into `main`, and the Meilisearch provider belongs in OpenWA-plugins as we discussed. That's not the end of this work, just the redirect we agreed on. Thanks again for pushing the design in the right direction — the `message:persisted` hook in particular ended up shaping how the whole search subsystem indexes.

PR #651, opened 2026-07-06
feat(database): add Postgres schema selection support

maplerichie:
## Description
Brief description of changes

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist
- [x] Tests added/updated
- [x] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

image


rmyndharis (approved):
Thanks for this one, @maplerichie — this is a really well-thought-out PR. Appreciate the care that went into it.

A few things that stood out (in a good way):

- **The `search_path` gotcha.** This is the subtle part a lot of people miss — TypeORM's `schema` option alone does *not* set the session `search_path`, so raw unqualified migration DDL (`CREATE TABLE "x"...`) silently lands in `public` while the migration ledger ends up in the configured schema. You caught this and paired the `schema` option with `extra.options: '-c search_path=<schema>,public'`, and documented exactly why. That note in `data-source.ts` is going to save someone a very confusing afternoon.
- **Refactoring for testability.** Pulling `buildPostgresDataSourceOptions(env)` out so the schema/search_path logic is unit-testable without mutating `process.env` or reloading the module — exactly right. The spec covers the three cases that matter (default → no `options` key, custom → `options` set, pool timeouts preserved).
- **Fail-fast validation.** Rejecting `pg_` prefix, invalid identifiers, and injection-ish values at boot — but only when `DATABASE_TYPE=postgres` (the sqlite-with-bogus-schema test is a nice touch). This is the right shape for a config that ends up in `search_path`.
- **The init script honesty.** Calling out that `docker-entrypoint-initdb.d` only runs on first init, and that an existing volume needs a manual `CREATE SCHEMA` — that limitation is real and you surfaced it in the `.env.example`, the docs, and the script header. Exactly how it should be done.
- **Belt-and-suspenders on the smoke test.** Parameterizing `table_schema = $1` in `pg-uuid-smoke.js` keeps it meaningful under a custom schema instead of silently asserting against `public`.

Merging now. Two tiny things for next time (not blocking): the cosmetic `logger.service.spec.ts` and `Infrastructure.css` `.form-group small{` hunks are unrelated to the schema work — easier to review if those go in their own commit/PR. No worries this round.

Thanks again for the thorough work and the i18n coverage!

maplerichie:
> Thanks for this one, @maplerichie — this is a really well-thought-out PR. Appreciate the care that went into it.
> 
> A few things that stood out (in a good way):
> 
> - **The `search_path` gotcha.** This is the subtle part a lot of people miss — TypeORM's `schema` option alone does *not* set the session `search_path`, so raw unqualified migration DDL (`CREATE TABLE "x"...`) silently lands in `public` while the migration ledger ends up in the configured schema. You caught this and paired the `schema` option with `extra.options: '-c search_path=<schema>,public'`, and documented exactly why. That note in `data-source.ts` is going to save someone a very confusing afternoon.
> - **Refactoring for testability.** Pulling `buildPostgresDataSourceOptions(env)` out so the schema/search_path logic is unit-testable without mutating `process.env` or reloading the module — exactly right. The spec covers the three cases that matter (default → no `options` key, custom → `options` set, pool timeouts preserved).
> - **Fail-fast validation.** Rejecting `pg_` prefix, invalid identifiers, and injection-ish values at boot — but only when `DATABASE_TYPE=postgres` (the sqlite-with-bogus-schema test is a nice touch). This is the right shape for a config that ends up in `search_path`.
> - **The init script honesty.** Calling out that `docker-entrypoint-initdb.d` only runs on first init, and that an existing volume needs a manual `CREATE SCHEMA` — that limitation is real and you surfaced it in the `.env.example`, the docs, and the script header. Exactly how it should be done.
> - **Belt-and-suspenders on the smoke test.** Parameterizing `table_schema = $1` in `pg-uuid-smoke.js` keeps it meaningful under a custom schema instead of silently asserting against `public`.
> 
> Merging now. Two tiny things for next time (not blocking): the cosmetic `logger.service.spec.ts` and `Infrastructure.css` `.form-group small{` hunks are unrelated to the schema work — easier to review if those go in their own commit/PR. No worries this round.
> 
> Thanks again for the thorough work and the i18n coverage!

Thank glm 5.2 for these 😎

PR #654, opened 2026-07-07
fix(dashboard): display selected state for webhook events

### LINKED ISSUE #653: [Bug]: Dashboard: Webhook event selection does not visually reflect selected state
### Pre-flight

### OpenWA version

0.8.10

### Deployment

Docker Compose

### Database

SQLite (default)

### WhatsApp engine

whatsapp-web.js

### What happened?

When creating a webhook from the dashboard, selecting webhook events does not provide any visual feedback. The selected events are saved correctly and the webhook functions as expected, but the UI does not indicate which events have been selected.

### Steps to reproduce

Open the Dashboard.
Navigate to Webhooks.
Click Create Webhook.
Enter a valid webhook URL.
Select one or more webhook events.

### Expected behavior

Selected webhook events should be visually highlighted so users can clearly see which events are selected.

### Relevant logs

```shell

```

### Actual Behavior

The webhook is created successfully with the selected events, but the event list does not visually reflect the current selection.

### Environment
OpenWA Version: 0.8.10
Browser: Chrome (latest)
Operating System: Windows 11
Installation: Local development

### Additional Information

This appears to be a dashboard UI issue only. The underlying webhook functionality works correctly; only the visual selected state is missing.

Deep-Bhanushali:
## Summary

This PR fixes the visual state of the webhook event selector in the dashboard.

## Problem

While creating a webhook, selecting events correctly updated the underlying data and the webhook was created with the selected events. However, the dashboard did not visually indicate which events were selected, making it appear as though the selection had not been registered.

## Solution

Updated the event selection component so that selected events are visually highlighted whenever events are selected or deselected.

## Testing

- Manually verified that selected events are visually highlighted.
- Manually verified that webhook creation continues to work as expected.
- No backend or API changes.

Fixes #653

### DIFF
--- a/dashboard/src/pages/Webhooks.css
+++ b/dashboard/src/pages/Webhooks.css
@@ -220,6 +220,9 @@
   border-radius: 6px;
   cursor: pointer;
   transition: all 0.2s;
+  display: inline-flex;
+  align-items: center;
+  gap: 0.375rem;
 }
 
 .webhooks-page .modal-body .event-tag:hover{
@@ -228,6 +231,32 @@
 }
 
 .webhooks-page .modal-body .event-tag.selected{
+  background: var(--primary);
+  color: white;
+  border-color: var(--primary);
+}
+
+.webhooks-page .modal-body .event-tag .tag-check-icon {
+  flex-shrink: 0;
+  display: inline-block;
+}
+
+/* Ensure modal event tags have proper unselected & selected styles in dark mode */
+.webhooks-page [data-theme='dark'] .modal-body .event-tag,
+:root:not([data-theme='light']) .webhooks-page .modal-body .event-tag {
+  background: var(--bg-light);
+  color: var(--text-secondary);
+  border-color: var(--border);
+}
+
+.webhooks-page [data-theme='dark'] .modal-body .event-tag:hover,
+:root:not([data-theme='light']) .webhooks-page .modal-body .event-tag:hover {
+  border-color: var(--primary);
+  color: var(--primary);
+}
+
+.webhooks-page [data-theme='dark'] .modal-body .event-tag.selected,
+:root:not([data-theme='light']) .webhooks-page .modal-body .event-tag.selected {
   background: var(--primary);
   color: white;
   border-color: var(--primary);
--- a/dashboard/src/pages/Webhooks.tsx
+++ b/dashboard/src/pages/Webhooks.tsx
@@ -337,16 +337,20 @@ export function Webhooks() {
               />
               <label>{t('webhooks.events')}</label>
               <div className="event-tags">
-                {availableEventNames.map(name => (
-                  <button
-                    key={name}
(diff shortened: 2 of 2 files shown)


rmyndharis (approved):
Thanks @Deep-Bhanushali — nice catch, and the fix lands in the right place. Approving.

I dug into the CSS to make sure this actually resolves #653 (not just paper-overs it), and it does — but there's some nuance worth surfacing, plus one small residual gap that's fine to leave for a follow-up.

**Root cause (more specific than the issue title implies):** this isn't really "selected state missing" — the `.event-tag.selected` rule has been present all along. The actual bug is a **CSS specificity collision with the dark-mode overrides** at the bottom of `Webhooks.css`:

```
.webhooks-page .modal-body .event-tag.selected { ... }      /* 0,4,0 — the selected style */
.webhooks-page [data-theme='dark'] .event-tag { ... }        /* 0,3,0 — dark chip override */
```

The dark override keys the *whole* `.event-tag` green (background `rgba(37,211,102,0.2)`), and because it stomps `background`/`color`/`border-color`, the `.selected` variant renders identically to the unselected one in dark mode — so there's no visible difference. Since the default theme is `system` (`useTheme.ts:30`, which *removes* `data-theme` and falls through to the `prefers-color-scheme: dark` block), this hits anyone whose OS is in dark mode — which lines up with the report (Windows 11 + Chrome).

**Why the fix works:** scoping the override to `.modal-body` and bumping its specificity to `0,5,0` lets `.selected` win again in dark mode. The `Check` glyph is a clean addition too — gives an unambiguous affordance beyond just the color change (good for color-blind users; the green-on-green was the weakest possible signal). And `Check` was already imported, so no new import churn.

**One residual gap (not blocking — fine for a follow-up):** the unselected chips *inside the modal* still pick up the green tint from the system-preference dark override:

```
@media (prefers-color-scheme: dark) {
  :root:not([data-theme='light']) .webhooks-page .event-tag { ... }   /* not scoped to modal-body */
}
```

So under "system" theme + OS-dark, the modal's unselected chips render greenish (matching the existing webhook chips on the page) rather than the neutral `--bg-light` / `--text-secondary` the base `.modal-body .event-tag` rule intends. The `selected` state is now correct, so functionally #653 is fixed — this is purely a visual-consistency nit. The `data-theme='dark'` (explicit) path *is* fully scoped after this PR; only the `prefers-color-scheme` (system) path is missing the equivalent `.modal-body` scoping. If you want to close it fully, mirroring the three new rules inside the `@media (prefers-color-scheme: dark)` block would do it. Happy either way — your call.

Minor nit: the two new `data-theme='dark'` blocks duplicate the three property declarations across the unselected/hover/selected variants. Not worth refactoring for a fix this small, just flagging it in case you'd prefer to DRY it with a shared selector list.

Merging once CI is green. Thanks again for the clean, well-scoped fix and the i18n-aware handling.

PR #656, opened 2026-07-07
fix(session): purge on-disk engine auth directory when deleting a session

m7fz7:
## Description
`DELETE /sessions/:id` tore down the running engine and removed the session's
database rows, but left the engine's **persistent on-disk auth directory** behind
(`data/baileys/<name>` for the Baileys engine, `data/sessions/session-<name>` for
whatsapp-web.js). These directories are keyed by session **name** and live
independently of any engine instance, so recreating a session under the same name
reloaded stale credentials from the leftover store.

This adds `EngineFactory.purgeSessionData(name)`, called from `SessionService.delete()`
after the DB removal succeeds, to remove that directory:

- Derived from config (not from a live engine), so it works even when the session was
  already stopped and has **no loaded engine** - the common delete case.
- Matches each adapter's exact path shape (`path.join(authDir, name)` for Baileys;
  resolved `session-${name}` under `sessionDataPath` for whatsapp-web.js).
- Guarded by `isSafeSessionName` as defense-in-depth against traversal into the
  `rm -rf` sink, and a no-op when the directory is absent.
- Best-effort: an unsafe name or an `rm` failure is logged, never thrown, so it can't
  turn an otherwise-successful delete into a 500. Runs only after the DB transaction
  commits, so a failed delete never purges auth for a session that still exists.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [x] Tests added/updated
- [x] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

N/A - backend/storage behavior, no UI surface.

## Related Issues
Closes #


rmyndharis (approved):
Thanks for this — well-scoped fix and clean execution. I verified the path shapes against both adapters and the keying, and they match exactly:

- **Baileys:** adapter `authPath = path.join(authDir, sessionId)` ↔ factory `path.join(authDir, name)` (authDir left unresolved, matching the adapter).
- **whatsapp-web.js:** adapter `clearLocalAuth` = `resolve(sessionDataPath)/session-${sessionId}` ↔ factory `resolve(sessionDataPath)/session-${name}`.

`create()` is called with `sessionId: session.name`, so `purgeSessionData(session.name)` hits the right key. The traversal guard (`isSafeSessionName`, the same one `create()` uses) keeps `../`/`.`/`\` away from the `rm -rf` sink, and the best-effort semantics (log + never throw, run only after the DB transaction commits) are correct.

I rebased this onto `main` — it had conflicted with a session-delete change that just landed — and added the CHANGELOG contributor credit. It's now mergeable and CI is green.

A few non-blocking notes for future polish:

1. **Engine-type switch.** `sessionAuthDir` uses the factory's *currently-configured* `engineType`, so if `ENGINE_TYPE` is changed and sessions created under the previous engine are later deleted, the purge targets the new engine's path and misses auth left under the old one. Bounded (rare, operator-initiated) and there's no per-session engine record to do better today, but a one-line note in the method comment would help a future reader.

2. **Path-shape duplication.** The factory re-derives the auth paths rather than calling the adapter (justified — on delete there's often no live engine). The match is therefore enforced only by comments, and the unit tests construct the expected path from the factory's own computation, so they won't catch a future divergence. A tiny shared `engineAuthDir(engineType, name, config)` helper used by both, or a test asserting the factory's path equals each adapter's constructed path, would lock it in.

3. **Device deregistration (follow-up).** This purges the local credentials but doesn't log the session out, so the linked-device entry stays registered on the WhatsApp account and accumulates over time. Out of scope here, but worth a separate issue for whether `delete()` should also call `logout()` when an engine is loaded.

Nice work on the test coverage — both engines, the missing-dir no-op, the unsafe-name refusal, and the no-loaded-engine integration case.

PR #660, opened 2026-07-08
feat(cache): add support for Redis username authentication

muhfalihr:
## Description
Added support for Redis authentication via username. This allows the application to connect to modern managed Redis services (such as those using Redis ACLs) that require both a username and password.

The username is retrieved from the `REDIS_USERNAME` environment variable, securely mapped through `ConfigService` into the global configuration, and passed to the `ioredis` instance during initialization.
I have also added this variable to `.env.example` and documented the update in `CHANGELOG.md`.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist
- [ ] Tests added/updated
- [x] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

Screenshot from 2026-07-08 12-18-39

## Related Issues
Closes #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Redis authentication via username.** Added support for `REDIS_USERNAME` to configure Redis connections that require a username.
+
 ## [0.8.11] - 2026-07-08
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- **Redis authentication via username.** Added support for `REDIS_USERNAME` to configure Redis connections that require a username.
+- **Redis authentication via username.** Added support for `REDIS_USERNAME` to configure the Redis cache connection and BullMQ queue connections that require a username.
 
 ## [0.8.11] - 2026-07-08
 
--- a/src/config/configuration.ts
+++ b/src/config/configuration.ts
@@ -11,6 +11,7 @@ export default () => ({
   redis: {
     host: process.env.REDIS_HOST || 'localhost',
     port: parseInt(process.env.REDIS_PORT || '6379', 10),
+    username: process.env.REDIS_USERNAME,
     password: process.env.REDIS_PASSWORD,
     connectTimeoutMs: parseInt(process.env.REDIS_CONNECT_TIMEOUT_MS || '5000', 10),
   },
--- a/.env.example
+++ b/.env.example
@@ -194,6 +194,7 @@ REDIS_HOST=localhost
 REDIS_PORT=6379
 # Fail Redis queue/cache connection attempts after this many ms.
 REDIS_CONNECT_TIMEOUT_MS=5000
+# REDIS_USERNAME=
 # REDIS_PASSWORD=
 
 # =============================================================================
--- a/docs/06-api-specification.md
+++ b/docs/06-api-specification.md
@@ -3793,6 +3793,7 @@ Merge-save infrastructure config to `data/.env.generated` (a `0600` secret file)
 | `database.sslRejectUnauthorized` | boolean | No | — | Only written when `sslEnabled` is true; default true |
 | `redis.enabled` / `.builtIn` | boolean | No | — | `builtIn`+enabled forces `redis` container + profile |
 | `redis.host` / `.port` | string | No | `port` is a string | Defaults `localhost`/`6379` |
+| `redis.username` | string | No | — | Redis ACL username |
 | `redis.password` | string | No | secret | Empty keeps existing |
 | `queue.enabled` | boolean | No | — | Writes `QUEUE_ENABLED` |
 | `storage.type` | `'local' \| 's3'` | No | — | `local` drops stale S3 keys; `s3` drops `STORAGE_LOCAL_PATH` |
--- a/docs/14-migration-guide.md
+++ b/docs/14-migration-guide.md
@@ -194,6 +194,7 @@ REDIS_ENABLED=true
 REDIS_BUILTIN=false      # false = external Redis
 REDIS_HOST=your-redis-host.com
 REDIS_PORT=6379
+REDIS_USERNAME=optional
 REDIS_PASSWORD=optional
 ```
 
--- a/pr_template_draft.txt
+++ b/pr_template_draft.txt
@@ -1,0 +1,22 @@
+## Description
+Added support for Redis authentication via username. This allows the application to connect to modern managed Redis services (such as those using Redis ACLs) that require both a username and password.
+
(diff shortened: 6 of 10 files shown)


src/config/configuration.ts
```diff
@@ -11,6 +11,7 @@ export default () => ({
   redis: {
     host: process.env.REDIS_HOST || 'localhost',
     port: parseInt(process.env.REDIS_PORT || '6379', 10),
+    username: process.env.REDIS_USERNAME,
```
rmyndharis, line 14: Heads up — this new `redis.username` is currently consumed only by the cache client (`cache.service.ts`). The two BullMQ sites need it too:

- `src/modules/queue/queue.module.ts` — the `forRootAsync` `connection` reads `redis.password` but not `redis.username`
- `src/modules/queue/redis-connection.ts` — `workerConnectionOptions()` reads `process.env.REDIS_PASSWORD` but not `REDIS_USERNAME`

Without those, an ACL Redis authenticates the cache but fails auth on the webhook/ingress queues — a confusing split where `/infra/status` shows Redis healthy. Small symmetric change mirroring `password`; exact lines are in the review summary.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Redis authentication via username.** Added support for `REDIS_USERNAME` to configure Redis connections that require a username.
```
rmyndharis: Two small things here:

1. "Redis connections" reads as app-wide, but only the cache connection is wired so far — worth narrowing to "the Redis cache connection" (or completing the queue-side wiring so it's accurate).
2. Per the repo's convention for external PRs, append ` Thanks @muhfalihr.` to the line.

Happy to handle both on merge if you'd prefer.

rmyndharis (changes requested):
Thanks @muhfalihr — nice, clean addition. The cache-client wiring is exactly right, and I appreciate that `.env.example` and a CHANGELOG entry came along with it. One thing I'd like to land before merging, plus a couple of small polish items.

## One change before merge: thread `REDIS_USERNAME` through the BullMQ connections too

The repo opens Redis connections in **three** places, and this PR covers one of them (the ioredis cache client). The other two — the BullMQ shared producer and the worker factory — still authenticate with password only:

- **`src/modules/queue/queue.module.ts`** — the `BullModule.forRootAsync(...).connection` (shared by *both* the WEBHOOK and INGRESS queues) reads `redis.host/port/password/connectTimeoutMs` but not `redis.username`. So the `redis.username` this PR adds is currently read by nobody on the queue side.
- **`src/modules/queue/redis-connection.ts`** — `workerConnectionOptions()` reads `process.env.REDIS_PASSWORD` directly but not `REDIS_USERNAME`, and the `WorkerConnectionOptions` interface has no `username` field.

Why it matters: the whole reason ACL username auth exists is managed/production Redis, and in OpenWA that's primarily the **BullMQ queue subsystem** (the cache is optional and off by default). On a Redis 6+ ACL instance with a non-default user, the cache authenticates and connects fine — `/infra/status` shows Redis green — while the producer and workers send legacy `AUTH <password>` and get `WRONGPASS`. That silent split-brain is precisely the case this feature is meant to enable, so I'd rather ship it complete.

The good news: it's a small, symmetric change, mirroring how `password` is already wired:

```ts
// queue.module.ts — inside the connection: { … } block
username: configService.get<string>('redis.username'),

// redis-connection.ts
export interface WorkerConnectionOptions {
  host: string;
  port: number;
  username?: string;        // new
  password?: string;
  connectTimeout: number;
}
// and in workerConnectionOptions():
  username: process.env.REDIS_USERNAME,   // new
  password: process.env.REDIS_PASSWORD,
```

(BullMQ accepts `username` — `RedisOptions = IORedis.RedisOptions & BaseOptions` — so this is a missing field, not an upstream limitation.) While you're in `redis-connection.ts`, the header docstring (~line 12) promises the worker creds are "host/port/password/timeout identical to the producer" — that list should include `username` once added.

Not a data-loss risk even as-is: when `queue.add()` fails fast (`enableOfflineQueue: false`), the dispatch path catches it and falls back to direct delivery. So this is about delivering the feature properly, not preventing an outage.

## Small polish (happy to do these on merge if you'd rather)

- **CHANGELOG wording** — the entry says "Redis connections" (plural), but right now only the cache connection is covered. Either narrow it to "the Redis cache connection", or — preferably — complete the fan-out above so the wording's accurate. Also, per the repo's convention for external PRs, please append ` Thanks @muhfalihr.` to the line.
- **Docs** — `docs/06-api-specification.md` (the `SaveConfigDto` config table, next to where `redis.password` is listed) and `docs/14-migration-guide.md` (the "external Redis" example) list `REDIS_PASSWORD` but not `REDIS_USERNAME`. One-line additions each.

## Further-out items (not blocking, just flagging for completeness)

The dashboard Infrastructure form and the `/infra/config` round-trip don't surface a Redis username yet (so it's env-only for now), and `docker-compose.yml` doesn't forward `REDIS_USERNAME` into the container (it doesn't forward `REDIS_PASSWORD` either — pre-existing). I'm happy to file these as follow-ups rather than load them onto this PR.

## What's done well

- `configService.get<string>('REDIS_USERNAME')` mirrors exactly how the adjacent password line is read — locally consistent. (The flat-vs-nested key difference versus `queue.module.ts` predates this PR.)
- Passing `undefined` when unset is byte-for-byte backward compatible with existing password-only deployments — no regression; existing setups are completely unaffected.
- CHANGELOG placement/format is correct (`[Unreleased] → Added`, bold lead phrase), and a non-breaking additive feature correctly maps to a patch with no tag needed pre-merge.

Happy to help land this — if it's easier, I can prepare the queue-side changes as a follow-up PR stacked on yours so you're not blocked. Just let me know.

muhfalihr:
Thanks for the thorough review and the detailed guidance!

I've gone ahead and made all the requested changes in my latest commit, so no need for a follow-up PR:

- **BullMQ Connections:** Threaded `REDIS_USERNAME` through both the shared producer (`queue.module.ts`) and the worker connection options (`redis-connection.ts`).
- **Tests:** Updated `redis-connection.spec.ts` to ensure the `username` field is correctly read and defaulted.
- **Docs:** Added `REDIS_USERNAME` to `docs/06-api-specification.md` and the external Redis example in `docs/14-migration-guide.md`.
- **Changelog:** Refined the wording in `CHANGELOG.md` to reflect that both cache and queue connections are now supported, and added the requested attribution.

Everything should be fully wired up now and ready to go. Let me know if anything else is needed before the merge!

rmyndharis (approved):
Thanks @muhfalihr — all resolved. `REDIS_USERNAME` now threads through all three Redis connection sites (cache client + BullMQ producer + workers), the worker docstring's producer-parity invariant is restored, the spec covers the new field, docs are updated, and the stray draft file is gone.

Verified locally (the fork has no CI, so I ran the gate): lint clean, `nest build` green, and the queue + cache Jest suite passes (22/22).

One tiny housekeeping note: the `Thanks @muhfalihr.` CHANGELOG attribution didn't survive the merge sync in your branch, so I'll append it as part of the merge. Merging now — really nice work, and thanks for adding the test beyond what I asked.

PR #662, opened 2026-07-08
Sync reply mode for plugin: supabase-otp-hook

maplerichie:
## Description
Brief description of changes

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [x] Tests added/updated
- [x] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

## Related Issues
Closes #


rmyndharis:
Hey @maplerichie — thanks so much for this one. The refactor pulling out `dispatchWebhookRaw` is genuinely nicely done, the test coverage is thorough, and the motivation really resonates with me: Supabase Auth's send-OTP hook can't reasonably live with a fast-ack `202`, since Supabase acts on the HTTP response. So some flavor of synchronous ingress is absolutely worth us getting right together.

I'd love to think through the shape with you before it lands in the core fabric, though — reading it alongside the companion plugin PR (rmyndharis/OpenWA-plugins#33) left me with a question I can't quite resolve on my own, and I may well be missing something.

The docs and CHANGELOG frame `sync-reply` as a way to report the *actual OTP-send outcome* (200 success, 502 downstream, 504 timeout) so the provider can act on it. But that plugin PR notes that the WhatsApp send is fire-and-forget *and* that Supabase's hook has a 5s timeout with no retry. If I'm reading that right, the response we send back can really only carry the pre-flight result (the session liveness probe) — by the time we reply `200`, the OTP hasn't necessarily gone out, and there's no provider retry for the response to steer. So the premise in the docs doesn't quite line up with how the one current consumer behaves, and I wanted to check that with you rather than assume.

That left me wondering what the mode adds over plain `async 202`. As far as I can tell, the honest delta is surfacing a `400`/`503` up front instead of into the DLQ — and the `401` path is already identical across both modes, since signature verification is host-side and runs before the split. That's a real, worthwhile benefit (failing fast on an auth-critical path matters!), and it feels like the kind of thing a host-side preflight could give us without a second execution mode: liveness + payload check returns the status, the actual send stays enqueued, and the durability model stays intact. Does that sound right to you, or is there a gap I'm not seeing?

A small category thing nagging at me too: a sync OTP hook reads more like an RPC ("send this OTP, tell me within 5s if you can't") than an event-delivery webhook, and routing it through dedup → queue → DLQ feels like it borrows machinery the use case doesn't really want — the dedup row lands before dispatch, which is harmless for Supabase (no retries) but would quietly short-circuit retries for any future sync consumer that does retry. Dormant today, but a bit of a trap later. Curious what you think.

And I want to be straight with you about the cost: §25 is built around ingress never running a plugin inline, and `sync-reply` relaxes that — HOL blocking on the HTTP worker, no DLQ, sync bursts sharing sandbox capacity with the async queue, and the worker setting response status + headers on the request thread. That's an acceptable trade *if* the mode earns its keep — which is what I'm hoping we can nail down together.

So rather than ask for specific changes, a few things I'd genuinely value your read on:

1. Given the plugin PR says no retry + fire-and-forget send, is there a benefit to `sync-reply` that a host-side preflight (status returned, send still enqueued) wouldn't cover? If there is, I'm really eager to hear it — I may well be missing context.
2. Does synchronous ingress belong *inside* the event-ingress path, or would it be happier as its own minimal RPC primitive — or even living in the plugin?
3. If supabase-otp-hook is the only consumer for now, would you be open to deferring the core change until a second consumer shows the abstraction's worth — and shipping the plugin's OTP delivery via the preflight path in the meantime?

And one more thing — if anyone else following along has a provider that needs a synchronous response (Slack slash commands, 3DS / payment callbacks, other auth hooks, anything where the HTTP reply carries a real outcome), I'd genuinely love to hear about it. That kind of signal is exactly what helps us decide whether sync ingress deserves a permanent home in the fabric and what shape it should take, and I'd be happy to work on the design together with anyone in that boat.

I'm honestly not trying to block the plugin — it should ship regardless, and I'm glad to help wire whichever path you prefer. I just want us to be thoughtful about adding a second execution mode to the fabric on a rationale that, as it stands, its own consumer doesn't seem to lean on. And please push back freely if I've misread how Supabase behaves here — you've clearly sat with this longer than I have.

(There are also a couple of small implementation details to tighten if we do go ahead — sanitizing/clamping the worker-supplied response headers and status — but those are quick and entirely secondary.)

maplerichie:
Thanks for the thorough review!
You're right on most of it, and I won't litigate the premise mismatch: the 200 is a pre-flight result, not a send outcome, and the README/CHANGELOG oversell it.

One thing I'd add on the delta: because this plugin declares signature.scheme: "none", verification happens in the worker, not host-side before the split. So in async mode the 401 (and the 503 on a dead session) get swallowed into a 202 and Supabase walks away thinking the OTP went out. Sync-reply is the only path where those reach the provider. The dead-session fail-fast is the UX I actually built it for.

Given the 5s cap and no retry, preflight + fire-and-forget is about the only sane shape, awaiting the send risks blowing the budget and leaving a send in flight after a 504.

The dedup trap and whether sync ingress earns a permanent home are fabric-level calls, and I don't want to push a second execution mode on the core on my plugin's back. The one wrinkle: the plugin pins to 0.8.11's sync-reply, so deferring would drop it back to async-202 and lose the fail-fast. I'd rather leave it as-is and let the shape evolve when new consumer shows up, but I'll adapt to whatever you settle on.

Appreciate!

PR #667, opened 2026-07-08
fix: session-init hang on OOM + pre-existing spec typecheck errors

INAPA-desarrolloTIC:
## Summary

- **Session init can hang forever with no error.** `engine.initialize()` in `session.service.ts` is awaited with no timeout, and neither whatsapp-web.js nor Puppeteer time-bound the initial launch/navigation (`page.goto` is called with `timeout: 0`, and the web-version-cache fetch has none either). If Chromium dies mid-navigation — in our case, OOM-killed under container memory pressure — the session sits wedged in `INITIALIZING` indefinitely, and `GET /sessions/:id/qr` 400s forever with nothing in the logs to explain why. Reproduced in production: three same-day hangs each coincided with a `dmesg`-confirmed OOM-kill inside the container's memory cgroup.
- Fixed by racing `engine.initialize()` against a 60s timeout. On timeout it force-kills the wedged engine, marks the session `DISCONNECTED`, and rethrows — which plugs into the existing retry path (`executeReconnect()`'s catch already reschedules with backoff), and surfaces the failure to a manual `POST /start` caller instead of hanging the request.
- Also set `protocolTimeout: 60000` on the Puppeteer launch options (default 180s let a dead browser's CDP calls, e.g. `getChats`/`getGroups`, hang 3 minutes before failing).
- Separately: fixed 12 pre-existing `tsc --noEmit` errors across 7 spec files (unrelated to the above, found while verifying the build). Each fix is explained in its commit.

## Test plan

- [x] `tsc --noEmit` clean
- [x] Full suite green: 111 suites / 1374 tests (`npx jest`, Node 22.19.0 — this project's `undici` requires ≥22.19.0)
- [x] Docker image rebuilt and redeployed in production; both live sessions auto-reconnected cleanly on restart

### DIFF
--- a/src/core/plugins/plugin-net.spec.ts
+++ b/src/core/plugins/plugin-net.spec.ts
@@ -6,7 +6,9 @@ function fakeSafeFetch(response: Response, sink: { init?: RequestInit }): typeof
   return (<T>(_url: string, init: RequestInit, use: (r: Response) => Promise<T> | T): Promise<T> => {
     sink.init = init;
     return Promise.resolve(use(response));
-  }) as typeof withSafeFetch;
+    // withSafeFetch's real signature types `use`'s Response as undici's (which adds fields like
+    // `textStream` the DOM lib type lacks) — bridge through unknown for this intentional test double.
+  }) as unknown as typeof withSafeFetch;
 }
 
 function cannedResponse(body: string, headers: Record<string, string>, status = 200): Response {
--- a/src/engine/adapters/whatsapp-web-js.adapter.ts
+++ b/src/engine/adapters/whatsapp-web-js.adapter.ts
@@ -304,6 +304,11 @@ export class WhatsAppWebJsAdapter extends EventEmitter implements IWhatsAppEngin
           // Only override the executable when explicitly configured; otherwise let
           // whatsapp-web.js fall back to Puppeteer's bundled Chromium.
           ...(this.config.puppeteer?.executablePath ? { executablePath: this.config.puppeteer.executablePath } : {}),
+          // Puppeteer's default (180000ms) lets a dead/OOM-killed Chromium's CDP calls (e.g.
+          // getChats/getGroups, observed as "Runtime.callFunctionOn timed out" in prod logs) hang
+          // 3 minutes before failing. Shortened to fail fast; session.service.ts's initializeEngine
+          // timeout (60s) is the primary guard against a wedged initial launch/navigation.
+          protocolTimeout: 60_000,
         },
         ...(authTimeoutMs !== undefined ? { authTimeoutMs } : {}),
         ...(versionPin ?? {}),
--- a/src/modules/audit/audit.controller.spec.ts
+++ b/src/modules/audit/audit.controller.spec.ts
@@ -11,7 +11,7 @@ describe('AuditController access control', () => {
     // Read the handler off the prototype as an opaque object so the lint unbound-method rule
     // (which guards against detached method `this`) doesn't fire on a metadata-only lookup.
     const proto = AuditController.prototype as unknown as Record<string, object>;
-    const role = new Reflector().get<ApiKeyRole | undefined>(REQUIRED_ROLE_KEY, proto.findAll);
+    const role = new Reflector().get<ApiKeyRole | undefined>(REQUIRED_ROLE_KEY, proto.findAll as unknown as Function);
     expect(role).toBe(ApiKeyRole.ADMIN);
   });
 });
--- a/src/modules/infra/infra.controller.spec.ts
+++ b/src/modules/infra/infra.controller.spec.ts
@@ -60,7 +60,7 @@ describe('InfraController access control (Vuln 2)', () => {
 
   it.each(adminOnly)('%s requires the ADMIN role', method => {
     const handler = InfraController.prototype[method as keyof InfraController] as object;
-    const role = reflector.get<ApiKeyRole | undefined>(REQUIRED_ROLE_KEY, handler);
+    const role = reflector.get<ApiKeyRole | undefined>(REQUIRED_ROLE_KEY, handler as unknown as Function);
     expect(role).toBe(ApiKeyRole.ADMIN);
   });
 });
--- a/src/modules/session/session.service.spec.ts
+++ b/src/modules/session/session.service.spec.ts
@@ -1214,7 +1214,7 @@ describe('SessionService', () => {
           driverError: { code: 'SQLITE_CONSTRAINT_UNIQUE', message: 'UNIQUE constraint failed' },
         }); // re-fire
 
-      const msg = {
+      const msg: IncomingMessage = {
         id: 'wa-1',
         from: 'peer@c.us',
         to: 'me@c.us',
--- a/src/modules/session/session.service.ts
+++ b/src/modules/session/session.service.ts
@@ -529,7 +529,16 @@ export class SessionService implements OnModuleDestroy, OnModuleInit, OnApplicat
     // initializes, so writing INITIALIZING afterwards would clobber that progress.
     await this.updateStatus(id, SessionStatus.INITIALIZING);
 
-    await engine.initialize({
+    // engine.initialize() launches Chromium and navigates to WhatsApp Web with no internal
+    // timeout of its own (whatsapp-web.js disables Puppeteer's navigation timeout and its
+    // web-version-cache fetch has none either). If Chromium is killed mid-navigation (e.g. an
+    // OOM-kill under memory pressure — observed 2026-07-08, 3x same day) this await never
+    // resolves or rejects, wedging the session in INITIALIZING forever with no error logged.
+    // Race it against a timeout so a wedged init fails fast instead: executeReconnect()'s
+    // existing catch already retries on any thrown error here, and start()'s caller sees the
+    // rejection surfaced as an API error rather than a silently hung request.
+    const engineInitTimeoutMs = 60_000;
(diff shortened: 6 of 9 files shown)


rmyndharis:
Hi @INAPA-desarrolloTIC — thanks for this one! The init-hang fix addresses a genuinely painful failure mode (a session wedged in `INITIALIZING` with nothing in the logs is exactly the kind of thing that's hard to diagnose in prod), and the race itself is written correctly — the `initPromise.catch(() => undefined)` guard and the timer cleanup in `finally` are spot on. The pre-existing spec typecheck fixes are appreciated too; I verified them and they all match the real signatures. Nice work.

A few things to work through before merging — none of them are objections to the approach, they're mostly about fitting alongside changes that landed on `main` since you branched:

**1. Please rebase first — this branch is behind `main`.** The PR is based on `v0.7.4`, and `main` has moved ~191 commits since (now v0.8.11+), including a couple that touched session-init failure handling (#600, #631). It merges cleanly textually, but `start()` on `main` now has its own init-failure `catch` that marks the session `FAILED` *with the failure reason surfaced*. Your new `catch` inside `initializeEngine` runs first, deletes the engine, and rethrows — so by the time `start()`'s catch sees it, `engines.get(id)` is already empty and the `FAILED`+reason write is skipped. The net effect on merge is that a genuine init failure (e.g. Chromium can't launch — not the 60s-timeout case you're targeting) ends up `DISCONNECTED` with the reason hidden, which undoes the diagnosability #631 added.

   The cleanest reconcile: have `initializeEngine`'s catch handle the **timeout** case (→ `DISCONNECTED`, retryable) and let real rejections propagate to `start()`'s existing catch (→ `FAILED` + reason). That keeps both behaviors intact.

**2. `protocolTimeout: 60_000`.** Quick heads-up that this isn't only an init setting — it caps *every* CDP call for the browser's whole lifetime. Once a session is `READY`, `GET /sessions/:id/chats` / `/groups` calls `client.getChats()`, which serializes all chats in a single CDP round-trip; on a busy/large account on a loaded container that used to land in the ~60–90s window, this would now time out where the 180s default was fine. And since the `Promise.race` already bounds the init launch, shortening `protocolTimeout` doesn't actually help the hang. Two options: revert it to the default (and rely on the race for init), or make it configurable (`WWEBJS_PROTOCOL_TIMEOUT_MS`) so heavier deployments can tune it.

**3. Tests for the new path.** The new catch block (force-kill → engine evict → `DISCONNECTED` → rethrow) is the heart of this fix and currently has no test coverage — and I noticed no CI ran on the branch. Would you mind adding a couple, reusing the existing `jest.useFakeTimers()` pattern? The two that would lock it in:
   - `initialize()` never resolves → advance timers past 60s → assert `forceDestroy` called, engine evicted, status `DISCONNECTED`, and `start()` rejects with "timed out after 60000ms".
   - `initialize()` rejects with a real error → assert the same cleanup runs *and* the original error is rethrown (so reconnect retries on it).

**Minor / nice-to-haves:**
- The 60s init timeout is hardcoded, which silently overrides `WWEBJS_AUTH_TIMEOUT_MS` for anyone who legitimately raised it for slow first boots (WSL2 / low-resource containers). An `ENGINE_INIT_TIMEOUT_MS` env (default 60000), clamped `>= authTimeoutMs`, would keep that knob effective.
- Side note while we're here: CI never type-checks spec files (no `tsc --noEmit` step; `nest build` excludes specs), which is how those 12 errors went unnoticed. Worth a follow-up, but it doesn't need to be in this PR.

The core fix is solid and we definitely want it — I just want to make sure it lands cleanly against current `main` and preserves the failure diagnostics that #631 put in place. Happy to help with the rebase or the test cases if useful. Thanks again for digging into this one.

rmyndharis:
Hey @INAPA-desarrolloTIC — circling back to this one. I dug deeper into the actual init mechanism against current `main`, and wanted to follow up with what I found (and a correction to my earlier comment). The short version: the core fix is even more worth taking than I first said, there's one real bug in the catch block I'd like to correct, and two of the three parts don't apply against `main` anymore.

**1. The init hang is real, and I confirmed it from the library source — not just from the repro.** `whatsapp-web.js`'s `Client.initialize()` navigates with `page.goto(WhatsWebURL, { waitUntil: 'load', timeout: 0 })`; `timeout: 0` is Puppeteer's "never time out" sentinel. `authTimeoutMs` doesn't help here — it only bounds the *post*-navigation `inject()` auth poll, so it can't rescue a hung `goto`. And nothing on the OpenWA side bounds `await engine.initialize()` (session.service.ts:656). The #635 reaper is purely reactive (it fires on a terminal `onError` or a rejected re-init), so a session that simply never settles sits in `INITIALIZING` forever. Your `Promise.race` at the service layer is exactly the right shape for this — and better than I first gave it credit for, because `initializeEngine()` is the one chokepoint both engines (wwjs **and** the new Baileys v7 adapter) and both callers (`start()` + `executeReconnect()`) flow through. So one ~12-line race bounds all of them.

**2. The one bug to fix: the catch block downgrades *real* init failures.** This is the thing I'd most like to correct. Your `catch (err)` around the `Promise.race` handles both the timeout **and** a genuine rejection (e.g. Chromium can't launch). For both it does `engines.delete(id)` + `updateStatus(DISCONNECTED)` + rethrow. The consequence on current `main`: when that rethrow reaches `start()`'s catch, `engines.get(id)` is already `undefined`, so its `if (orphan)` guard skips the `updateStatus(FAILED)` write — and since `lastError` is only surfaced for `FAILED` sessions, the failure reason gets hidden from the API. That quietly undoes the diagnosability #631 added. (Second-order: on a backend restart, a `DISCONNECTED`-but-previously-paired session gets auto-started again, so a genuinely broken session — say, a missing Chromium binary — would loop on every boot instead of staying `FAILED`.)

The clean fix is to scope the catch to the timeout only, and let real rejections propagate untouched so `start()`'s existing catch keeps owning `FAILED`+reason. A typed error makes the split explicit:

```ts
// Only the timeout mutates state here. A REAL rejection must propagate untouched so the
// caller's catch (start() → FAILED+reason, executeReconnect() → retry) keeps the behavior
// #600/#631 established.
const engineInitTimeoutMs = 60_000;
initPromise.catch(() => undefined); // Promise.race can't cancel the losing promise

let initTimer: ReturnType<typeof setTimeout> | undefined;
try {
  await Promise.race([
    initPromise,
    new Promise<never>((_, reject) => {
      initTimer = setTimeout(() => reject(new EngineInitTimeoutError(engineInitTimeoutMs)), engineInitTimeoutMs);
    }),
  ]);
} catch (err) {
  if (err instanceof EngineInitTimeoutError) {
    this.logger.error(`Engine initialization timed out for session ${session.name}`, undefined, { sessionId: id, action: 'engine_init_timeout' });
    this.sessionErrors.set(id, err.message);
    await this.teardownEngineSafely(id, engine, e => e.forceDestroy(), 'force-destroy');
    this.engines.delete(id);
    await this.updateStatus(id, SessionStatus.DISCONNECTED); // retryable; executeReconnect backoff picks it up
  }
  throw err;
} finally {
  if (initTimer) clearTimeout(initTimer);
}
```

I verified this locally against `main`: a real `initialize()` rejection still ends up `FAILED` with the reason recorded (so `lastError` surfaces), and a wedged `initialize()` (never resolves) still gets force-destroyed, evicted, marked `DISCONNECTED`, and rethrown into the retry path after 60s. The `DISCONNECTED` write stays — it's the correct transient state for the reconnect path (which doesn't set a status itself between attempts).

**3. `protocolTimeout: 60_000` — I'd revert this one.** I was right that it's not init-scoped (it caps *every* CDP call for the browser's whole lifetime), and digging in I found it actually doesn't help the hang at all: `page.goto`'s load-event wait is event-driven, not a CDP command response, so `protocolTimeout` never bounds it — only the `Promise.race` does. Meanwhile `getChats()`/`getGroups()` are each a single `page.evaluate` that fires a WhatsApp-server round-trip per group and serializes everything into one CDP response, so on a busy account they legitimately land in the 60–180s band the 180s default exists to absorb. Cutting it to 60s would time those out for zero init benefit. (Also, the comment's premise — that a dead/OOM-killed browser hangs CDP calls for 3 minutes — isn't quite right; a killed browser rejects pending calls immediately on transport close. The actual forever-hang is the pre-kill memory-pressure stall, which the race covers.) Recommend dropping it and relying on the race; if you want fail-fast on heavy reads later, a per-call race around those specific methods is the safer lever than a global CDP-timeout cut.

**4. Correction to my earlier comment: the spec fixes already shipped.** You were right that CI didn't type-check specs at the `v0.7.4` base — but that's no longer true. #680 (`8f1343b`, 2026-07-10) both added the `tsc --noEmit` CI gate *and* fixed all 12 of those errors (four identically to this PR; the other three via a slightly cleaner function-typed-`Record` pattern). So against current `main`, Part 3 is a no-op on four files and a conflict on three. On a rebase, just drop Part 3 entirely — it's why the branch is currently conflicting.

Given it's been a few days and the branch now conflicts with `main` around exactly those stale spec hunks, I'm happy to carry this forward so the init-hang fix isn't blocked — I'd open a fresh PR with just the race (with the catch correction above) plus tests, and credit the diagnosis and the `Promise.race` approach as yours in the CHANGELOG. Totally fine if you'd rather drive the update yourself — just say the word and I'll hand it back. Either way, thanks for digging into this one; the `timeout: 0` landmine is a real latent failure mode for anyone running the wwjs adapter under memory pressure, and your approach is the right fix for it.

PR #686, opened 2026-07-10
Baileys v7 upgrade

m7fz7:
## Description

Upgrades OpenWA's Baileys engine from `6.7.23` to `7.0.0-rc13` and fixes a cluster of related reliability bugs in the Baileys session/message-retry path, uncovered while chasing reports of messages stuck on "Waiting for this message. This may take a while." (reported specifically on iOS recipients) and a first-message-after-reconnect drop.

- Added a `getMessage` implementation to the Baileys socket config — it defaulted to a no-op, so WhatsApp's own retry-on-decrypt-failure protocol had nothing to resend, leaving some recipients stuck indefinitely.
- Fixed `sessionId: session.name` → `id` in the engine factory call, which broke the message store's FK check and the auth directory keying on every session start/reconnect.
- Wrapped the signal key store in Baileys' own `makeCacheableSignalKeyStore`, closing a write-then-immediate-read race that could make a freshly-established session appear "missing" and force an unnecessary fresh PreKey handshake.
- Upgraded `@whiskeysockets/baileys` to `7.0.0-rc13` to pick up an upstream concurrency rewrite (WhiskeySockets/Baileys#2571, #2587) that closes a "dual uncoordinated prekey queues" race in the same session-establishment path — no `6.8.x` backport exists. Updated the LID-related code this breaking change touches (`remoteJidAlt`/`participantAlt` replacing the removed `senderLid`/`senderPn`/`participantLid`/`participantPn` fields, `Contact.phoneNumber` replacing the removed `jid`, `chats.phoneNumberShare` → `lid-mapping.update`). No ESM migration was needed — the existing lazy `await import()` boundary already isolated Baileys from the rest of this CJS codebase.
- Fixed `handleMessagesUpsert` treating every `type: 'append'` upsert as history sync — Baileys can tag a genuinely new customer message `'append'` when it arrives in the same window as a reconnect's state-sync handshake, silently dropping it. Now gated on the message's own timestamp against the connection's open time instead, with the account's own sent-message echoes excluded from that override so they don't double-fire the existing `emitOwnSendEcho` webhook.

**Confidence note**: this is the strongest available lead for the "Waiting for this message" symptom (a real, well-documented upstream concurrency race fixed only in this version line), not a confirmed-by-live-test fix. Manual verification against a live gateway is still pending.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

*(Not marking "Breaking change" - this is an internal dependency/engine upgrade behind the existing `IWhatsAppEngine` abstraction; no REST API surface changes.)*

## Checklist
- [x] Tests added/updated
- [x] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

image

## Related Issues
Closes #


rmyndharis (changes requested):
# Thanks for this — really thorough work on a genuinely hard problem

Hey @m7fz7, thanks for this. Chasing "Waiting for this message" is one of the more painful WhatsApp multi-device issues to debug, and I appreciate both the depth you went to and the honesty in your confidence note — that's exactly the right framing for a change of this shape.

I dug into both the upstream and the OpenWA side before commenting, and I want to start by saying: **the Baileys 6.7.23 → 7.0.0-rc13 upgrade is justified, and I'm not going to push back on the version bump itself.** The symptom is a documented Baileys-internal race (WhiskeySockets/Baileys#2115, resolved by the rc.9+ "stability fixes for resending messages"), and it's a *different* failure mode from the missing-`getMessage` one. Both produce the same recipient-facing symptom, and a user hitting the concurrency race wouldn't be helped by the OpenWA-side fixes alone — so the upgrade belongs in this PR rather than split off. The LID API adaptations (`remoteJidAlt`/`participantAlt`, `Contact.phoneNumber`, `lid-mapping.update`) all match the official v7 migration guide one-for-one, so that work is solid.

That said, there's one thing I think we have to fix before merge, and a couple we should verify. The headline issue is subtle and I want to spell out the why.

## 🔴 Blocker: `session.name → id` relocates the on-disk auth directory

This is the part I want to be most careful about, because the *intent* is right but it has a side effect the PR doesn't account for.

`config.sessionId` is overloaded today — it's used for two unrelated things:

- **As the DB session key** for the message store: `messageStore.put(this.config.sessionId, …)` / `getMessage(this.config.sessionId, …)` (`src/engine/adapters/baileys.adapter.ts:481`, `:1011`, `:1429`, `:1468`). `BaileysStoredMessage.sessionId` is an FK to `sessions.id` (the UUID), so with `session.name` here every `put()` hits a foreign-key violation and gets **silently dropped** by the orphan guard in `src/engine/adapters/baileys-message-store.service.ts:71-83` — the store stays empty, `getMessage` returns null, the retry protocol stalls. Your fix (use `id`) is exactly right for this side. ✓
- **As the on-disk auth-directory key**: `this.authPath = path.join(config.authDir, config.sessionId)` (`src/engine/adapters/baileys.adapter.ts:142`).

Changing `sessionId` from `session.name` to `id` fixes the first use but silently relocates the second. The catch is that `EngineFactory.purgeSessionData(session.name)` (`src/engine/engine.factory.ts:135`, called from `src/modules/session/session.service.ts:435`) still keys the auth dir by **name**, and its docstring explicitly says it must match what `create()` uses. After this PR, the engine reads/writes auth at `authDir/<uuid>` while `purgeSessionData` still removes `authDir/<session.name>`. Two consequences:

1. **Every existing session re-links on upgrade** — its auth state lives at `authDir/<name>`, the engine now looks in `authDir/<uuid>`, finds nothing, and asks for a QR. For a multi-session gateway that's a mass re-link.
2. **On delete, the auth dir leaks** — purge removes the wrong (now-empty or stale) directory, so the live `<uuid>` credentials accumulate on disk.

This is also why I'd gently push on the "non-breaking" box in the description — the REST surface is unchanged, but the on-disk keying change is a breaking operational change unless we handle it explicitly.

**Suggested fix:** decouple the two concerns rather than flipping the shared field. Keep `config.sessionId = session.name` for the auth path (preserves the existing dir, no re-link, stays consistent with `purgeSessionData`), and pass the UUID separately for the DB-keyed consumers — e.g. `config.dbSessionId = id` used by `messageStore.put/getMessage/clearSession`, and the `BaileysSessionStore` (worth checking what `LidMappingStore.remember` expects too). If you'd rather move auth dirs to UUIDs as a clean break, that's defensible too, but it needs a coordinated `purgeSessionData`/`getAuthDirectory` update *and* a migration that relocates `authDir/<name>` → `authDir/<uuid>` for existing sessions. Happy to pair on either path.

## 🟠 Verify: `whatsapp-rust-bridge` native module in Docker

v7 pulls in `whatsapp-rust-bridge` (signal crypto moved to Rust). That's a native dep, and the test suite mocks `@whiskeysockets/baileys` entirely (`src/engine/adapters/baileys.adapter.spec.ts:57`), so CI won't tell us whether it actually loads at runtime. Could you confirm the Docker build succeeds on **both** linux/amd64 and linux/arm64 (prebuilt binaries), or that the image has a Rust toolchain if not? This bit us recently with another native binary on arm64, so I'd like to be sure before we merge.

## 🟡 Confirm: auth state supports the new LID keys

The v7 migration guide calls out that the auth state must support the `lid-mapping`, `device-list`, and `tctoken` keys ("Make sure you have updated your authentication state"). We're on the stock `useMultiFileAuthState` plus your new `makeCacheableSignalKeyStore` wrap. Is the built-in multi-file state in rc13 persisting those? My read is it should (it's Baileys' own), but if it isn't, LID↔PN resolution won't survive a restart. A quick line confirming you checked would be great.

## 🟢 Minor: 6.x branch check

You dismissed a 6.x target on "no 6.8.x backport" — `6.17.16` is the published 6.x tip and looks like it wasn't evaluated. I'm fairly sure it lacks the resend-stability fix (the release notes tie that to rc9+), so 7.0 is almost certainly right — but could you confirm you looked at 6.17.16?

## Nits

- **CHANGELOG**: the new entry crams all five changes into one bolded paragraph. Mind splitting into one bullet per change? Easier to skim, and it matches the rest of the file.
- **`#703`** is referenced in a test name but the PR body leaves `Closes #` empty — does that issue exist?
- `type BaileysContactWithPhone = BaileysContact` is now a no-op alias (`src/engine/adapters/baileys-session-store.ts`) — fine to drop and use `BaileysContact` directly.

## What I really liked

- The `getMessage` implementation is the documented Baileys fix, and the `!key.id` guard is correct.
- `makeCacheableSignalKeyStore` closes a real write-then-read race — good catch.
- The `fromMe` exclusion in the append gate is **safe**: `emitOwnSendEcho` already fires `onMessageCreate` for own sends on both send paths (`src/engine/adapters/baileys.adapter.ts:487`, `:1438`), so excluding them here prevents a double-fire rather than dropping a message.
- The `lidPnPair` helper is clean, and the `connectedAt` clock-skew comment shows the right kind of thinking.
- The three new append tests are solid adapter-level coverage.
- Your honesty in the confidence note — that framing genuinely speeds up review.

## Path forward

The auth-dir decoupling is the one real merge blocker; the Docker + auth-state items are "confirm and we're good." If it'd help, I can prototype the `dbSessionId` decoupling so you can see what I have in mind — or take it whichever direction you prefer. Thanks again, this is a meaningful piece of work.

rmyndharis (approved):
# Re-review: blocker resolved — approving

Hey @m7fz7, thanks for the quick turnaround. The decouple is exactly right and lands cleanly. I traced every consumer before approving:

- `engineFactory.create({ sessionId: session.name, dbSessionId: id })` — auth-dir key stays the session name, DB-row key is the UUID. ✓
- `authPath = path.join(authDir, config.sessionId)` is unchanged, so existing auth dirs stay put — no re-link on upgrade. ✓
- All three `messageStore.put` sites, both `getMessage` sites (the new retry impl + `requireStored`), and `clearSession` now use `config.dbSessionId`, so the FK to `sessions.id` is satisfied and the store actually populates. ✓
- `purgeSessionData(session.name)` still matches the auth-dir key — no credential leak on delete. ✓
- The `isSafeSessionName` guard correctly stays on the auth-dir key (`sessionId`), the one that reaches the filesystem. ✓

The CHANGELOG bullet captures the why accurately, and good call evaluating `6.17.16` rather than just citing "no 6.8.x backport."

I also followed up on the `BaileysSessionStore` flag from my first review: `lid-mapping.sessionId` is a plain `varchar` provenance column, not an FK, so keeping it on the name is safe — no silent-drop path there. Mentioning it so you know it wasn't an oversight.

## One thing not to spend time on: the Docker ❌

The "Docker Build: fail" check is **not a build failure** — the image compiles cleanly (native deps install via prebuilt binaries, no Rust toolchain needed). It fails only at the final push:

```
denied: installation not allowed to Write organization package
```

That's just the fork-PR token lacking write access to the GHCR org registry — nothing in this PR. It'll pass on merge to `main`, where the workflow runs with the repo's own credentials.

## Minor (non-blocking, your call)

- The `#703` reference in the test name is still dangling (`Closes #` is empty in the PR body) — fine to leave or rename, no strong preference.
- The arm64 image and the v7 auth-state keys (`lid-mapping` / `device-list` / `tctoken`) are runtime confirmations CI can't give us; your confidence note already frames those honestly. Worth a live-connect smoke before we cut a release tag, but not a merge blocker here.

Approving — thanks for closing this out cleanly.

PR #717, opened 2026-07-12
fix(engine): address 1:1 Baileys sends to the contact's LID when mapped

isaacmendes:
## Problem

On the Baileys engine, every 1:1 send addressed by phone number (`<phone>@c.us` / `<phone>@s.whatsapp.net`) to a LID-migrated contact is rejected by WhatsApp with **ack error 463** (`NackCallerReachoutTimelocked` / "account restricted or missing tctoken"), even when:

- the account is healthy (the same message sent from the paired phone delivers), and
- a valid, fresh `tctoken` for the contact is present in the auth store (Baileys 7.0.0-rc13 stores and attaches it).

The send API still returns a `messageId` (Baileys generates it locally), so the failure is silent from the caller's perspective — the message is simply never delivered. This matches the open upstream investigation in WhiskeySockets/Baileys#2698.

Verified live against a production number: the identical message sent to `<phone>@s.whatsapp.net` gets ack error 463; sent to the contact's `<lid>@lid` it is delivered and read.

## Fix

Resolve phone-dialect 1:1 chat ids to the contact's LID at the adapter's send boundary, using the same mapping the Baileys send path itself consults (`sock.signalRepository.lidMapping.getLIDForPN`):

- new `toDeliverableJid()` helper in `baileys.adapter.ts`
- applied in `sendTextMessage`, `sendContent` (all media/location/contact/poll sends route through it) and `sendChatState` (presence to a migrated contact previously failed with "No LID for user")

Groups, broadcast, already-`@lid` and unmapped ids pass through unchanged, so behavior for non-migrated contacts is identical to before. Resolution is best-effort: any lookup error falls back to the phone jid.

## Tests

- 2 new specs: phone-dialect id resolves to the LID when mapped; passes through when unmapped
- full `baileys.adapter.spec.ts` suite green (120/120), `tsc --noEmit` clean
- live E2E on a real session: inbound message → API reply → delivery + read receipts, no 463

### DIFF
--- a/src/engine/adapters/baileys.adapter.spec.ts
+++ b/src/engine/adapters/baileys.adapter.spec.ts
@@ -43,6 +43,7 @@ class FakeSock extends EventEmitter {
   public newsletterMetadata = jest.fn();
   public newsletterFollow = jest.fn().mockResolvedValue(undefined);
   public newsletterUnfollow = jest.fn().mockResolvedValue(undefined);
+  public signalRepository: { lidMapping: { getLIDForPN: jest.Mock } } | undefined;
   fire(event: string, arg: unknown): void {
     this.emitter.emit(event, arg);
   }
@@ -578,6 +579,7 @@ describe('BaileysAdapter location + contact + poll sends', () => {
 describe('BaileysAdapter messaging', () => {
   beforeEach(() => {
     fakeSock.user = { id: '628999:1@s.whatsapp.net', name: 'Me' };
+    fakeSock.signalRepository = undefined;
     fakeSock.resetEmitter();
     jest.clearAllMocks();
   });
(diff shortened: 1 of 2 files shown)


rmyndharis:
Rebased onto `main` and verified locally — backend gate green (lint 0 errors, build clean, 2404 tests pass). The rebase resolved one spec conflict (the new `signalRepository` fixture landed at the same spot as the #719 newsletter FakeSock fields; both kept).

Clean fix for a real silent-delivery bug. Resolving phone-dialect 1:1 ids through `getLIDForPN` at the send boundary is the right seam, and the best-effort fallback to the PN keeps non-migrated contacts unchanged. I confirmed the disappearing-timer path stays correct under the LID: `getEphemeralExpiration` already multi-keys on the raw/engine/neutral jid, so `withEphemeral` still resolves the timer after the switch.

Two non-blocking follow-ups worth a look later:
- Whether ack 463 also affects react/delete/chatModify to migrated contacts. Those route through the stored `target.key`, which should already carry the LID for a migrated sender, but it isn't live-verified here.
- A guard test asserting a send to a phone-dialect id with a known mapping still attaches `ephemeralExpiration` — locks the multi-key lookup against future regression.

Squash-merging.

PR #720, opened 2026-07-12
feat(sdk): add Go SDK

Revelts:
## Summary

Adds an official, hand-written **Go SDK** under `sdk/go/`, joining the existing JavaScript/Python/PHP/Java clients. It covers the full user-facing API surface and is built to the same "hand-written + mock-transport tested, wire types in a codegen-isolatable module" convention as the other SDKs.

Import path: `github.com/rmyndharis/OpenWA/sdk/go` (package `openwa`). Stdlib-only — **no third-party dependencies**. Go 1.22+.

```go
client, err := openwa.New("http://localhost:2785", "owa_k1_…")
ctx := context.Background()
client.Sessions.Start(ctx, "my-session")
res, err := client.Messages.SendText(ctx, "my-session", openwa.SendTextRequest{
    ChatID: "628123456789@c.us",
    Text:   "Hello from the OpenWA Go SDK!",
})
```

## Design

- **Client entry point** — `openwa.New(baseURL, apiKey, opts...)`; required creds positional, everything else via functional options. Safe for concurrent use.
- **Services by domain** — `client.Sessions`, `.Messages`, `.Contacts`, `.Groups`, `.Webhooks`, `.Chats`, `.Status`, `.Labels`, `.Channels`, `.Catalog`, `.Templates`, `.Health`, `.Search`, `.Auth`.
- **Context-first** — every network method takes `ctx` as its first arg; the context bounds the request and any retries.
- **Functional options + DI** — `WithHTTPClient`, `WithTransport`, `WithLogger`, `WithRetry`, `WithMiddleware`, `WithTimeout`, `WithUserAgent`, `WithHeader`, `WithInsecureHTTP`.
- **Transport middleware pipeline** — composable `Middleware` (auth, logging, retry built in; bring your own for tracing/metrics). Retry wraps auth so every attempt is re-authenticated; bodies rewind via `req.GetBody`.
- **Typed errors** — `*APIError` (StatusCode/Message/Kind/Body) plus sentinels matchable with `errors.Is`: `ErrBadRequest/ErrUnauthorized/ErrForbidden/ErrNotFound/ErrConflict/ErrRateLimited/ErrNotImplemented`; `TimeoutError` for deadlines.
- **Wire types** split into per-domain `types_*.go` files so an OpenAPI codegen pass can regenerate them without touching the hand-written service methods.

## Safety parity with the other SDKs

- API key sent as `X-API-Key`; insecure plaintext `http://` to a non-localhost host logs a warning.
- **Redirects never followed** — a `3xx` surfaces as `*APIError` so the key is never re-sent to a redirect target.
- Path segments percent-encoded (JID chars `@ : +` kept readable); base-URL path prefix preserved.
- Retry treats POST as non-idempotent on a network error to avoid double-sending a message; retryable HTTP statuses (429/5xx) still retry for any method because the server rejected the request before processing.

## Review fixes (in `fix(sdk/go): address PR #720 review feedback`)

Blockers
- **Bulk media 400.** New `BulkMediaContent` type without `chatId`, so image/video/audio/document bulk blocks no longer emit `"chatId":""` and are no longer rejected by `BulkMediaDto`'s `forbidNonWhitelisted`.
- **Status list decode.** `Status.List`/`FromContact` now unwrap the `{"statuses":[…]}` envelope and return `[]StatusRecord`. `StatusRecord` gains `Contact`/`Caption`/`MediaURL`/`ExpiresAt` to match the engine payload.
- **Health.Ready decode.** `HealthReadyResponse.Details` is `map[string]DependencyStatus`, matching the server's nested `{ mainDatabase: { status: "up" } }` — no more unmarshal failure on the healthy 200 path.

High
- **Webhook filter value.** `WebhookFilterCondition.Value` is `any`, `CaseSensitive` is exposed — text/boolean/id-array filters and read-back all round-trip cleanly.
- **Status POST response.** Send{Text,Image,Video}Status now returns `*StatusResult` (`statusId`/`timestamp`/`expiresAt`), matching the engine's `StatusResult` — no more empty `ID`.
- **Label color.** `LabelRecord.HexColor` uses `json:"hexColor"` — labels no longer decode a blank color.
- **Retry idempotency.** Non-idempotent methods (POST) are skipped after a network error; retryable HTTP statuses still retry for any method.
- **Retry-After tested.** Added tests for the seconds form, the HTTP-date form, and honoring `Retry-After` end-to-end.

Medium
- `WithTimeout(0)` now uses `DefaultTimeout` per its doc (was infinite).
- Split `SendMediaRequest` / `SendAudioRequest`; PTT moved onto the audio-only type so it can't 400 on the other endpoints.
- `GroupSummary.ParticipantsCount`/`IsAdmin` are pointers with a doc note that the LIST endpoint strips them (use `Groups.Get` for details).
- New `ErrBadRequest` sentinel for HTTP 400.
- `WithHeader` multi-value now applies **all** values, not just the first.
- Retry no longer retries `context.Canceled` / `context.DeadlineExceeded`.
- `SessionsService.List` accepts a `*ListSessionsQuery` with optional `limit`/`offset`.

## Tests

`go test -race ./...` — **113 tests, coverage 79.7%**, no deps.

- **`TestRouting`** — a table asserting the exact method + path of **all 90 service calls**. This is the drift gate that makes the historical `/messages/text` vs `/messages/send-text` class of bug fail at test time.
- Error mapping (typed + NestJS array-message + new `ErrBadRequest`), retry with body rewind, Retry-After (seconds + HTTP-date), non-idempotent skip vs. idempotent retry on network error, middleware pipeline, query encoding + typed-nil handling, redirect-not-followed, `WithHeader` multi-value, `WithTimeout(0)` fallback, `BulkMediaContent` `chatId` omission, `Status.List` envelope decode, `Health.Ready` `DependencyStatus` decode, context cancellation.

## Notes for the maintainer

- Verified locally: `go build ./...`, `go vet ./...`, `gofmt -l .` clean, `go test -race`.
- There is no Go CI in the repo; the SDK is self-contained under `sdk/go/` and doesn't touch the NestJS app or its pipeline.
- **CHANGELOG:** I did not push a `CHANGELOG.md` edit here (the file is ~240 KB and awkward to round-trip via the API). Suggested `[Unreleased] → Added` entry:

  > **Official Go SDK (`sdk/go`).** Hand-written, stdlib-only Go client covering the full user-facing API surface. `openwa.New(baseURL, apiKey, opts...)` entry point with per-domain service fields; context-first methods; functional options for config/DI (`WithHTTPClient`/`WithTransport`/`WithLogger`/`WithRetry`/`WithMiddleware`); composable transport middleware; typed errors (`*APIError` + `errors.Is` sentinels). A `TestRouting` table asserts the exact method+path of all 90 service calls so contract drift is caught at test time. Joins the JavaScript/Python/PHP/Java SDKs.

### DIFF
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -16,6 +16,7 @@ hand-written resource methods.
 | Python                  | [`rmyndharis-openwa`](python/)      | sync (httpx), PEP 561 typed |
 | PHP                     | [`rmyndharis/openwa`](php/)         | sync (Guzzle, PHP 8.1+)     |
 | Java                    | [`com.rmyndharis:openwa`](java/)    | sync (java.net.http + Gson, Java 17) |
+| Go                      | [`github.com/rmyndharis/OpenWA/sdk/go`](go/) | stdlib-only, context-first, injectable transport (Go 1.22+) |
 
 ## Coverage
 
@@ -160,6 +161,37 @@ Requires Java 17+. Errors are a typed, unchecked hierarchy — branch with
 custom `HttpTransport` that records the request — no network. See
 [`java/README.md`](java/README.md) for the full guide.
 
+## Go
+
+```bash
+go get github.com/rmyndharis/OpenWA/sdk/go
+```
+
+```go
+import (
+    "context"
+    openwa "github.com/rmyndharis/OpenWA/sdk/go"
+)
+
+client, err := openwa.New("http://localhost:2785", "owa_k1_…")
+if err != nil {
+    log.Fatal(err)
+}
+
+ctx := context.Background()
+client.Sessions.Start(ctx, "my-session")
+res, err := client.Messages.SendText(ctx, "my-session", openwa.SendTextRequest{
+    ChatID: "628123456789@c.us",
+    Text:   "Hello from the OpenWA Go SDK!",
+})
+fmt.Println(res.MessageID)
+```
+
+Requires Go 1.22+. Stdlib-only, context-first. Errors are typed — match with
+`errors.Is(err, openwa.ErrConflict)` or unwrap `*openwa.APIError` with
+`errors.As`. Inject an `http.RoundTripper` with `openwa.WithTransport(...)` for
+testing, retry, tracing, or metrics. See [`go/README.md`](go/README.md).
+
 ## Reliability & security
 
 - **Use HTTPS in production.** The API key is sent as `X-API-Key` on every
@@ -186,6 +218,8 @@ cd sdk/python && python -m pytest -q
 cd sdk/php && composer install && ./vendor/bin/phpunit
 # Java
 cd sdk/java && mvn -B verify
+# Go
+cd sdk/go && go test -race ./... && go vet ./...
 ```
 
 Each test suite mocks the HTTP layer and asserts on the exact path, so the
--- a/sdk/go/README.md
+++ b/sdk/go/README.md
@@ -1,0 +1,174 @@
+# OpenWA Go SDK
+
+Idiomatic Go client for the [OpenWA](https://github.com/rmyndharis/OpenWA) WhatsApp
+API Gateway. Stdlib-only (no dependencies), context-first, with typed errors and
+an injectable transport pipeline.
+
+```bash
+go get github.com/rmyndharis/OpenWA/sdk/go
+```
+
+Requires Go 1.22+.
+
+## Quick start
+
+```go
+package main
+
+import (
+	"context"
+	"log"
+
+	openwa "github.com/rmyndharis/OpenWA/sdk/go"
+)
+
+func main() {
+	client, err := openwa.New("http://localhost:2785", "owa_k1_…")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	ctx := context.Background()
+	if _, err := client.Sessions.Start(ctx, "my-session"); err != nil {
+		log.Fatal(err)
+	}
+
+	res, err := client.Messages.SendText(ctx, "my-session", openwa.SendTextRequest{
+		ChatID: "628123456789@c.us",
+		Text:   "Hello from the OpenWA Go SDK!",
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Println(res.MessageID)
+}
+```
+
+## Design
+
+- **Client entry point** — `openwa.New(baseURL, apiKey, opts...)` returns a
+  `*Client`. Required credentials are positional; everything else is a functional
+  Option. The client is safe for concurrent use.
+- **Services by domain** — the API is grouped onto exported fields:
+  `client.Sessions`, `client.Messages`, `client.Contacts`, `client.Groups`,
+  `client.Webhooks`, `client.Chats`, `client.Status`, `client.Labels`,
+  `client.Channels`, `client.Catalog`, `client.Templates`, `client.Health`,
+  `client.Search`, `client.Auth`.
+- **Context-first** — every network method takes `ctx context.Context` as its
+  first argument; the context bounds the request (and any retries).
+- **Functional options + DI** — inject dependencies instead of relying on
+  globals: `WithHTTPClient`, `WithTransport`, `WithLogger`, `WithRetry`,
+  `WithMiddleware`, `WithTimeout`, `WithUserAgent`, `WithHeader`.
+- **Typed errors** — match with `errors.Is` against the sentinels, or unwrap the
+  concrete `*APIError` with `errors.As`.
+
+## Configuration
+
+| Option | Purpose |
+| ------ | ------- |
+| `WithTimeout(d)` | Per-request timeout (default 30s). |
+| `WithHTTPClient(hc)` | Inject a preconfigured `*http.Client` (pool, jar, timeout). |
+| `WithTransport(rt)` | Inject the base `http.RoundTripper` (proxy, TLS, test double). |
+| `WithLogger(l)` | Inject a `Logger` (default: no-op). |
+| `WithRetry(p)` | Enable automatic retries (off by default). |
+| `WithMiddleware(mw...)` | Add transport middleware (tracing, metrics, auth). |
+| `WithUserAgent(ua)` | Override the `User-Agent`. |
+| `WithHeader(k, v)` | Add a default header on every request. |
+| `WithInsecureHTTP()` | Suppress the plaintext-`http://` warning. |
+
+## Typed errors
+
+```go
+res, err := client.Messages.SendText(ctx, "my-session", req)
+switch {
+case errors.Is(err, openwa.ErrConflict):
+	// 409 — engine not ready; retry once the session is "ready".
+case errors.Is(err, openwa.ErrNotFound):
+	// 404 — unknown session/resource.
+case err != nil:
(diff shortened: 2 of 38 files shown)


rmyndharis:
# PR #720 — Go SDK

Hey @Revelts, really nice work — stdlib-only, clean options/DI pipeline, typed errors, never-follow-redirects, `GetBody` rewind, and a `TestRouting` table. `gofmt`/`vet`/`build` clean and `go test -race` green at 77.1% (reproduced locally). Great bones.

A few things to address before merge — three are runtime bugs that bite on the first call. The fixes are small and mostly live in `types_*.go`. I traced each method against the actual controllers/DTOs, not just `TestRouting`.

## 🔴 Blockers

**B1. Bulk media sends return 400.** `BulkMessageContent` reuses `*SendMediaRequest`, whose `chatId` has no `omitempty` (`types_message.go:20`) — so every media block emits `"chatId":""`, which `BulkMediaDto` rejects via `forbidNonWhitelisted`. Text-only bulk works; image/video/audio/document bulk 400s. Fix: a dedicated bulk media type without `chatId`.

**B2. `Status.List`/`FromContact` decode into the wrong type** — `map[string][]StatusRecord` (`status.go:14`), but the server returns `{"statuses":[...]}` (`status.controller.ts:18`). The decode yields one bogus key `"statuses"`, and the "grouped by contact" doc isn't accurate. Fix: an envelope struct `{ Statuses []StatusRecord \`json:"statuses"\` }`.

**B3. `Health.Ready()` errors on a healthy gateway.** `Details map[string]string` (`types_misc.go:15`) vs the server's nested `{ mainDatabase: { status: "up" } }` (`health.controller.ts:9`) → unmarshal fails on the 200 path. Fix: `map[string]DependencyStatus`.

**B4.** Mind dropping the `Generated with [Claude Code]` line + commit markers? This project ships no AI attribution.

## 🟠 High

- **Webhook `Value []string` rejects half the filters.** Server `value` is `string|string[]|boolean` per kind (`filter-validation.ts:38`); `[]string` 400s `body`/`isGroup`/`fromMe`/`hasMedia`, and `caseSensitive` is missing. Same struct also breaks read-back unmarshal. (`types_webhook.go:6`)
- **`StatusRecord` drops the useful fields** — no `contact`/`caption`/`mediaUrl`/`expiresAt` (`types_status.go:4` vs `whatsapp-engine.interface.ts:238`); `statusId`/`body` aren't on the GET object.
- **Label color lost forever** — tagged `colorHex` but server sends `hexColor` (`types_misc.go:63`). One-liner: `json:"hexColor"`.
- **Retry can duplicate messages** — retries every network error regardless of method (`retry.go:115`), replaying non-idempotent POSTs verbatim. Worth guarding POST/DELETE, or sending an idempotency key.
- **Retry-After is untested** — `parseRetryAfter`/`DefaultRetryPolicy` are 0% covered (`retry.go:36,79`); the one retry test disables the feature.

## 🟡 Medium

- `WithTimeout(0)` → infinite timeout, contradicting its "Zero uses DefaultTimeout" doc (`client.go:114`).
- `PTT` sits on the shared media struct but the server only accepts it on `send-audio` → 400 if set elsewhere (`send-message.dto.ts:107`).
- `GroupSummary.ParticipantsCount`/`IsAdmin` always decode `0`/`false` (LIST strips them — `session.service.ts:1533`).
- `Status.Send*` return `*StatusRecord` but the server returns `StatusResult{statusId,timestamp,expiresAt}` → `ID` empty, `expiresAt` dropped.
- No `ErrBadRequest` sentinel for HTTP 400 (`errors.go:20`).
- `WithHeader` multi-value keeps only the first (`transport.go:55`).
- Retry retries `context.Canceled`/`DeadlineExceeded` (`retry.go:115`).
- `SessionsService.List` has no pagination (`sessions.go:10`).

## Why most of this slipped through

`TestRouting` asserts method+path only — never the request body or response shape. Adding a body-key assertion to a couple of mutating methods and a decoded canned response to the reads would have caught B1/B2/B3/H1/H2/H3 at test time.

Nice SDK underneath it all — the fixes are small. Happy to help with any of them. 🙂

Revelts:
Hi @rmyndharis Thanks a lot for the thorough review! I really appreciate you taking the time to trace everything against the controllers/DTOs and not just the routing tests.

I've pushed a new set of commits addressing a number of the issues you pointed out. The updates include fixes for several of the blockers and high-priority items, along with additional tests where appropriate.

There are still a couple of items I'm reviewing separately, but I'd appreciate it if you could take another look at the latest commits when you have a chance. Thanks again for the detailed feedback—it was extremely helpful.

rmyndharis:
Thanks @Revelts — this is a really thorough fix pass. I re-ran `gofmt`/`vet`/`build`/`go test -race` (green, 79.7%, 113 tests) and traced each item against the current code. The blockers and highs are all resolved:

✅ B1 bulk-media `chatId`, B2 status envelope, B3 health `details`, H1 webhook `Value`, H2 `StatusRecord` fields, H3 `hexColor`, H4 network-error replay, H5 Retry-After tests, plus the mediums — `WithTimeout(0)`, the `SendAudioRequest`/PTT split, `GroupSummary` pointers, `StatusResult`, `ErrBadRequest`, `WithHeader` multi-value, `Sessions.List` pagination, context-error short-circuit. Nice calls on splitting `SendAudioRequest` and making the `GroupSummary` fields pointers with the doc note — both the right shape.

Two small things from this round:

🟡 **Bulk audio lost PTT.** The new `BulkMediaContent` (no `PTT`) replaced `*SendMediaRequest` on `BulkMessageContent.Audio`, so bulk audio can't be sent as a voice note any more — but the server's `BulkMediaDto` still accepts `ptt`. Easy fix: mirror `BulkMediaDto` in `BulkMediaContent` (`url/base64/mimetype/filename/ptt`) — which also drops the `Caption` field, since `BulkMediaDto` doesn't whitelist it (it'd 400 if set).

🟡 **Retry on 5xx still replays POSTs** (the H4 residual). Network-error replay is correctly fixed; but a retryable *status* (500/502/504 in the default policy) still retries any method, including POST. The "server rejected before processing" reasoning holds for 429/503, but 502/504 are proxy codes that can land *after* the backend has already sent the WhatsApp message — so a verbatim replay could double-send. Retry is opt-in / off by default so this isn't a blocker, but worth either gating status-retries to idempotent methods too, or trimming the default to `429`/`503` for non-idempotent ones.

Still open — all minor/doc, fine as follow-ups:
- `sdk/README.md`: "No automatic retries" now contradicts Go's `WithRetry`; the Go quick-start snippet is missing `fmt`/`log` imports.
- `auth.go` doc comment points at `auth.controller.ts`, but `Validate` lives in `auth-validate.controller.ts`.
- Insecure-`http://` warning is silent under the default `nopLogger`.
- `MessageRecord` still drops server `chatName`; status `timestamp`/`expiresAt` are typed `any`.
- `send-poll` / `@mentions` / api-key CRUD still unexposed — looks like a family-wide scope call across all the SDKs, so probably intentional.

On my side I'll add the Go CI job and the CHANGELOG `[Unreleased]` entry (`Thanks @Revelts.`) at merge.

The bulk-audio PTT tweak is the only thing I'd want in before merge — everything else can ride as follow-ups. Once that lands this looks good to go. Thanks again for the quality work here. 🙂

PR #734, opened 2026-07-14
feat(messages): synchronize edited messages in real time

rogeriorioli:
## Description

Adds end-to-end `message.edited` support for both WhatsApp engines. Edited bodies and media captions are persisted in occurrence order, delivered through webhooks and WebSocket subscriptions, and reflected in the Chats dashboard without a refresh.

## What changed

- Normalize edit payloads across whatsapp-web.js and Baileys, including sender, direction, group, type, media, mention, and occurrence-time fields.
- Serialize message mutations so rapid edits remain latest-write-wins and persistence completes before consumer notification.
- Support smart webhook filters and stable per-occurrence idempotency keys for edited messages.
- Update stored chat previews and dashboard caches without allowing edits to older messages to reorder conversations.
- Add the event to dashboard selectors, SDK event types, OpenAPI, API documentation, operational documentation, and the changelog.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Verification

- [x] Backend lint, full-program typecheck, formatting, build, OpenAPI, and version checks
- [x] 2,498 backend unit tests and 53 e2e tests
- [x] Dashboard lint, typecheck, i18n parity, production build, and 112 unit tests
- [x] JavaScript, Python, PHP, Java, and Go SDK suites
- [x] Deep self-review of adapters, persistence ordering, webhook filters, WebSocket delivery, dashboard caches, and downstream documentation

## Contributor

Thanks @rogeriorioli for the contribution.


rmyndharis:
Thanks @rogeriorioli — this is a clean, well-scoped addition. The event flows through the stack exactly the way the existing `message.*` events do (interface → both adapters → service → gateway → idempotency → docs), the protocolMessage handling sits in the right branch, and the happy-path test coverage across all three layers is good. A few things to tighten up before merge — all small.

## Must-fix before merge

**1. `openapi.json` not regenerated — the `openapi:check` drift gate will fail.**
`message.edited` is added to `WEBHOOK_EVENTS` in `src/modules/webhook/dto/webhook.dto.ts`, and that array is spread directly into the `events` enum on both `CreateWebhookDto` and `UpdateWebhookDto` (`@ApiPropertyOptional({ enum: [...WEBHOOK_EVENTS, '*'] })`). Those enums land in the committed `openapi.json` (~lines 5847 and 5989), but this PR doesn't touch `openapi.json`, so `npm run openapi:check` (`… && diff -u openapi.json …`) will go red on the next CI run. CI hasn't run on this fork branch yet, which is why it's latent.

Fix: run `npm run openapi:export`, commit the regenerated `openapi.json`, verify with `npm run openapi:check` (should exit 0).

**2. Idempotency key collides on sequential edits of the same message (wwjs).**
The new key is `edit_{sessionId}_{messageId}_{timestamp}`. For the whatsapp-web.js adapter, `message.timestamp` is the message's **immutable original-send time** (`data.t` — "Unix timestamp for when the message was created"), not the edit time. So editing the same message twice produces an **identical** idempotency key. OpenWA doesn't self-dedup — the key is sent as the `X-OpenWA-Idempotency-Key` header — so a consumer honoring the documented dedup contract treats the second edit as a replay and drops it (silent data loss). The Baileys path is fine (`msg.messageTimestamp` is the edit's own time), but the key should be correct for both adapters. (It's partly masked today by upstream wwjs only firing `message_edit` on the first edit, but the key is structurally non-unique by design.)

Fix: mirror what `message.reaction` already does — drop `timestamp` and salt with the per-dispatch `occurrence` (`occurredAt`), whose comment spells out the identical rationale:

```ts
case 'message.edited':
  return `edit_${toStr(data.sessionId)}_${toStr(data.messageId)}${occurrence}`;
```

Keep the docstring in `docs/06-api-specification.md` in sync.

## Correctness gaps

**3. Media caption edits are lost and clobber the stored row (Baileys).**
Body extraction only checks `pm.editedMessage?.conversation || pm.editedMessage?.extendedTextMessage?.text`. `editedMessage` is a full `Message`, so an edited image/video/document caption arrives as `editedMessage.imageMessage.caption` (etc.) → body resolves to `''`, and then `session.service`'s `onMessageEdited` runs `messageRepository.update({...}, { body: '' })`, **destructively overwriting** the previously stored caption. The helper to do this correctly already exists and is already imported — `extractBaileysBody` (in `baileys-message-mapper.ts`), which `mapMessage` uses for live inbound:

```ts
const body = extractBaileysBody(pm.editedMessage ?? {});
```

Worth adding an edge-case test (`editedMessage: { imageMessage: { caption: '…' } }`).

**4. Wrong `senderId` for your own outgoing edits (Baileys).**
The edit branch sets `senderId: …(msg.key.participant ?? remoteJid)` with no `fromMe` branch, unlike the REVOKE sibling right above it (`fromMe ? normalizedSelfJid() : remoteJid`). An edit of your own message (synced from a linked device, arrives as `notify` with `fromMe=true`) in a 1:1 chat has `participant === undefined`, so `senderId` resolves to the **recipient** instead of yourself. (The `fromMe` skip in `handleMessagesUpsert` only applies to non-`notify` batches, so self-edits do reach this branch.)

Fix: mirror the REVOKE pattern —

```ts
senderId: this.sessionStore.toNeutralJid(
  msg.key.fromMe === true ? this.normalizedSelfJid() : (msg.key.participant ?? remoteJid),
),
```

## Nits

- **No `message.edited` case in `idempotency.util.spec.ts`** — every sibling event (`received`/`ack`/`revoked`/`reaction`) has one. Add a retry-stability + distinct-occurrence case (especially once #2 changes the key shape).
- **PR template** is left unfilled — checklist boxes, `Closes #`, and the description placeholder. Minor.

## Note

Heads-up for the changelog when this ships: subscribers using the `*` wildcard will start receiving `message.edited` too (inherent in adding any event). Under 0.x SemVer this is a patch (additive, non-breaking) — worth flagging the wildcard behavior in the release notes regardless.

---

Architecture and conventions are solid — once #1 is in and #2–#4 are addressed (a single follow-up commit should cover all of them), this is good to merge. Thanks again for the contribution.

rmyndharis:
Status update, @rogeriorioli — this is still on my list, not stalled, and no rush from my side. I've re-read the diff against current `main` and found two things my earlier review missed. Both are the kind you can't hit locally, so I'd rather hand them over now than have you discover them on a second round-trip.

## 1. The WebSocket drift guard will fail `npm test`

`events.gateway.spec.ts` has a guard that reflects over the gateway prototype, calls every `emit*` method, and asserts the emitted set equals `SUBSCRIBABLE_EVENTS` **in both directions**:

```ts
expect(new Set(SUBSCRIBABLE_EVENTS)).toEqual(deriveEmittedEvents());
```

This PR adds `emitMessageEdited` to `events.gateway.ts` but doesn't touch `src/modules/events/dto/ws-messages.dto.ts`, so the guard sees 10 emitters against 9 advertised events and the suite goes red. The fix is one line — add `'message.edited'` to `SUBSCRIBABLE_EVENTS`:

```ts
export const SUBSCRIBABLE_EVENTS = [
  'message.received',
  'message.sent',
  'message.ack',
  'message.revoked',
  'message.reaction',
  'message.edited',   // <- here
  'session.status',
  ...
] as const;
```

I looked at whether I could just land that on `main` ahead of you and save you the trouble. I can't: the guard runs the other way too, so advertising `message.edited` without an emitter present would turn `main` red instead. The two halves have to arrive together, which means in this PR. Sorry — that one's an artifact of how the guard is built, not something you could have known.

## 2. The SDK event enums

All the typed SDKs enumerate the webhook event union by hand (`sdk/javascript/src/types.ts`, `sdk/python/openwa/types.py`, `sdk/java/.../model/`, `sdk/go`). Nothing cross-checks them against `WEBHOOK_EVENTS`, so SDK CI stays green while a TypeScript user writing `events: ['message.edited']` fails to compile against a server that accepts it. We closed exactly this class of drift last week in #754/#777, so I'd rather not add a fresh instance in the same release. Adding `'message.edited'` to each SDK's union is mechanical — happy to take that as a follow-up PR instead if you'd prefer to keep this one focused; just say so and I'll do it once this lands.

## The four points from my review still stand

Re-checked against the current diff — none are addressed yet, and I don't think any of them are controversial:

1. **`openapi.json` is still drifted.** The PR touches `src/modules/webhook/dto/webhook.dto.ts`, which is a contract surface, so `openapi:check` (a `diff -u` gate) fails until you run `npm run openapi:export` and commit the result. This is the hard merge-blocker.
2. **wwjs idempotency key** — still `edit_{sessionId}_{messageId}_{timestamp}`. The point stands as written: that key is structurally non-unique for the case it exists to protect.
3. **Baileys caption clobber** — unchanged.
4. **Baileys self-edit `senderId`** — unchanged.

One correction to my own review while I'm here: I noted the Baileys path is fine on timestamps (`msg.messageTimestamp` is the edit's own time) and that's still true — don't go changing that one.

## On CI

No checks have run on this branch yet, which is why 1 and 2 are latent rather than red. That's a fork-PR approval thing on our side, not anything you did.

## The feature itself

Wanted, and unsuperseded — `git grep -i message.edited` on `main` is empty, both engines expose the source event, and it slots into the same path `message.revoked` and `message.reaction` already take. I'm not looking for a reason to say no here; I'd like this in. Take whatever time you need.

celcoin-carlosorioli:
@rmyndharis Thank you for the detailed feedback and breakdown!

First of all, I want to apologize for the delay in pushing the fixes. My computer is currently at the service center for repairs, so I am temporarily blocked from making changes. I expect to have it back by next week, and I will send over all the improvements.
I really appreciate you taking the time to review this so thoroughly and for keeping it open. I'll get back to work on this next week!

rogeriorioli:
@rmyndharis 

***

Thanks for the detailed review! I've updated the PR with a follow-up commit to address all these points. Here's a summary of the changes:

### Must-fix before merge
1. **openapi.json regenerated**: Ran `npm run openapi:export` and verified that the drift gate `npm run openapi:check` passes successfully.
2. **wwjs idempotency key collision**: Switched the `message.edited` idempotency key to mirror reactions by using the `occurredAt` dispatch salt (`edit_{sessionId}_{messageId}_{occurredAt}`). This ensures sequential edits of the same message do not share identical keys. Also updated the docstring in `docs/06-api-specification.md` and added tests verifying this retry-stability + distinct-occurrence behavior in `idempotency.util.spec.ts`.

### Correctness gaps
3. **Media caption edits (Baileys)**: Replaced the manual text checks with `extractBaileysBody(pm.editedMessage ?? {})` to handle image/video/document captions correctly. Added an edge-case test in `baileys.adapter.spec.ts` to verify captions are preserved.
4. **Self-edit `senderId` (Baileys)**: Mirrored the `REVOKE` pattern in the edit branch so self outgoing edits (`fromMe === true`) resolve to the bot's own self JID. Added a test validating this JID mapping in `baileys.adapter.spec.ts`.

### Nits & Drift Guards
- **WebSocket Drift Guard**: Added `'message.edited'` to `SUBSCRIBABLE_EVENTS` in `src/modules/events/dto/ws-messages.dto.ts`. Verified the drift guard passes.
- **SDK Event Unions**: Added `'message.edited'` / `MESSAGE_EDITED` to the webhook event unions/enums in the JavaScript, Python, and Java SDK types.
- **Imports**: Fixed missing `EditedMessage` imports in both `baileys.adapter.ts` and `whatsapp-web-js.adapter.ts` that were causing compilation errors.
- **PR Description**: Updated the PR template description with the checklist items and details.

PR #737, opened 2026-07-14
feat/session webhook events

rafaelmerlotto:
## Description
this PR adds webhook dispatch support for the `session.status` event.

changes:
- added `session.status` webhook dispatch whenever a session status changes.

This allows the webhook subscribers to receive session status updates as expected.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

N/A

## Related Issues
N/A

### DIFF
--- a/src/modules/session/session.service.ts
+++ b/src/modules/session/session.service.ts
@@ -43,7 +43,7 @@ export class SessionService implements OnModuleDestroy, OnModuleInit {
     private readonly eventsGateway: EventsGateway,
     private readonly webhookService: WebhookService,
     private readonly hookManager: HookManager,
-  ) {}
+  ) { }
 
   /**
    * On backend startup, reset all active session statuses to disconnected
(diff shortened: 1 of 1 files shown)


rmyndharis:
Hi @rafaelmerlotto — thanks for this, and for thinking about webhook coverage for session status. It's a genuinely useful event to surface.

This is already covered on `main`, though. Session-status dispatch shipped back in **v0.4.3 (#354)** from `SessionService.updateStatus()`, and was hardened in **#546** with a de-duplication guard — because engines like wwebjs signal one transition through *both* `onStateChanged` and a dedicated callback (`onQRCode` / `onDisconnected`), which would otherwise POST the same status twice. The current implementation also sends `{ sessionId, status }` per the contract in `docs/06-api-specification.md` (§6.6), so consumers know *which* session changed. That de-dup behavior is locked in by the `does not double-dispatch` cases in `session.service.spec.ts`.

As written, the unconditional block here would reintroduce that double-dispatch, and its `{ status }`-only payload wouldn't carry the `sessionId` the contract requires. The branch was also cut from before #354 landed, which is why it conflicts with `main`. If you rebase onto `main` you'll see the dispatch already in place.

Really appreciate you taking the time — happy to point you at a few areas that could use the help if you'd like to contribute again. Thanks!

rafaelmerlotto:
Hi @rmyndharis, I see now, I was missing the latest upstream changes, which explains why I couldn't find the session.status webhook dispatch locally.

Thanks for the clarification!

PR #749, opened 2026-07-15
chats and infrastructure removed from ui

kabir74705:
# OpenWA — change summary (UI / local work)

## Dashboard
- Honest KPIs (connected = READY only); no fake trends
- Paginate sessions/webhooks past 1000
- Charts: zero-fill time series + by-session chart
- Logs CSV export page size fixed (200)
- API Keys: usage + expires columns

## Sessions
- Hide Session ID when status is disconnected
- Prefer WhatsApp `pushName` as display name

## UI removals (nav/routes only)
- **Chats** — removed from nav + route; page/API kept
- **Infrastructure** — removed from nav + route; page/API kept

## Templates
- Add / insert / remove `{{var}}` variables in header/body/footer + preview


rmyndharis:
Thanks @kabir74705 — there's a lot of genuinely careful work in here, and you clearly studied the existing conventions: the new util tests follow the project's `node:test` runner, all 11 locale files are updated (no i18n gaps), and the new dashboard fields map onto real backend surfaces. Two of the changes even fix real bugs I'm happy to take. The reason I can't merge this as a single PR is scope, not quality — let me break it down.

## The ask: split into focused PRs

This PR bundles ~10 unrelated changes across 6 themes (KPIs, pagination, charts, session display, a 211-line template-variables feature, and two nav removals) with no linked issue and a title that only describes one of them. Bundled like this I can't review it atomically, revert one part if it regresses, or attribute it to a single changelog entry. The project's pattern is focused PRs, each tied to an issue. If you can carve these up I'll fast-track the good parts.

## Fast-track these as their own small PRs (real fixes)

**1. Honest dashboard KPIs.** The old card showed `stats.active` as "Active Sessions" with a green `+N` trend arrow. But `active` is `this.engines.size` (`session.service.ts` ~L1695) — running engine instances, including `qr_ready`/`connecting`/`initializing`, not connected sessions — and the `+${stats.ready}` "trend" wasn't a delta at all, just the ready count rendered as an up-arrow. Your change (headline = `ready`, transparent `{active} running · {total} total` detail, no fake arrow) is the correct fix. Ship it.

**2. Logs CSV export truncation.** Requesting `PAGE=500` while the backend clamps to `MAX_AUDIT_PAGE_SIZE = 200` (`audit.service.ts:10`) meant `res.data.length < 500` was always true, so the export silently stopped after the first 200 rows. Your `PAGE=200` + `=== 0` guard fixes it cleanly. Ship it.

## Separate feature PRs

**3. Template variable builder.** This is 211 lines of net-new UI (add/insert/remove `{{var}}` with cursor-position-aware insertion). It's a clean feature, not a fix, and deserves its own PR and review — including a note that substitution is preview-only (the API caller still fills values at send time), so reviewers know the scope. Please give it its own PR with a short description of the intended UX.

**4. Pagination + chart improvements.** `fetchAllPages` past the 1000-item window (backend supports it — `session.controller.ts:61`, `webhooks-list.controller.ts:19`), zero-filled time series, and the new by-session chart (consuming the existing `bySession` from `stats.service.ts:67`) are a coherent "dashboard stats" bundle. Group them in one PR. One note: the `listIncomplete` banner will rarely fire now that `fetchAllPages` walks to 50k — fine as a defensive touch, just flag if you intended it to surface more often.

## Needs a product decision before code

**5. `pushName` as the primary display name** and **6. hiding the Session ID when disconnected.** Both are real (`pushName` is returned at `session.service.ts:438`), but these change what operators see everywhere — the primary identifier switches from the gateway session name to the WhatsApp account name, and the session UUID disappears from stopped/disconnected cards. These are UX calls, not bug fixes, and hiding the ID can confuse operators who manage stopped sessions by UUID. Please open an issue first so we can agree on the behavior, then PR against it.

## Please drop the Chats + Infrastructure removal

These give the PR its title, but they remove two working pages from **both** the nav and the routes (`App.tsx`), orphaning the components. Hiding the Infrastructure page specifically removes the admin infra-status monitor — that's a tooling regression unless there's a stated reason, and "page/API kept" in the description is misleading since the route is gone, leaving dead unreachable code. If there's motivation behind this, let's discuss in an issue; otherwise please drop these hunks.

## Process

No CI has run on `front-end_updates` yet, so before re-requesting review please run the local gate (`lint` + `build` + `test:unit` for the dashboard) so a typecheck or test failure doesn't surface on review. And do update the PR title/body to match whatever each split PR actually contains.

Happy to help triage the split or re-review the focused PRs as they come in — thanks again for the careful work, the convention-matching especially stood out.

rmyndharis:
Following up on my earlier comment — I'm going to close this one, and I want to be straight about the fact that I'm changing my position rather than let it look like the plan I laid out an hour ago.

What I said earlier was: split this into focused PRs and I'll re-review them as they come in. Having also looked at #756 (a full 36k-line copy of the dashboard into `dashboard-ui/`), it's clear the two PRs are going in a direction the project can't absorb, and I'd rather tell you that now than have you spend a weekend splitting this one up against a review I'm no longer going to be able to give as described.

So, closing this — but the technical assessment in my earlier comment stands, and none of it was padding:

**The two bugs you found are real, and I'm taking them.** I verified both against source:

- **Logs CSV export truncation** — requesting `PAGE=500` against a backend that clamps to `MAX_AUDIT_PAGE_SIZE = 200` (`audit.service.ts:10`) meant `res.data.length < 500` was always true and every export silently stopped at 200 rows. That's a real bug that shipped, and nobody else reported it.
- **Dashboard KPI dishonesty** — `stats.active` is `this.engines.size` (`session.service.ts:1695`), which counts `qr_ready`/`connecting`/`initializing` engines, not connected sessions, and the `+${stats.ready}` "trend" was never a delta at all. Also real.

I'll land both, and the commits will credit you.

**Why closing rather than waiting for a split:** the reviewable core here is those two fixes — maybe 15 lines. The rest is ~890 lines spanning a template-variables feature, pagination, chart work, a display-name change, and the Chats/Infrastructure nav removals, none tied to an issue and several needing a product decision before any code is worth writing. Reviewing that as one unit isn't something I can do properly, and I don't want to keep it open implying otherwise.

**If you want to contribute again — and I hope you do** — the path that works here is: open an issue first, agree on the behavior, then a PR that does one thing. The convention-matching in this PR was genuinely good (the `node:test` util tests, all 11 locales updated, correctly consuming `bySession` and `pushName` from the existing backend), which is why I'd rather point you at a workflow that gets that work merged than have it stall in review again.

Thanks for the two catches — they were worth the review on their own.

rmyndharis:
Following up as promised — both fixes are now open, with credit to you:

- **#766** — Logs CSV export truncating at 200 rows
- **#767** — dashboard reporting `engines.size` as connected sessions + the fabricated trend arrow

Both CI green. Two notes on what changed from your versions:

On the CSV export, I took the diagnosis but not the `PAGE = 200` line — hardcoding the backend's clamp into the dashboard means the same bug returns silently the day `MAX_AUDIT_PAGE_SIZE` moves. The loop now terminates on the server-reported `total` instead, which is immune to the clamp. Your `=== 0` guard is in there as the empty-page case.

On the KPI card, your i18n strings for all 11 locales are used as-is, with one fix: `he.json` had `"Sessions מחוברים"` — the English word had survived. It's `"סשנים מחוברים"` now.

Thanks again. These were two real bugs sitting in shipped code that nobody had reported.

Issue #750, opened 2026-07-15
[Bug]: sending message is working but i am getting 400 as response


jithivasudev:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm on the latest released version (or noted my version below)

### OpenWA version

v0.7.20

### Deployment

Docker Compose

### Database

SQLite (default)

### WhatsApp engine

whatsapp-web.js

### What happened?

sending message is working but i am getting 400 as response
<img width="1622" height="872" alt="Image" src="https://github.com/user-attachments/assets/f5e26502-ded4-4fc2-bca5-dc989b83a657" />

### Steps to reproduce

1. message tester
2. enter the phone number
3. enter message
4. press send message
5. then it shows 400 as response

### Relevant logs

```shell

```

gadajatah:
https://github.com/rmyndharis/OpenWA/pull/752  You can try this; I hope it works.

PR #752, opened 2026-07-15
[SENDING MESSAGE GETTING 400] - to fix whatsapp-web-js.adapter.ts

gadajatah:
## Description
Brief description of changes

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] Lint passes
- [X] Self-reviewed

WhatsApp Image 2026-07-16 at 00 19 50

## Related Issues
Closes #

### DIFF
--- a/dashboard/package-lock.json
+++ b/dashboard/package-lock.json
@@ -700,9 +700,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -720,9 +717,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -740,9 +734,6 @@
         "ppc64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -760,9 +751,6 @@
         "s390x"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -780,9 +768,6 @@
         "x64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -800,9 +785,6 @@
         "x64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
--- a/src/engine/adapters/whatsapp-web-js.adapter.ts
+++ b/src/engine/adapters/whatsapp-web-js.adapter.ts
@@ -981,6 +981,30 @@ export class WhatsAppWebJsAdapter extends EventEmitter implements IWhatsAppEngin
     }
   }
 
+  /**
+   * Extract a `MessageResult` from the value returned by `client.sendMessage`. On some whatsapp-web.js
+   * versions, sending to a `@lid` recipient succeeds (the message is delivered) but the method returns
+   * `undefined` rather than a `Message` object, causing a runtime crash on `msg.id._serialized`.
+   * When `msg` is falsy we synthesise a best-effort id from the chatId and current timestamp so the
+   * caller can still record the outbound message without throwing.
+   */
+  private extractMessageResult(msg: Message | undefined | null, chatId: string): MessageResult {
+    if (msg?.id) {
+      return {
+        id: msg.id._serialized,
+        timestamp: msg.timestamp,
+      };
+    }
+    // Synthesise a stable-enough id for persistence when wwebjs returns nothing.
+    const ts = Math.floor(Date.now() / 1000);
+    const synthetic = `true_${chatId}_${ts}`;
+    this.logger.warn(
+      'sendMessage returned no Message object (known @lid quirk) — using synthetic id',
+      { chatId, syntheticId: synthetic },
+    );
+    return { id: synthetic, timestamp: ts };
+  }
+
   async sendTextMessage(chatId: string, text: string, mentions?: string[]): Promise<MessageResult> {
     this.ensureReady();
     // wwebjs accepts neutral `<phone>@c.us` WIDs directly as mentionedJidList, so no de-normalization
@@ -988,10 +1012,7 @@ export class WhatsAppWebJsAdapter extends EventEmitter implements IWhatsAppEngin
     const msg = await this.sendResolved(chatId, to =>
       mentions?.length ? this.client!.sendMessage(to, text, { mentions }) : this.client!.sendMessage(to, text),
     );
-    return {
-      id: msg.id._serialized,
-      timestamp: msg.timestamp,
-    };
+    return this.extractMessageResult(msg, chatId);
   }
 
   async sendImageMessage(chatId: string, media: MediaInput): Promise<MessageResult> {
@@ -1043,10 +1064,7 @@ export class WhatsAppWebJsAdapter extends EventEmitter implements IWhatsAppEngin
       }),
     );
 
-    return {
-      id: msg.id._serialized,
-      timestamp: msg.timestamp,
-    };
+    return this.extractMessageResult(msg, chatId);
   }
 
   async getContacts(): Promise<Contact[]> {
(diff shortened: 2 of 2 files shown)


rmyndharis:
Thanks for this, @gadajatah — the crash you're guarding against is real, and routing the send paths through one helper is the right shape. The fallback choice is the part I'd like to change before merging. Detail below.

## What's good

- The `msg?.id` guard is the correct fix for the `TypeError: Cannot read properties of undefined (reading 'id')` crash on `msg.id._serialized`.
- Coverage is complete: `sendImage/Video/Audio/Document` all funnel through the shared `sendMediaMessage` (one changed return covers all four), and `sendSticker/Location/Contact/Poll` + `replyToMessage` are changed directly. `forwardMessage` is correctly left alone, and `sendProduct/sendCatalog` throw `EngineNotSupportedError`. No `MessageResult` path was missed.
- Uses `this.logger.warn`, consistent with the repo's `LoggerService` convention.

## Requested changes

### 1. Don't synthesise an id — return empty, like `forwardMessage` already does

The repo already solves the identical situation ("the send happened but we have no usable id") in `forwardMessage`, and it deliberately chose the **opposite** of this PR. It returns `{ id: '' }` with a comment explaining why:

> `whatsapp-web-js.adapter.ts:1266-1271` — "a synthetic [id] would leave the forward stuck at SENT … unlike a synthetic or source id, which could cross-drive another row's delivery status."

There's even a dedicated test pinning that contract (`whatsapp-web-js.adapter.spec.ts:398-412`: "returns an explicit-unknown id (empty, not a real/synthetic id)").

Why the synthetic id is problematic here:

- **It can report success for a message that was never sent.** The docblock says "the message is delivered but the method returns undefined," but the upstream source contradicts that. `Client.sendMessage` in `whatsapp-web.js` returns `undefined` via two branches — `if (!chat) return null` (`Client.js:1539`, chat couldn't be resolved so **no send was attempted**) and `return msg ? getMessageModel(msg) : undefined` (`:1550`, the in-page send came back falsy). Neither means "delivered." When `extractMessageResult` hits a falsy `msg`, `persistSentState` (`message.service.ts:570`) does `if (result.id) message.waMessageId = result.id;` — any truthy id is treated as real, so `true_<chatId>_<ts>` is persisted as `waMessageId`, the row is marked `SENT` (`:571`), and the fabricated id goes back to the client as `messageId` with a 201. The helper treats every falsy return as success. Previously this path crashed on `msg.id._serialized` (TypeError → `failSend`: row = `FAILED`, `message:failed` webhook, 500). For a send endpoint, fail-loud is the right default.
- **It can collide on the unique index (narrow, but silent when it hits).** `message.entity.ts:39` declares `UQ_messages_sessionId_waMessageId`. The synthetic id is `true_<chatId>_<floor(Date.now()/1000)>` — deterministic per (chat, second), and the unique index does not exempt it the way it exempts NULL `waMessageId`. Two sends to the **same chat within the same second** while this branch is firing (a burst or immediate retry to one recipient — bulk fan-out to different chats is unaffected, since the `chatId` differs) produce an identical `(sessionId, waMessageId)`; the second `save()` violates the constraint, which `persistSentState`'s try/catch swallows (`:573-579`), so the client still gets 200 but the second row is stranded `PENDING` with a null `waMessageId`.

The empty-id fallback still achieves your crash-avoidance goal — `msg?.id` short-circuits the `TypeError` either way:

```ts
private extractMessageResult(msg: Message | undefined | null, chatId: string): MessageResult {
  if (msg?.id) {
    return { id: msg.id._serialized, timestamp: msg.timestamp };
  }
  this.logger.warn('sendMessage returned no Message object — leaving waMessageId unset', { chatId });
  return { id: '', timestamp: Math.floor(Date.now() / 1000) };
}
```

### 2. The trigger isn't `@lid`-specific

Upstream `Client.sendMessage` (`whatsapp-web.js/src/Client.js`) only special-cases recipients by `@newsletter` and `@broadcast` (`:1387-1432`) — never `@lid`. Its `undefined` returns come from `if (!chat) return null` (`:1539`) and `return msg ? getMessageModel(msg) : undefined` (`:1550`), both recipient-agnostic. OpenWA already resolves `@c.us`→`@lid` before the send, and this repo's own tests assert a send to a resolved `@lid` returns a normal `Message`. The real trigger is WhatsApp Web in-page version drift (same class as #747). The docblock and the `warn` message saying "known @lid quirk" will point future debugging at LID resolution, which is the wrong place — could you reword to recipient-agnostic?

### 3. The title says 400, but this path produces a 500

`msg.id._serialized` on an undefined `msg` throws a `TypeError`, which surfaces as a 500, not a 400. The send DTO only `@IsString`/`@IsNotEmpty`-validates `chatId` (no suffix gate), the controller doesn't reject `@lid`, and the error mapper only converts `SsrfBlockedError` to 400 — so a literal 400 isn't reachable on this path. If the reporter actually saw a 400, it came from a different layer and this PR wouldn't address it. Worth confirming the original symptom.

### 4. Add a test for the new branch

Every `sendMessage` mock in `whatsapp-web-js.adapter.spec.ts` resolves a real `Message`; none resolves `undefined`/`null`, so the falsy branch is uncovered. One parametrized case (mock `sendMessage → undefined`/`null`, assert the fallback shape and the warn) would lock it in. If you go with the empty-id approach from #1, the existing forward test at `:398-412` is the template.

### 5. Drop the `dashboard/package-lock.json` change

Those six removed `libc` arrays are npm-version-skew lockfile churn from a root `npm install`, unrelated to the adapter fix. Please revert that file (`git checkout -- dashboard/package-lock.json`) so the PR stays scoped to the bug.

---

Net: the guard is welcome, but please switch the fallback to the empty-id sentinel the repo already uses for this case, fix the `@lid` framing, add a test, and drop the lockfile hunk. Happy to merge once those land.

rmyndharis:
Closing this, and I owe you a straight explanation rather than a "superseded" label.

**Your diagnosis was right, and it was first.** `client.sendMessage()` resolving falsy while the send itself succeeds is exactly what was happening, and you identified it before #757, #750, and the two other reporters who hit the same thing. Routing every send site through one helper was also the right shape — [#762](https://github.com/rmyndharis/OpenWA/pull/762) does precisely that, and it covers the same sites you found.

**What happened on my side:** I reviewed this asking for five changes and said I'd merge once they landed. Then #757 came in with a full trace, two more reporters piled on, and the underlying cause turned out to be a WhatsApp Web build renaming `WAWebMsgKey._serialized` to `$1` — which broke media downloads and the chat list at the same time (#747). That got cut as v0.8.18, and I wrote the send-side fix directly rather than waiting on a round-trip here. That was my call about release timing, not a reflection on you or on this PR, and you'd had about a day to respond. I should have said that here at the time instead of leaving the PR open with a merge promise on it.

**One correction worth recording, because I got it wrong in my review.** I asked you to return an empty id instead of the synthetic one. That is not quite what shipped. The reason is in the part of my own review I didn't follow through on:

`Client.js:1558` collapses two opposite outcomes into a single `undefined` — `if (!chat) return null` (`Client.js:1539`: the chat couldn't be resolved, so **nothing was sent**) and a `Msg.get` miss (`Injected/Utils.js:585`: the message **was** dispatched and only its id was unreadable). An empty id means "sent, id unknown," so returning it for a bare `undefined` would claim a delivery that may never have happened — the same false-positive class as the synthetic id, just quieter. So #762 splits it:

- `msg` is **undefined** → ambiguous → reported as a **failed send**. A false negative is visible and retryable; a `201` for a message that never left is not recoverable.
- `msg` **exists** but its id is unreadable → dispatch is proven (whatsapp-web.js only builds a `Message` from a real message model) → the empty sentinel, never a synthesised id.

Your instinct that the synthetic id was "good enough for persistence" is where the two approaches actually part ways: `true_<chatId>_<ts>` looks exactly like a real WhatsApp id, so nothing downstream can tell it isn't one — it gets written to `waMessageId`, indexed, and ack-matched as genuine. That's why the objection was to the fabrication specifically, not to the helper.

There was also a piece neither of us could see from the adapter alone: the empty id is only safe because it lands as NULL. `UQ_messages_sessionId_waMessageId` is not a partial index, so `''` collides with the next id-less send in the same session — and the bulk path swallows that violation into a warning, losing the row silently. #762 normalizes at `saveOutgoingMessage` to close that.

The fix ships in the next release. Thanks for finding this and for taking the time to open a PR — please do send more; a report that comes with a diff and a screenshot of the actual failure is genuinely more useful than most issues, and this one drove a fix that four people needed.

PR #758, opened 2026-07-16
fix(wwjs): backport upstream #201832 to restore id._serialized access (#747)

### LINKED ISSUE #747: [Bug]:  WhatsApp pushed a WhatsApp Web update that changed how a photo/file is stored internally (the encryption key + path fields).  The break: to download a photo, the library reads specific fields, then does .

rmyndharis:
## What

Restores inbound media download, message ids, ack tracking, and reply quoting — all broken on ~2026-07-14 when WhatsApp Web build 2.3000.x renamed the internal serialized message-id property from `id._serialized` to `id.$1`. whatsapp-web.js 1.34.7 (the version OpenWA pins) reads `_serialized` in the Message constructor and ~40 downstream sites, so the rename took down `downloadMedia()`, message-id extraction, acks, and quoted-message resolution for every bot at once. Reported in #747.

## How

Backports upstream fix [wwebjs/whatsapp-web.js#201832](https://github.com/wwebjs/whatsapp-web.js/pull/201832) (`Base._normalizeId`, applied across the Message/Chat/Contact/Channel/Broadcast/GroupNotification/ClientInfo constructors plus `from`/`to`/`author` fallbacks and the Client/Utils paths) into `node_modules/whatsapp-web.js` at Docker build time.

The patcher (`scripts/patch-wwebjs-201832.js`):

- applies the **real upstream diff** via `patch` — not `git apply`, which silently no-ops under the parent repo's git discovery (documented in the script),
- asserts the reject set equals the one known harmless divergence (`Contact.js` hunk 2 targets a LID-aware block path absent in 1.34.7) and **loud-fails the build** on any other reject, so a future `whatsapp-web.js` bump that drifts the shape can't ship a silently partial patch,
- **feature-detects `Base._normalizeId`** and no-ops once it exists, so the backport auto-disables the moment a future `whatsapp-web.js` release ships the fix upstream.

A guard spec (`wwebjs-backport-201832.spec.ts`) runs the patcher against a temp copy of the installed dep and asserts the normalization sites land, self-removal triggers, and skew aborts — covering the boot-smoke blind spot (boot-smoke only hits `/api/health/live`).

## Scope

Stopgap, not a fork. The patcher is removed in the same change that bumps `whatsapp-web.js` to a release containing #201832. No `package.json` change (`patch` is a system package via apt, not an npm dep), so no lockfile churn.

## Verify

- `npm run lint` — 0 errors (4 pre-existing `baileys.adapter.ts` warnings, unchanged).
- `npm run build` — ok.
- `npm test` — 2418 passed, 0 failed (incl. the 3 new guard tests).
- Dashboard untouched by this change.

Fixes #747.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Inbound media download, message ids, acks, and reply quoting restored**
+  (whatsapp-web.js `id._serialized` → `id.$1` rename). WhatsApp Web build 2.3000.x
+  (rolled out ~2026-07-14) renamed the internal serialized message-id property
+  from `id._serialized` to `id.$1`, which broke whatsapp-web.js 1.34.7's
+  `downloadMedia()`, message-id extraction, ack tracking, and quoted-message
+  resolution for every bot at once. The production Docker image now backports
+  upstream fix [#201832](https://github.com/wwebjs/whatsapp-web.js/pull/201832)
+  (`Base._normalizeId`) into the installed dependency at build time via
+  `scripts/patch-wwebjs-201832.js`. The patcher applies the real upstream diff
+  (with a loud-fail guard against version skew) and auto-disables the moment a
+  future whatsapp-web.js release ships the fix, so it is a stopgap, not a fork.
+  Fixes #747.
+
 - **Engine start timeouts now return a diagnostic 504 instead of a bare 500.** Two
   `POST /api/sessions/:id/start` failure modes previously escaped to NestJS's default handler as a
   meaningless `500 Internal Server Error`: (1) the **auth-timeout** — whatsapp-web.js throws the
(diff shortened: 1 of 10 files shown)


rmyndharis:
Superseded by #759, which carried all four commits from this branch (the backport, the self-review fixes, the source-install wiring, and the half-apply guard) and merged as 3e3c167. Closing this one — the work is in `main`.

PR #759, opened 2026-07-16
fix(wwjs): apply the id backport on source installs, not just in Docker (#747)

rmyndharis:
Stacked on #758 — please merge that one first; this PR's diff is only meaningful with the patcher it adds.

## What

#758 wires the whatsapp-web.js id backport into the Dockerfile only. Anyone following [Option B: Local Development](https://github.com/rmyndharis/OpenWA#option-b-local-development) (`npm install`) therefore gets an unpatched dependency and still hits the broken media downloads from #747 — while the CHANGELOG announces the fix unconditionally. This closes that gap: `postinstall` applies the backport too.

## Why it isn't just "call the script from postinstall"

The patcher aborts when it can't apply. That's correct for the image — an unpatched dependency should fail the build rather than ship — but wrong for `npm install`:

- Windows outside WSL has no `patch` binary at all.
- A Baileys-only user has no reason to care about a whatsapp-web.js patch.
- CI runs `npm ci` in five places.

Failing anyone's install over a degraded-but-working dev setup is worse than the bug. So `postinstall` passes `--best-effort` (warn, exit 0); the image build keeps the strict behaviour. Both paths are covered by tests.

## Fixture correctness

The guard spec copied `node_modules/whatsapp-web.js` assuming it was pristine. Once `postinstall` patches it, that assumption breaks — the same specs would pass in the Docker builder (which installs before `scripts/` is copied) and fail on a dev machine. `copyWwjs()` now reverse-applies the same diff on its copy to reach a known-unpatched tree, so it behaves identically either way. Verified against both states.

## Verify

- `npm install` → `patch-wwebjs-201832: applied`, `node_modules/whatsapp-web.js` patched, exit 0.
- No `patch` on PATH + `--best-effort` → warns, exit 0. Same without the flag → exit 1.
- Guard spec green against a patched *and* a pristine `node_modules`.
- `npm run lint` 0 errors; `npm test` 2424 passed.

Refs #747.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Inbound media download, message ids, acks, and reply quoting restored**
+  (whatsapp-web.js `id._serialized` → `id.$1` rename). WhatsApp Web build 2.3000.x
+  (rolled out ~2026-07-14) renamed the internal serialized message-id property
+  from `id._serialized` to `id.$1`, which broke whatsapp-web.js 1.34.7's
+  `downloadMedia()`, message-id extraction, ack tracking, and quoted-message
+  resolution for every bot at once. The production Docker image now backports
+  upstream fix [#201832](https://github.com/wwebjs/whatsapp-web.js/pull/201832)
+  (`Base._normalizeId`) into the installed dependency at build time via
+  `scripts/patch-wwebjs-201832.js`. The patcher applies the real upstream diff
+  (with a loud-fail guard against version skew) and auto-disables the moment a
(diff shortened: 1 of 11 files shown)


rmyndharis:
Note for whoever reviews this: CI won't report here. `ci.yml` only triggers on pull requests targeting `main`/`develop`, and this one targets #758's branch — so the absence of checks is the trigger config, not a passing run. The checks will appear once #758 merges and GitHub retargets this PR to `main`.

Verified locally in the meantime:

- `npm run lint` — 0 errors (4 pre-existing `baileys.adapter.ts` warnings, unchanged)
- `npm run build` — ok
- `npm test` — 2424 passed, 0 failed
- A real `npm install` → `patch-wwebjs-201832: applied`, `node_modules/whatsapp-web.js` patched, exit 0
- Guard spec green against both a patched and a pristine `node_modules`

PR #760, opened 2026-07-16
chore(sdk): gate the Go SDK in CI and changelog it

rmyndharis:
Follow-up to #720, which added the Go SDK but landed without a CI job or a changelog entry.

## Changes

- **`sdk-ci.yml`: add the Go job.** Mirrors the existing per-SDK jobs: `gofmt` (fails on unformatted
  files), `go vet ./...`, `go test -race ./...`, working-directory `sdk/go`. The workflow's path
  filters already cover `sdk/**` plus the server contract surfaces (`src/**/dto/**`,
  `src/**/*.controller.ts`, the engine interface), so the SDK's `TestRouting` drift gate now re-runs
  whenever a controller or DTO changes — which is the point of that table.

  Caching is disabled deliberately: the SDK is stdlib-only, so there is no `go.sum` to cache and
  pointing `cache-dependency-path` at the absent file fails the step.

  `go-version` is pinned to `1.22` — the `go.mod` minimum — so CI gates the version the SDK claims to
  support rather than whatever is newest.

- **`CHANGELOG.md`: record the Go SDK** under `[Unreleased] → Added`, with contributor credit.

## Verification

`gofmt -l .` clean, `go vet ./...` clean, `go test -race ./...` passing (79.7% coverage) against the
merged SDK tree; the workflow YAML parses and resolves to five jobs.

### DIFF
--- a/.github/workflows/sdk-ci.yml
+++ b/.github/workflows/sdk-ci.yml
@@ -28,6 +28,9 @@ jobs:
           cache-dependency-path: sdk/javascript/package-lock.json
       - run: npm ci
       - run: npm test
+      # The build tsconfigs exclude test/, and vitest strips types without checking them, so the
+      # type-level wire contract in test/*.test-d.ts only gates if tsc runs over test/ too (#754).
+      - run: npm run typecheck
       - run: npm run build
       - run: npm run smoke # dual CJS+ESM consumable — must run after build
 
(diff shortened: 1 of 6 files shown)


rmyndharis:
Superseded — this branch was cut from a feature branch rather than `main`, so it pulled in an unrelated dashboard commit. Reopening from a clean `main` base.

manojdmali:
add new package in it for flow manage

PR #822, opened 2026-07-20
feat(baileys): make the linked-device name brandable via BAILEYS_BROWSER_NAME

clicsoluciones:
Operators white-labeling the stack need their own brand in WhatsApp's **Settings → Linked Devices** entry, which is currently hardcoded to `OpenWA`.

This makes the first element of `BAILEYS_BROWSER` configurable through a new optional env var, `BAILEYS_BROWSER_NAME` (default unchanged: `OpenWA`). Documented in `.env.example`.

Note: the name is fixed at pairing time — it applies to new QR pairings; existing sessions keep the identity they were linked with until re-paired.

## How to test
1. Set `BAILEYS_BROWSER_NAME=mybrand` on the container and start a session (`ENGINE_TYPE=baileys`).
2. Pair via QR and check the phone's Linked Devices list: the entry shows **mybrand (Chrome)**.
3. Unset the var and re-pair: entry shows **OpenWA (Chrome)** as before.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -118,6 +118,10 @@ BAILEYS_AUTH_DIR=./data/baileys
 # Pull the FULL message-history archive on connect (large). Off by default: the engine still enables the
 # initial sync (contacts, chats, recent messages, lid mappings) without downloading the entire history.
 BAILEYS_SYNC_FULL_HISTORY=false
+# Device name shown in WhatsApp Settings → Linked Devices for NEW pairings (existing sessions keep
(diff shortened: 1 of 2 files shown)


rmyndharis:
Merging this as the global fallback for device-name branding. Note for the roadmap: per-session and per-tenant resolution of the linked-device name is the superseding path — see docs/28-multitenancy.md (§28.6), where this exact env var is the final fallback in the resolution chain.

PR #840, opened 2026-07-20
fix(engine): clone puppeteer args to prevent multi-session browser crash

szmazhr:
### Description

This PR fixes a critical bug where starting multiple sessions simultaneously or sequentially with distinct configurations (e.g. session-specific proxy configurations) could lead to Chromium/Puppeteer crashes. 

### Cause of the Bug

In the original implementation, the array representing `puppeteerArgs` was initialized as:
```typescript
const puppeteerArgs = this.config.puppeteer?.args || [ ... ];
```
Since arrays in JavaScript/TypeScript are passed by reference, `this.config.puppeteer?.args` retrieved the direct reference to the shared, global configuration array. 

Later in the code, when session-specific parameters (such as proxy server credentials or other arguments) were configured, they were appended directly to `puppeteerArgs`. This mutated the global `this.config.puppeteer.args` array in memory. When subsequent sessions initialized, they inherited the mutated arguments from previous sessions, leading to duplicate or conflicting command-line options (e.g. multiple `--proxy-server` flags) and eventual browser launch failure/crashes.

### Solution

We cloned the arguments array using the spread operator (`[...this.config.puppeteer.args]`), ensuring that any mutations to `puppeteerArgs` remain local to the current session adapter instance:
```typescript
const puppeteerArgs = this.config.puppeteer?.args
  ? [...this.config.puppeteer.args]
  : [
      '--no-sandbox',
      '--disable-setuid-sandbox',
      '--disable-dev-shm-usage',
      '--disable-accelerated-2d-canvas',
      '--no-first-run',
      '--no-zygote',
      '--disable-gpu',
    ];
```

### Checklist
- [x] Tested starting multiple sessions.
- [x] Verified that global configuration is not mutated when session-specific arguments are appended.

### DIFF
--- a/src/engine/adapters/whatsapp-web-js.adapter.ts
+++ b/src/engine/adapters/whatsapp-web-js.adapter.ts
@@ -378,15 +378,17 @@ export class WhatsAppWebJsAdapter extends EventEmitter implements IWhatsAppEngin
 
     try {
       // Build puppeteer args, including proxy if configured
-      const puppeteerArgs = this.config.puppeteer?.args || [
-        '--no-sandbox',
-        '--disable-setuid-sandbox',
-        '--disable-dev-shm-usage',
-        '--disable-accelerated-2d-canvas',
-        '--no-first-run',
-        '--no-zygote',
-        '--disable-gpu',
-      ];
+      const puppeteerArgs = this.config.puppeteer?.args
+        ? [...this.config.puppeteer.args]
+        : [
+            '--no-sandbox',
+            '--disable-setuid-sandbox',
+            '--disable-dev-shm-usage',
+            '--disable-accelerated-2d-canvas',
+            '--no-first-run',
+            '--no-zygote',
+            '--disable-gpu',
+          ];
 
       // Add proxy configuration if provided — but only when the URL parses to a supported scheme, so
       // a malformed/stored proxy value can't break the Chromium launch or smuggle a non-proxy scheme.


rmyndharis:
Thank you for this, @szmazhr — really nice find.

Reference aliasing through a cached config tree is the kind of bug that's easy to stare past, because the code reads perfectly naturally and the array only misbehaves on the *second* session. Your write-up got the root cause exactly right, including the part that makes it non-obvious: `ConfigService.get()` returns a live reference into the configuration built once at load, so every adapter was appending to the same array. Diagnosing that from a Chromium crash is a good piece of work.

Your commit is now on `main` as [`74bcdb8`](https://github.com/rmyndharis/OpenWA/commit/74bcdb8), merged via #841. I rebased rather than squashed specifically so it keeps your authorship in the history rather than collapsing into mine. On top of it I added:

- a regression test — the existing suite couldn't have caught this, since every helper builds its adapter with `puppeteer: {}` and falls through to a fresh default literal, which is exactly the case that was never broken. The new test hands two adapters one shared array and asserts it survives untouched.
- a changelog entry under `[Unreleased]`, crediting you.

Worth mentioning: your fix resolves more than the crash you set out to fix. Because the `--openwa-session=<id>` marker is appended unconditionally, the leak wasn't limited to proxy users. Two further symptoms fall out of the same aliasing:

- a session configured **without** a proxy would still launch carrying the previous session's `--proxy-server`, and as the only proxy flag on the command line it took effect — so that session's traffic was silently routed through a proxy it was never configured to use.
- `killOrphanedChromiumProcesses()` identifies targets by substring-matching that marker against the full `ps` command line, so once later sessions carried earlier sessions' markers, restarting one session could `SIGKILL` another session's perfectly healthy browser.

Both are gone with the copy. There's a comment at `whatsapp-web-js.adapter.ts:472` asserting the sweep "cannot kill a live browser" — your patch is what makes that true again.

One housekeeping note in case it looked odd from your side: no CI ever reported on this PR. That wasn't anything you did — first-time fork contributions sit at `action_required` until a maintainer approves the workflow run, and the merge overtook it. Nothing was wrong with your branch.

Closing this in favour of #841, but the commit and the credit are yours. Thanks again — contributions like this one are genuinely valuable, and I'd be glad to see more from you.

PR #843, opened 2026-07-20
Improve session resilience and error handling for WhatsApp integration

Rohanx04:
## Description
Brief description of changes

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] Lint passes
- [ ] Self-reviewed

## Related Issues
Closes #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,42 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Outbound message edit.** `POST /api/sessions/:sessionId/messages/edit` edits the text of a
+  message sent by the account, on both engines (whatsapp-web.js `Message.edit`, Baileys
+  `sendMessage` with an `edit` key). Attempting to edit another sender's message fails with `403`,
+  an unknown message/chat with `404`, and the stored record's body is updated through the same
+  serialized mutation queue as inbound edit events. `message.edited` continues to cover inbound edits.
+- **Live group events.** `group.join`, `group.leave`, and `group.update` are now actually
+  dispatched — to webhooks (HMAC-signed, with stable idempotency keys) and to Socket.IO
+  subscribers — on both engines. They were previously accepted in subscriptions but never emitted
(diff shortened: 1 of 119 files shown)


rmyndharis:
Thanks for opening this, @Rohanx04 — but this PR is empty, so there's nothing to review yet.

The head and base commits are the same (`426603e`): the `main` branch on your fork is currently identical to upstream `main`, so GitHub reports 0 commits and 0 changed files. The description is also still the unedited template, and no issue is linked.

On the substance the title points at — session resilience and reconnect handling — that work has since landed in `main` via #798:

- `resolveReconnectConfig` now defaults `maxReconnectAttempts` to unlimited, so a long outage no longer declares a session terminally `FAILED`. An explicit `0` still means "auto-reconnect disabled".
- The attempt budget resets after 5 minutes of stable uptime, so unrelated transient drops don't accumulate against an explicit cap.
- Backoff is clamped finite and within `setTimeout`'s safe range, so a large exponent can't overflow and fire immediately.
- A `session.reconnect_loop` webhook plus a counter fires periodically, so an unlimited retry is still visible to an operator instead of silently spinning.
- On the Baileys side, `scheduleReconnect()` uses capped exponential backoff with no attempt ceiling; only `loggedOut` (401), `forbidden` (403), and `connectionReplaced` (440) are terminal, and a `connect()` failure mid-attempt reschedules rather than giving up.

Worth saying plainly: the diagnosis in your earlier #751 was correct. A fixed 5-attempt ceiling meant roughly two to three minutes of backoff before a session was declared dead and stayed down even after connectivity returned, which matches the "disconnected after a few hours and never came back" reports. You closed that PR yourself before it was reviewed — no hard feelings either way, but the underlying call was right.

I'm closing this one as there's no diff to act on. If you'd like to contribute again, two things will make it land smoothly:

1. Work on a topic branch on your fork rather than `main` — e.g. `git checkout -b fix/session-resilience`, commit there, then open the PR from that branch. PRs opened from a fork's `main` can't produce a diff once the fork is in sync, and they make rebases awkward for both sides.
2. Fill in the description and link the issue you're addressing (`Closes #NNN`), so the intended behaviour change is reviewable against a reported problem.

Happy to review a fresh PR whenever you're ready.

PR #852, opened 2026-07-21
fix: honour PUPPETEER_EXECUTABLE_PATH, and stop reporting delivered messages as failed

Ngonie69:
## Description

Two bugs found while bringing a session up on a Windows host that had no Puppeteer browser download, then sending through it.

### 1. `PUPPETEER_EXECUTABLE_PATH` is silently ignored

It is documented in `.env.example`, but the string appears nowhere in `src/`. Puppeteer therefore always resolves its own downloaded build, and a host with a working system Chrome still fails:

```
Error: Could not find Chrome (ver. 146.0.7680.31)
```

Resolved in `whatsapp-web-js.adapter.ts`, where the engine-plugin and factory creation paths converge, and also threaded through `configuration.ts` → `engine.factory.ts` → the plugin so it can be supplied either way. The browser actually launched is now logged.

### 2. Delivered messages are reported as failures

`sendTextMessage` (and five sibling methods) read `msg.id._serialized` from `client.sendMessage()`. That returns `undefined` rather than throwing, so the result was:

```
TypeError: Cannot read properties of undefined (reading 'id')   →   HTTP 500 "Internal server error"
```

**`undefined` does not mean the send failed.** `WWebJS.sendMessage` dispatches via `addAndSendMsgToChat` and only *then* looks the message back up by key:

```js
const [msgPromise] = require('WAWebSendMsgChatAction').addAndSendMsgToChat(chat, message);
await msgPromise;
return require('WAWebCollections').Msg.get(newMsgKey._serialized);   // returns nothing
```

Against current WhatsApp Web builds that lookup returns nothing, so a **delivered** message still comes back `undefined`. Confirmed by a message arriving on a real handset while the API reported a 500.

Observed on `whatsapp-web.js@1.34.7`, WhatsApp Web `2.3000.1043537533` / `2.3000.1043541377`.

#### Pinning the web version does not fix it

With `2.3000.1039777962` pinned and `webVersionCache: {type:'local', strict:true}` satisfied from cache, the page still loaded a newer build:

```
pinned=2.3000.1039777962   loaded=2.3000.1043541377
```

WhatsApp Web updates itself past the pinned index HTML, and two runs minutes apart loaded different builds. Recording this in case others reach for the same lever.

#### What this does instead

The library collapses "chat could not be opened" and "sent but not found" into the same `undefined`, so they cannot be distinguished. This errs deliberately toward reporting delivery:

- `MessageResult` gains `unconfirmed?: boolean` and a nullable `id`
- the API returns **201** with `messageId: null, unconfirmed: true`
- the stored message stays **`pending`** rather than `failed`
- the adapter logs a warning naming the chat

Reporting failure produces false negatives, and any caller that retries on failure re-sends a message the recipient already received.

`Message.waMessageId` needs an explicit `varchar` column type: TypeORM reflects `string | null` as `Object`, which SQLite rejects at startup. That crash is caught by neither the build nor the test suite.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Breaking change
- [ ] Documentation update

**Breaking:** `MessageResponseDto.messageId` becomes `string | null` and `MessageResult.id` likewise. Strictly-typed consumers will need updating. Happy to hide this behind an opt-in flag instead if you would rather not break the contract in a patch release.

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

**On tests:** none added. The existing suite has no adapter coverage to extend, and the behaviour needs a live WhatsApp session to reproduce. `npm run build` is clean, `npx eslint` is clean on the touched trees (two pre-existing errors elsewhere left alone), and all **110 existing tests pass**. Glad to add a unit test around the unconfirmed-result path if you would like one before merging.

Verified against a live gateway with a QR-linked session: a send that previously returned 500 now returns 201 with `unconfirmed: true`, is stored as `pending`, and the message reaches the handset.

## Related Issues
None found open for either bug.


Ngonie69:
Closing this — I opened it from a fork that turned out to be ~681 commits behind, and `main` already fixes both issues.

`PUPPETEER_EXECUTABLE_PATH` is threaded through config and used in the adapter, and `toMessageResult` already guards the absent-message case. Apologies for the noise.

On the second point you've clearly already reasoned about the ambiguity, and I have some field evidence that may be useful for it — that the message usually *is* delivered in that case, and that pinning the web version doesn't avoid it. Filing that as an issue rather than a patch, since it's information for your call rather than a change to it.

PR #870, opened 2026-07-22
docs(README): add example for connecting session until ready

MustafaTaj:
## Description
Adds a step-by-step API guide for the full session connection flow: create a session, explain every session status, start the engine, authenticate with QR or pairing code, and poll until status is `ready`. Also links the new example from `docs/README.md`.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist
- [ ] Tests added/updated
- [x] Documentation updated
- [ ] Lint passes
- [x] Self-reviewed

N/A — documentation only.

## Related Issues
Closes #

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -65,6 +65,7 @@
 
 | Example | Description |
 | ------- | ----------- |
+| [Connect Session Until Ready](./examples/session-connect-until-ready.md) | Create a session, understand all statuses, get QR or pairing code, poll until `ready` |
 | [Session Phone-Number Pairing](./examples/session-phone-number-pairing.md) | Link an existing WhatsApp account by phone number instead of scanning QR |
 | [Chat History Limits](./examples/chat-history-limits.md) | Understand local message history vs bounded live WhatsApp history |
 | [Webhook Signature Verification](./examples/webhook-signature-verification.md) | Verify signed OpenWA webhook deliveries in Node.js and Python |
--- a/docs/examples/session-connect-until-ready.md
+++ b/docs/examples/session-connect-until-ready.md
@@ -1,0 +1,411 @@
+# Connect a Session: Create → QR / Pairing Code → Ready
+
+This guide walks through the full OpenWA session lifecycle: create a session, understand every status, start the engine, authenticate with a **QR code** or **pairing code**, and poll until the session is `ready`.
+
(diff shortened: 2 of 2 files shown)


rmyndharis:
Thanks for this — a linear QR/pairing walkthrough is a real onboarding gap in `docs/examples/` (we only had `session-phone-number-pairing.md`), and the end-to-end scripts plus troubleshooting table are the genuinely useful part.

I verified the technical claims against the source and they hold up:

- The 7 statuses match `SessionStatus` exactly (`src/modules/session/entities/session.entity.ts`).
- HTTP codes are correct: `start` → 200, `pairing-code`/create → 201.
- The `name` rules (3–50, `^[a-zA-Z0-9-]+$`, duplicate → 409) match `CreateSessionDto` and the service's conflict check.
- Nice catch on `lastActiveAt` vs `lastActive`: `POST /api/sessions` returns the raw entity (so `lastActiveAt` is correct there), while the read/status routes go through `SessionResponseDto` and rename it to `lastActive`.

A few things before merge:

1. **`Closes #` is empty.** Please either fill in the issue number or drop the line.

2. **Trim the duplicated reference material.** The status table and the per-endpoint response JSON already live in `docs/06-api-specification.md` (status wire values, QR/pairing/start/create shapes, the `lastActiveAt` → `lastActive` rename) and in the interactive Swagger at `/api/docs`. Duplicating them here means two places to keep in sync, and reference JSON tends to rot. Could you keep the parts that are actually new — the flow diagram, the two end-to-end scripts, and the troubleshooting table — and link to §06 / Swagger for the schemas instead of re-embedding them? That keeps the doc focused as a tutorial and removes the drift risk.

3. **The create-response example is incomplete.** `POST /api/sessions` returns the raw entity, so it also exposes `config` / `proxyUrl` / `proxyType`, which the example omits. Either note that, or switch the example to `GET /api/sessions/:id` (the clean DTO shape) so readers aren't surprised by the extra fields.

(1) is required; (2) is what I'd really like to see; (3) is optional. Once those are in, happy to merge.

One process note for next time: this PR is opened from your fork's `main` branch — a dedicated feature branch keeps your `main` clean and makes follow-up PRs easier.

Issue #365, opened 2026-06-19
[Feature]: Voice-note transcription (speech-to-text) so bots/AI can read & reply to audio messages


ulises2k:
### Pre-flight

- [x] I searched existing issues and this isn't already requested.

## Problem / motivation

In many markets, **voice notes are the most common inbound message type** — users would rather talk than type. Today, when a contact sends a voice note, OpenWA correctly maps it (`audioMessage` → `voice`/`audio`) and the adapter already extracts the media (`mimetype` + base64) — see [`src/engine/adapters/baileys-message-mapper.ts`](../blob/main/src/engine/adapters/baileys-message-mapper.ts). But the `message.received` webhook / WebSocket payload only carries the **binary audio**, never **text**.

That means any downstream bot or AI auto-responder (an n8n flow, a Cloudflare Worker calling an LLM, a chatbot, etc.) **cannot "read" or reply to a voice note** unless it bolts on its own speech-to-text pipeline (download the audio → pick an STT provider → handle formats/auth/retries → re-inject the text). Every integrator re-implements the same plumbing, and the most natural use case — *"a user sends an audio, the AI understands it and replies"* — doesn't work out of the box.

OpenWA is already 90% of the way there: the audio bytes are in hand at the `message:received` boundary. The missing piece is an **optional transcription step** that turns that audio into text and attaches it to the message before it goes out.

## Proposed solution

Add an **optional transcription capability plugin** (Tier-2 extension, modeled exactly on the existing `translation` plugin from #294 — see [`src/plugins/extensions/translation/`](../blob/main/src/plugins/extensions/translation/index.ts)). It would:

1. Register on the `message:received` hook and act **only** on `type === 'voice' || type === 'audio'`.
2. Reuse the media already extracted by the adapter (`mimetype` + base64) — no second download.
3. Call a **configurable STT provider** (see below) and obtain the transcript.
4. Attach the result to the `IncomingMessage` so it flows through unchanged to the **`message.received` webhook** and the **WebSocket** event — exactly how `translation` enriches messages today.
5. Ship **disabled by default**, enabled via `POST /plugins/transcription/enable`, configured from the dashboard (same pattern as `translation`).

### Pluggable provider (no hard dependency / cost lock-in)

Mirror the `translation` plugin's provider abstraction (`LibreTranslateClient` behind a port). Suggested options, in order of "self-hosted first":

[...]

PR #873, opened 2026-07-22
feat(agent-context): add bounded canonical context endpoint

xbirimensah:
## Summary
- adds an opt-in (`AGENT_CONTEXT_ENABLED=true`), read-only GET endpoint for one selected inbound message
- projects bounded same-session context with canonical LID/direct-JID identity, typed quote/current-reaction data, and aggregate receipt status
- adds additive quote/reaction persistence and migration without reading legacy metadata or media JSON
- documents the generic local-bridge boundary and safety contract

## Safety
- route is absent by default; no send/search/list/write surface
- API-key session scope is enforced; cross-session IDs resolve as not found
- history is capped at six entries and 1,000 characters per body
- history is trigger-anchored; ambiguous same-second siblings are omitted to prevent post-trigger disclosure
- resolved aliases only: raw JID, direct-user dialects, and capped cached LID mappings; unresolved LIDs are never guessed

## Verification
- `npm test` — 207 suites passed, 2,864 tests passed; 5 existing skips
- focused Phase 2 suites — 287 tests passed / 6 suites
- agent-context E2E — 3 passed
- `npx tsc --noEmit`, `npm run build`, `npm run lint`, and `npm run format -- --check` — passed (one pre-existing unrelated lint warning)
- `npm --prefix dashboard run build` — passed

## Scope
No transcription provider, autoreply, calendar/CRM integration, or local bridge adapter is included.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -30,45 +30,46 @@
 
 **Full Index (by number)**
 
-| No  | Document                                                         | Description                                       |
-| --- | ---------------------------------------------------------------- | ------------------------------------------------- |
-| 01  | [Project Overview](./01-project-overview.md)                     | Vision, goals, scope, current status              |
-| 02  | [Requirements Specification](./02-requirements-specification.md) | Functional and non-functional requirements        |
-| 03  | [System Architecture](./03-system-architecture.md)               | Architecture, modules, and runtime flows          |
-| 04  | [Security Design](./04-security-design.md)                       | Auth, rate limiting, and security controls        |
-| 05  | [Database Design](./05-database-design.md)                       | Entities and storage considerations               |
-| 06  | [API Specification](./06-api-specification.md)                   | REST API and WebSocket protocol                   |
-| 07  | [API Collection](./07-api-collection.md)                         | Example requests and Postman import tips          |
-| 08  | [Development Guidelines](./08-development-guidelines.md)         | Coding standards and workflow                     |
-| 09  | [Testing Strategy](./09-testing-strategy.md)                     | Test types and tooling                            |
-| 10  | [DevOps & Infrastructure](./10-devops-infrastructure.md)         | Docker, deployment, and environment configuration |
-| 10-ID| [Panduan Docker (ID)](./DOCKER_ID.md)                            | Panduan deployment Docker dalam Bahasa Indonesia  |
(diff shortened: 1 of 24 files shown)


rmyndharis:
Thanks for this, @obirimensah05 — it's a genuinely careful piece of work. The identity discipline in particular is exactly right: keeping `LidMappingStore` as the sole resolver, treating a resolved mapping as an observation rather than a permanent claim, and never inferring a phone from a `@lid` all match the hard constraints we've built the LID layer around. The "OpenWA must start, send, receive and operate normally if the bridge is absent" principle and the test coverage are appreciated too.

That said, I'm going to close this, and I want to be transparent about why so the effort isn't wasted.

**The driving use case is already shipped.** The `bridge-hybrid-design.md` is anchored on transcription/agent context, which traces back to #365 (voice-note transcription). That issue is closed as *completed*: it ships today as the released `voice-transcription` plugin in `OpenWA-plugins`, which runs STT out-of-band and delivers a `message.transcription` webhook — deliberately off the critical path, since a `message:received` hook has a ~5s budget and blocks dispatch. The sanctioned *core* upgrade, if real-world demand justifies it, is a single first-class `message.transcription` event — not a new subsystem.

**This PR builds foundation for a consumer that lives outside the repo.** By the design doc's own framing, the whole right-hand side (local bridge, context adapter, transcription provider, agent) is deployment-local and "never packaged as a default OpenWA plugin." The `GET /api/sessions/:id/agent-context/messages/:id` endpoint here has no producer or consumer inside OpenWA — it exists only to serve that external bridge. With nothing in-tree depending on it, it's permanent API surface we'd carry for a speculative need (YAGNI).

**It also introduces a second source of truth.** The new typed `message_quotes` / `message_reactions` tables sit in parallel with the quote/reaction shapes already in `messages.metadata`, which the PR keeps writing "for compatibility." Two stores for the same data means double-write and drift risk for every consumer, for data the endpoint could already project from what we persist.

None of this is a knock on the execution — it's a scope call. OpenWA stays a general WhatsApp gateway; deployment-specific agent behavior belongs in a local bridge consuming our existing `message.received` webhook and message endpoints, which already carry what an external agent needs.

If you'd like to land something upstream from this, the piece I'd genuinely consider is a first-class `message.transcription` event (per the #365 discussion) — small, targeted, with a concrete consumer. Happy to spec that with you in a fresh issue. Thanks again for the thoughtful proposal. 🙏

PR #874, opened 2026-07-22
feat(annotations): add provider-neutral message annotations

xbirimensah:
> **Stacked on #873.** This branch builds on the Phase 2 agent-context commit, so #873 shows up here until it merges. The annotation slice genuinely depends on that work (`docs/bridge-hybrid-design.md` is introduced there), so it cannot be based on `main` as an independent branch. Review the second commit (`feat(annotations): ...`) for the changes this PR is about; merge #873 first and this diff reduces to that commit alone.

## What this adds

An additive, session-scoped store for provider-produced derived content, so a transcription provider can persist a transcript **beside** a message without widening any existing payload.

- **`message_annotations` table + entity**, keyed on `(sessionId, messageId, provider, kind)`, with an additive, idempotent, reversible migration.
- **Write-only `ctx.annotations.upsert` capability** behind a new `message-annotations:write` permission. The host binds `provider` to the manifest id (not caller-supplied) and applies the existing session-scope and activation checks. There is deliberately **no get, list, or search** operation.
- **Generic lifecycle hooks** `message:annotation-requested` and `message:annotation-updated`. Both carry text-free projections.
- **Bounded validation** for transcript text, language tag, SHA-256 media fingerprint, processor version, expiry, and a small scalar-only metadata object.

## What this deliberately does not add

No transcription provider, no external AI dependency, no media-read permission, and no annotation HTTP endpoint. Nothing is enabled by default.

Transcript text stays out of `Message`, webhook payloads, and agent context. Any future transcript retrieval needs an explicit, separately authorized read contract — and any future media processor needs a dedicated, separately reviewed media-read permission with its own consent/retention boundaries.

## Note on `createdAt`

`createdAt` is a bare `@CreateDateColumn`, matching every other entity in the project. Pairing a generated create-date with `dateColumnType()`/`DateTransformer` resolves to `text` on SQLite, which TypeORM does not treat as a date type — it then writes NULL and **every** annotation write trips the `NOT NULL` constraint.

Because the service spec mocks the repository, it cannot catch a disagreement between the entity and the migration that creates its table. `message-annotation.persistence.spec.ts` drives the entity through the real migration on `better-sqlite3` to keep that contract covered.

## Test plan

- [x] `npm test` — 210 suites passed, 2885 tests (5 pre-existing skips)
- [x] `npm --prefix dashboard run build`
- [x] `npx tsc --noEmit`
- [x] `npm run build`
- [x] `npm run lint` — 0 errors (1 pre-existing unrelated warning in `dto-strict-coercion.spec.ts`)
- [x] `npm run format -- --check`
- [x] `git diff --check`
- [x] New persistence spec verifies insert, update-on-conflict, and round-trip of `expiresAt`/`metadata`/`externalProcessing` against the migrated SQLite schema

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Plugins can now store bounded, provider-neutral transcript annotations without changing normal message or webhook payloads.** The additive `message_annotations` store is scoped to a session and a persisted message, validates a small typed transcript record, and keeps transcript text outside `Message.metadata`. An opt-in plugin declares `message-annotations:write` to call the write-only `ctx.annotations.upsert` capability; it cannot list, search, or read annotations through this foundation. Eligible live media emits minimal annotation lifecycle hooks, while no transcription provider, external AI dependency, or annotation REST endpoint is enabled by default.
+
 ## [0.10.6] - 2026-07-22
 
 ### Changed
--- a/docs/19-plugin-architecture.md
+++ b/docs/19-plugin-architecture.md
@@ -124,7 +124,7 @@ a host version. The config schema is the top-level `configSchema` (note: not nes
 
   "main": "dist/index.js",
 
-  "permissions": ["messages:send", "engine:read", "net:fetch"],
+  "permissions": ["messages:send", "engine:read", "net:fetch", "message-annotations:write"],
 
   "sessionScoped": true,
   "sessions": ["*"],
@@ -279,6 +279,7 @@ export interface PluginContext {
   messages: PluginMessagingCapability; // requires 'messages:send'
   engine: PluginEngineReadCapability;  // requires 'engine:read' (read-only)
   net: PluginNetCapability;            // requires 'net:fetch' (SSRF-guarded, net.allow-scoped)
+  annotations: PluginMessageAnnotationsCapability; // requires 'message-annotations:write' (write-only)
 }
 
 export interface PluginLogger {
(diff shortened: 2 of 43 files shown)


rmyndharis:
Thanks, @obirimensah05 — closing this alongside #873, since it's the next phase of the same bridge-hybrid initiative (stacked on #873, so it re-carries the agent-context module, identity engine, migration, and design doc on top of the annotations slice). The full reasoning is in the #873 comment; the short version: the transcription use case this is foundation for already ships as the `voice-transcription` plugin in `OpenWA-plugins` (#365, closed as completed), and the sanctioned core upgrade there is a single `message.transcription` event, not a new extension surface.

Specifically for the annotations work here:

- The `message-annotations:write` capability, the `message-annotations:write` permission, and the `message:annotation-requested` / `message:annotation-updated` hooks all become permanent parts of the plugin contract the moment they merge — but nothing in-tree writes or reads an annotation. There's no provider producing them and no read path consuming them (by design: "no get, list, or search"). That's a contract we'd maintain forever for a consumer that lives outside the repo.
- The `createdAt` / SQLite `NOT NULL` note and the real-migration persistence spec are good catches and good practice — no complaint on the engineering.

I do think the annotations *store* is the most reusable idea in the two PRs — a typed place to persist derived content beside a message is a reasonable primitive. But it earns its place once there's a concrete core consumer for it (the `message.transcription` event being the obvious candidate), not as a standalone write-only endpoint with nothing on either side. Introducing it before then locks in a schema and a permission we'd have to live with regardless of how that consumer actually turns out.

So: closing for now, with the annotations table explicitly on the table to revisit the moment we scope that event. If you're interested, that's the collaboration I'd point you to — open an issue for `message.transcription` and we can shape the store around a real writer/reader. Thanks again for the depth here. 🙏

PR #888, opened 2026-07-23
fix(ssrf): cancel unread withSafeFetch bodies to stop undici process crashes

### LINKED ISSUE #887: Process crash: undici TypeError: terminated (ECONNRESET) when webhook withSafeFetch does not consume response body
## Summary

Self-hosted OpenWA on Render (`ghcr.io/rmyndharis/openwa:latest`, Baileys engine) intermittently crashes the whole Node process with:

```
Uncaught exception (uncaughtException) — process will exit
TypeError: terminated
  at Fetch.onAborted (node:internal/deps/undici/undici:…)
  …
  [cause]: Error: read ECONNRESET
    errno: -104
    code: 'ECONNRESET'
    syscall: 'read'
```

Node then exits (by design via `registerUncaughtExceptionMonitor`), and the orchestrator restarts the instance. This briefly takes all WhatsApp sessions offline.

## Observed occurrences (production logs)

| UTC timestamp | Event |
|---|---|
| 2026-07-21 ~22:36 | Instance restarted (+ HTTP 502s on dashboard assets) |
| 2026-07-22 10:44:45 | Identical `TypeError: terminated` + `ECONNRESET` → exit → restart |
| 2026-07-22 13:32:55 | Same stack again → exit → restart |

Runtime: Node.js v22.23.1 inside the published image. Engine: Baileys. Two authenticated sessions auto-start after each restart.

## Stack (verbatim)

```
TypeError: terminated
    at Fetch.onAborted (node:internal/deps/undici/undici:11457:53)
    at Fetch.emit (node:events:519:28)
    at Fetch.terminate (node:internal/deps/undici/undici:10615:14)
    at Object.onError (node:internal/deps/undici/undici:11580:38)
    at LegacyHandlerWrapper.onResponseError (/app/node_modules/undici/lib/dispatcher/dispatcher1-wrapper.js:47:19)
    at Request.onResponseError (/app/node_modules/undici/lib/core/request.js:425:44)
    at Object.errorRequest (/app/node_modules/undici/lib/core/util.js:843:13)
    at TLSSocket.onHttpSocketClose (/app/node_modules/undici/lib/dispatcher/client-h1.js:1035:10)
…
Emitted 'error' event on Readable instance at:
…
  [cause]: Error: read ECONNRESET
```

This matches the known undici pattern where a mid-stream / socket-close error is emitted on the response body `Readable` **outside** the `await fetch()` / `try/catch` path, becoming a process-fatal `uncaughtException` ([nodejs/undici#3813](https://github.com/nodejs/undici/issues/3813), similar to [earendil-works/pi#6133](https://github.com/earendil-works/pi/issues/6133)).

### 1. Webhook processor does not consume the response body

In `src/modules/queue/processors/webhook.processor.ts`, the `withSafeFetch` callback only reads status metadata:

```ts
response => ({ status: response.status, statusText: response.statusText, ok: response.ok })
```

It never drains or cancels `response.body`.

### 2. `withSafeFetch` destroys the dispatcher immediately after `use` settles

In `src/common/security/ssrf-guard.ts`, `withSafeFetch` documents that `use` must finish with the response before return, then tears down the per-request Agent in `finally`:

```ts
try {
  const response = await undiciFetch(rawUrl, { ...init, redirect: 'manual', dispatcher });
  assertNoRedirect(response, rawUrl);
  return await use(response);
} finally {
  if (dispatcher) await dispatcher.destroy().catch(() => undefined);
}
```

If the body is still open when `dispatcher.destroy()` runs (or the peer resets the TLS socket while an unread body stream exists), undici can emit `TypeError: terminated` / `ECONNRESET` on the `Readable` with no listener → process crash.

### 3. Intentional fatal posture amplifies the outage

`src/config/process-error-monitor.ts` / `src/main.ts` only **log** via `uncaughtExceptionMonitor` and deliberately do **not** swallow the exception, so one stray undici stream error takes down every session.

## Suggested fix

1. **Always consume or cancel the body** in the webhook (and any other status-only) `withSafeFetch` callbacks, e.g.:

   ```ts
   async response => {
     try {
       return { status: response.status, statusText: response.statusText, ok: response.ok };
     } finally {
       await response.body?.cancel().catch(() => undefined);
     }
   }
   ```

2. Harden `withSafeFetch` itself: before destroying the dispatcher, cancel an unused body; optionally attach a no-op `error` listener on the body stream / dispatcher so secondary undici `error` events cannot become process-fatal.

3. Optionally treat transient network causes (`ECONNRESET`, `TypeError: terminated`, `UND_ERR_*`) as retryable job failures instead of uncaught exceptions (without weakening the intentional crash-on-true-corruption policy for other errors).

Secondary call site to review with the same pattern: `src/common/media/load-remote-media.ts` (`response.body.getReader()` mid-stream reset).

## Environment

- Image: `ghcr.io/rmyndharis/openwa:latest`
- Host: Render web service (single instance, persistent disk at `/app/data`)
- Engine: Baileys
- Health check: `/api/health/ready`

Happy to provide more log excerpts if useful. Thanks!

waldyrlimaverde:
## Summary
- Cancels unread undici response bodies in `withSafeFetch` before tearing down the per-request dispatcher, so status-only webhook deliveries cannot turn a peer `ECONNRESET` into a process-fatal `uncaughtException`.
- Adds regression tests covering cancel-on-unread, skip-when-consumed, and cancel-when-`use` throws.
- Documents the fix under `CHANGELOG.md` `[Unreleased]`.

Fixes #887

## Test plan
- [x] `npx jest src/common/security/ssrf-guard.spec.ts --no-coverage` (84 passed)
- [ ] Confirm webhook delivery still records success/failure from status alone
- [ ] After merge/release, deploy image and watch Render logs for absence of `TypeError: terminated` / `Uncaught exception (uncaughtException)`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Outbound `withSafeFetch` no longer crashes the process on unread webhook response bodies.**
+  Status-only callers (queued webhook delivery, webhook test, deprecated direct delivery) left the
+  undici response body unread; when the peer reset the TLS socket — or the per-request dispatcher
+  was destroyed — undici could emit `TypeError: terminated` / `ECONNRESET` on the body stream with
+  no listener, becoming a fatal `uncaughtException` that took every WhatsApp session offline (#887).
+  `withSafeFetch` now cancels unread bodies before tearing down the connection; callers that already
+  stream the body (media / plugin downloads) are unchanged.
+
 ## [0.10.8] - 2026-07-23
 
 ### Added
--- a/src/common/security/ssrf-guard.spec.ts
+++ b/src/common/security/ssrf-guard.spec.ts
@@ -310,6 +310,64 @@ describe('withSafeFetch (guarded + pinned fetch)', () => {
     );
     expect(use).not.toHaveBeenCalled();
   });
+
+  it('cancels an unread response body before tearing down the dispatcher (#887)', async () => {
+    // Status-only callers leave the body unread; if we destroy the Agent while the stream is still
+    // open, undici can emit TypeError: terminated / ECONNRESET as an uncaughtException. Cancelling
+    // the unread body before destroy closes that path.
+    const cancel = jest.fn().mockResolvedValue(undefined);
+    (dnsPromises.lookup as jest.Mock).mockResolvedValueOnce([{ address: '93.184.216.34', family: 4 }]);
+    (undiciFetch as jest.Mock).mockResolvedValue({
+      status: 200,
+      type: 'basic',
+      bodyUsed: false,
+      body: { cancel },
+    });
+    const use = jest.fn(() => ({ ok: true, status: 200 }));
+
+    await expect(withSafeFetch('https://example.com/hook', {}, use, { guard: true })).resolves.toEqual({
+      ok: true,
+      status: 200,
+    });
+    expect(cancel).toHaveBeenCalledTimes(1);
+  });
+
+  it('does not cancel a body the caller already consumed', async () => {
+    const cancel = jest.fn().mockResolvedValue(undefined);
+    (dnsPromises.lookup as jest.Mock).mockResolvedValueOnce([{ address: '93.184.216.34', family: 4 }]);
+    (undiciFetch as jest.Mock).mockResolvedValue({
+      status: 200,
+      type: 'basic',
+      bodyUsed: true,
+      body: { cancel },
+    });
+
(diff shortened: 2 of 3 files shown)


rmyndharis:
Thanks @waldyrlimaverde — this is a clean, well-tested fix, and placing it centrally in `withSafeFetch` is better than patching each caller. Nice work. 🙌

I pushed one small follow-up commit on top (yours are untouched): the redirect-refusal path was still unsettled. When `assertNoRedirect` throws on a 3xx, execution skipped the settle wrapper, so a refused redirect could still reach `dispatcher.destroy()` with an unread body — the same crash vector as #887, e.g. when a webhook endpoint starts returning http→https redirects. The body is now cancelled on that path too, with a regression test (`85 passed` in `ssrf-guard.spec.ts`).

With that covered, this looks good to merge. 👍

PR #895, opened 2026-07-24
add .agents to .gitignore and add id.json for Indonesian locale.

okabrionz:
## Description

This pull request adds support for the **Indonesian (`id`)** language locale in the dashboard and updates `.gitignore` to ignore the `.agents` directory.

### Key Changes:

* **i18n / Localization:**
* Added `id.json` containing full Indonesian translations for UI components, navigation, chats, and error messages.
* Registered `id` in `supportedLanguages`, `languageOptions`, and `i18n` setup in `dashboard/src/i18n/index.ts`.

* **Repository Maintenance:**
* Added `.agents/` to `.gitignore`.

---

## Type of Change

* [ ] Bug fix
* [x] New feature
* [ ] Breaking change
* [ ] Documentation update

---

## Checklist

* [ ] Tests added/updated
* [x] Documentation updated
* [x] Lint passes
* [x] Self-reviewed


rmyndharis:
Thanks for the contribution! The Indonesian locale is well-structured and the translations look solid. However, there's a blocker below that needs to be fixed before this can merge, plus a couple of smaller points.

### Blocker: `id.json` is not valid JSON

`dashboard/src/i18n/locales/id.json` fails to parse:

```
SyntaxError: Bad control character in string literal in JSON
  at position 21971 (line 604, column 47)
```

**Line 604** contains a raw newline inside the string value:

```json
"bulkRecipientsPlaceholder": "+62812345678

",
```

For comparison, `en.json` (line 604) is a clean single-line value:

```json
"bulkRecipientsPlaceholder": "+62812345678",
```

All 12 existing locales parse without error, so this is a regression introduced here. Since `index.ts` does `import id from './locales/id.json'`, the dashboard build will fail once this file is imported.

**Line 610** also has a broken UTF-8 sequence in `bulkDelayHint`:

```
1000�60000 ms (bawaan 3000)
```

`en.json` has a proper en-dash: `1000–60000 ms`. It looks like the en-dash was corrupted into an invalid byte sequence.

Both need fixing. The "Lint passes" checkbox in the PR description can't be accurate while the JSON is invalid — worth re-running `npm run lint` / a build after the fix.

### What's good

- **Key structure is complete.** After repairing the broken line for comparison, `id.json` has all **720/720** keys matching `en.json` — 0 missing, 0 extra, 0 placeholder mismatches (`{{count}}` etc. all preserved).
- **Translation quality.** The 31 strings where `id` equals `en` are almost all legitimate loanwords that are correctly left untranslated in standard Indonesian (`Status`, `URL`, `Port`, `Host`, `Admin`, `Redis`, `SQLite`, …). This reads as a competent translation, not a lazy one.

### `.gitignore` change

`.gitignore` already ignores `.agent/` (singular). Adding `.agents/` is a variant spelling — it's fine in itself, but:

- It mixes two unrelated changes (locale + gitignore) into one PR. These should ideally be separate PRs / commits.
- It looks like it's serving your local tooling rather than the project. If you don't have a strong reason to land it here, consider dropping it so this PR stays focused on the locale.

Also note: the diff removes the trailing newline at the end of `.gitignore` (`\ No newline at end of file`). Minor, but worth restoring.

### Minor

- Language label is `'Indonesia'`. The more common endonym is `'Bahasa Indonesia'` — the other entries use endonyms (`Français`, `Italiano`, `Português`). Up to you, but consistency would favor `'Bahasa Indonesia'`.
- The second commit message ("fix: update Indonesian translation to use Indonesian language") is a bit unclear — a more specific message helps reviewers follow what changed.

### Suggested follow-up (optional)

A small CI check that validates every locale JSON parses would prevent broken locale files from slipping in again — something like looping `node -e "JSON.parse(require('fs').readFileSync(f))"` over `dashboard/src/i18n/locales/*.json`.

rmyndharis (changes requested):
Thanks for the locale — the translations themselves read well, and the `index.ts` wiring is correct. A few things need fixing before this can merge:

1. **`id.json` is invalid JSON (build-breaking).** `bulkRecipientsPlaceholder` contains a raw newline inside the string, so `JSON.parse` throws and the dashboard build fails at import. The value should be a single line, same as the English source: `"+62812345678"`.

2. **37 keys are missing relative to `en.json`.** The branch is behind `main` — several recently landed features have no Indonesian strings: `sessions.unlink.*`, `plugins.catalog.*`, `plugins.installModal.catalog*`, `sessionStatus.action_required / failed / authenticating`, `sessions.start.teardownPending*`, and a few others. Please rebase and translate the full set (755 keys); untranslated keys silently fall back to English.

3. **One logical change per PR** (CONTRIBUTING.md). The `.agents/` `.gitignore` entry is unrelated to the locale — please drop it from this PR (it also removes the file's trailing newline).

4. **Language label:** use the native name, consistent with the other entries (`Français`, `한국어`): `Bahasa Indonesia` instead of `Indonesia`.

Happy to re-review once these are addressed.

rmyndharis:
Hi @okabrionz — checking back in on this one.

Two weeks on, the branch has drifted a long way from `main`: it's now **575 commits behind**, and GitHub reports it as conflicting. The gap also keeps widening the key list — my review on 3 Aug counted 37 missing keys, and today it's **45** (`en.json` is at 763 keys, this branch has 720). Two keys here are also now stale: `chats.status.placeholderTitle` and `chats.status.placeholderDesc` were removed upstream in `2d2ceb7d`, which landed about eleven hours after you opened this PR — so that one is drift, not anything you did wrong.

Nothing in the [earlier review](https://github.com/rmyndharis/OpenWA/pull/895#pullrequestreview-4840789023) has changed, and I won't restate it here. The only hard blocker is still a **single line** — `bulkRecipientsPlaceholder` at line 604 carries a raw newline inside the string, so `JSON.parse` throws and the dashboard build fails at import. Everything else is a rebase plus translating the 45 new keys, mostly from the `sessions.unlink.*`, `plugins.catalog.*` and `sessionStatus.*` groups.

I want to be clear that the translation itself is good work. 720 strings, complete placeholder parity, and the loanwords left in English (`Status`, `Redis`, `SQLite`, …) are the right call rather than a shortcut. Indonesian is still the locale this dashboard is missing, so I'd genuinely like it to land.

**Where this stands:** if I don't hear from you by **14 August**, I'll close the PR to keep the queue tidy. That would not be the end of the work — with your okay I'm happy to adopt the branch: rebase it, fix line 604, and translate the remaining keys myself, landing it with you as the commit author and credited in the CHANGELOG. You'd keep the attribution without doing the rebase.

So, whichever suits you:

1. **You finish it** — rebase onto `main`, fix line 604, add the 45 keys, drop the `.gitignore` change ([one logical change per PR](https://github.com/rmyndharis/OpenWA/blob/main/CONTRIBUTING.md)), and switch the label to `Bahasa Indonesia`. Ping me and I'll re-review promptly.
2. **I adopt it** — just reply saying so, and I'll take it from here with your authorship intact.
3. **No reply by 14 August** — I'll close this and follow up with option 2 anyway, since the translation is worth keeping.

No rush and no hard feelings either way. Thanks for putting the time into this.

PR #909, opened 2026-07-26
chore(security): fix Trivy HIGH findings and add scan workflow

ulises2k:
## Summary

Resolves **6 of 7 HIGH** findings from `trivy fs --severity CRITICAL,HIGH --ignore-unfixed` against `main` (scanned `23f85fb`). Adds a Trivy workflow so future regressions surface in the Security tab and block merges.

The one intentional exclusion is `react-router` 7 -> 8 in `dashboard/` — that's a major bump requiring a code migration (loaders, `RouterProvider`) and belongs in its own PR.

### Vulnerabilities fixed (npm overrides)

| Package | CVE / GHSA | Was | Now |
|---|---|---|---|
| `brace-expansion` | CVE-2026-14257 (ReDoS) | 2.1.2 and 5.0.7 | **5.0.8** |
| `js-yaml` | GHSA-pm4m-ph32-ghv5 (exponential parsing DoS) | 5.2.1 | **5.2.2** |

Added to the existing `overrides` block in `package.json`. The lockfile refresh (`npm install --package-lock-only`) consolidates the tree onto the fixed versions and drops 241 lines of duplicated ranges.

### Dockerfile misconfigurations fixed (Trivy DS-0029)

Added `--no-install-recommends` to both `apt-get install` invocations (build deps L17-21, Chromium/Puppeteer runtime deps L57-81). Smaller image, no Recommends surface, no behavior change for the explicitly listed packages.

### Dockerfile misconfiguration accepted (Trivy DS-0002, `USER` directive)

The Node process **does not run as root** — the ENTRYPOINT chain (`dumb-init` -> `docker-entrypoint.sh`) starts as root ONLY to `chown` the named volume mount and Chromium's XDG dirs, then `exec gosu openwa "\$@"` (see [docker-entrypoint.sh](docker-entrypoint.sh)). A bare `USER openwa` in the Dockerfile would break the chown-before-drop pattern and prevent the container from booting on fresh volume mounts (#254, #259).

Suppressed via `.trivyignore` with a comment explaining the mitigation.

### Recommended: Trivy workflow

`.github/workflows/trivy.yml` runs `trivy fs` on every push/PR to `main` plus a weekly Monday cron (catches newly published CVEs against pinned deps). SARIF upload -> Security tab -> Code scanning alerts; a second table-format run gates merges by failing on any fixed HIGH/CRITICAL that isn't listed in `.trivyignore`.

## Test plan

- [ ] CI: existing test suite still green
- [ ] CI: new `trivy` workflow runs and passes (0 HIGH/CRITICAL after this PR)
- [ ] Image build: `docker build .` succeeds on both amd64 and arm64 (`--no-install-recommends` change to Chromium block validated)
- [ ] Container boot: entrypoint still chowns `/app/data` correctly and drops to `openwa` via `gosu` (no `USER` regression)
- [ ] `npm ci` reproducible from the updated lockfile


rmyndharis (changes requested):
Thanks, @ulises2k — the dependency half of this is a real, verifiable win, and I'd like to take it. The Trivy workflow I'd like to hold back and redirect, for reasons below. Suggestion at the end is to split this into two PRs so the good part isn't held hostage by the part that needs rework.

## What I verified

I checked the premise independently against `npm audit` rather than Trivy, and it holds up. On `main`:

```
{"moderate": 2, "high": 30, "critical": 0}
```

All 30 HIGH trace back to exactly the two packages you identified — `brace-expansion` (`<=5.0.7`) and `js-yaml` (`5.0.0 - 5.2.1`). Everything else in that list (jest, glob, minimatch, test-exclude, archiver, fork-ts-checker) is transitive from those two. Auditing your lockfile in isolation:

```
{"moderate": 2, "high": 0, "critical": 0}
```

So the two `overrides` do exactly what you say they do, the tree consolidation is genuine (the lockfile gets *smaller*), and CI is green across Lint/Test/Build/Docker Build. No argument from me there.

I also want to note that the `.trivyignore` rationale for DS-0002 is correct — I traced it. `docker-entrypoint.sh:30` is `exec gosu openwa "$@"` after the chowns on lines 7 and 25, and the Dockerfile has no `USER` by design. That explanation is worth keeping regardless of what happens to the workflow; a comment above the `ENTRYPOINT` line would stop the next person from "fixing" it.

## The overrides miss `dashboard/`

`dashboard/package.json` has no `overrides` block at all, and `dashboard/package-lock.json` still resolves `brace-expansion@5.0.6` — the same advisory, and unlike `react-router` it's a trivial fix. `npm audit` in `dashboard/`:

```
brace-expansion   5.0.6            fixAvailable: true
react-router      HIGH             fixAvailable: true
react-router-dom  HIGH             fixAvailable: true
```

So "0 HIGH/CRITICAL after this PR" is true for the root lockfile only. This matters beyond bookkeeping — see the gate problem below.

## The workflow doesn't currently run

The `trivy-fs` check failed after three seconds:

```
##[error]Unable to resolve action `aquasecurity/trivy-action@0.28.0`, unable to find version `0.28.0`
```

Upstream tags carry a `v` prefix, so the ref needs to be `v0.28.0` — and that's eight releases behind `v0.36.0`. Nothing in the workflow has executed a single step yet, which means none of the behaviour in the test plan is confirmed, including whether `.trivyignore` suppresses `DS-0002` with the short ID rather than `AVD-DS-0002`.

Two further problems that surface once the ref is fixed:

**The gate is permanently red.** The second step combines `scan-ref: .`, `ignore-unfixed: true`, and `exit-code: 1`. That scans the whole repository, `dashboard/package-lock.json` included, where both `react-router` (acknowledged in the description but not in `.trivyignore`) and `brace-expansion@5.0.6` are HIGH with fixes available. The job fails on merge and every PR after it goes red.

**SARIF upload can't work on fork PRs.** The job log shows the token this PR actually got:

```
GITHUB_TOKEN Permissions
  SecurityEvents: read
```

GitHub downgrades `security-events: write` to read for pull requests from forks regardless of the `permissions:` block, and `upload-sarif` runs under `if: always()`, so it fails the job. Every external contribution would land with a red check for a permissions reason that has nothing to do with the contribution.

## Where I'd point the scanning effort instead

The dependency coverage `trivy fs` provides here already exists. `npm audit` runs on every PR in the `audit` job, and the reason it hasn't been catching these is the threshold, not the tool:

```diff
# .github/workflows/ci.yml:76
-        run: npm audit --audit-level=critical
+        run: npm audit --audit-level=high
```

With your overrides in place (plus the dashboard one), that one-word change turns the existing gate into exactly the regression fence this PR is trying to build, using the GitHub Advisory Database — which is richer for npm than Trivy's — and without a third-party action, an ignore file, or the fork-permissions problem. `react-router` then goes through the Dependabot major group that's already configured.

What that leaves uncovered is the thing Trivy is genuinely best at and this workflow doesn't do: OS package CVEs in the shipped image. `trivy fs` reads manifests in the working tree; the Debian packages inside `node:22-slim` that we actually hand to users are only visible to `trivy image` against the built artifact. That's a real gap — nothing currently watches it — and it's the same class of problem that broke the 0.10.3 image. The natural home is `release.yml`, after the image builds and alongside the existing boot-smoke and SBOM gates: it runs at release time rather than on every PR, and it never touches fork permissions. I'd very much welcome that as its own PR.

## One question on the Dockerfile change

`--no-install-recommends` is right in principle, but the arm64 branch installs Debian's `chromium` (Dockerfile L57-58), which *Recommends* `chromium-sandbox`. Our defaults are safe — `configuration.ts:108` forces `--no-sandbox,--disable-setuid-sandbox` — but `PUPPETEER_ARGS` is a documented env var, so an arm64 user who overrides it without `--no-sandbox` would get a Chromium that no longer starts. A successful `docker build` won't catch that. Did you get a chance to boot an arm64 image, or should I run that here?

## Suggested split

1. **Ready to merge quickly:** the two root `overrides`, the same `brace-expansion` override added to `dashboard/package.json`, `--no-install-recommends`, a `CHANGELOG.md` entry under `[Unreleased]`, and the DS-0002 explanation moved into a Dockerfile comment. I'll tighten the audit threshold to `high` on top of it.
2. **Separate PR, different shape:** `trivy image` in the release pipeline, if you're up for it.

The dependency work here is solid and I'd rather ship it this week than have it wait on the workflow. Thanks for digging into this. 🙏

ulises2k:
Thanks for the careful review, @rmyndharis — every one of those catches was right. Pushed `7b84a66` implementing your split:

**Done in this commit**
- `dashboard/package.json`: added `overrides` -> `brace-expansion: ^5.0.8`. `npm audit` in `dashboard/` now reports `HIGH: 2` — exactly `react-router` and `react-router-dom`, the deferred pair.
- `CHANGELOG.md`: entry under `[Unreleased]` covering both the dependency overrides and the `--no-install-recommends` change.
- **Removed** `.github/workflows/trivy.yml` and `.trivyignore`. You're right that `trivy fs` here duplicates what `npm audit` already covers, and the fork-permissions + red-gate problems you flagged are real.
- **Moved** the DS-0002 rationale into a comment right above `ENTRYPOINT` in the Dockerfile (natural home; doesn't need an ignore file once the workflow is gone).

**Left to you** (as you offered): the `--audit-level=critical` -> `high` bump on `ci.yml:76`. Happy to add it if you'd rather I keep everything in one PR — just say the word.

**On the arm64 `chromium-sandbox` question:** honest answer — no, I didn't boot an arm64 image. My dev host is amd64 and I don't have an arm64 WhatsApp session I can validate against without disrupting a production run. If you can boot the arm64 build here that'd be the right call; if it does turn out to hurt the `PUPPETEER_ARGS`-override path, I'd rather drop `--no-install-recommends` from the arm64 branch than fight it — the size win isn't worth breaking a documented env var. Your call on how you want to handle that.

**On the follow-up PR:** yes, I'll send `trivy image` in `release.yml` as a separate PR once this one lands. That was the more interesting problem anyway. 🙏

ulises2k:
Ran it on real arm64 hardware (Raspberry Pi 5, `aarch64`, Debian 12 bookworm via `node:22-slim` — the same base as the production image), in a throwaway `--rm` container so nothing touched a live session:

```
chromium Recommends:              chromium-sandbox
after --no-install-recommends:    chromium-sandbox ABSENT
chromium 150.0.7871.181 launch:   --no-sandbox --headless=new --dump-dom about:blank  →  exit 0, DOM returned
```

So:

- **Our default path is safe — confirmed empirically.** With `--no-sandbox` (forced at `configuration.ts:108`), Debian's `chromium` launches fine on arm64 without `chromium-sandbox` present.
- **Your premise is exactly right.** `chromium` pulls `chromium-sandbox` only via `Recommends`, so `--no-install-recommends` is precisely what drops it. Before this PR the sandbox binary was installed; after it, an arm64 user who overrides `PUPPETEER_ARGS` to remove `--no-sandbox` would newly get a Chromium that can't start. Narrow, but real and new.

Three ways to handle the edge case — your call:

1. **Accept as-is** — defaults are verified safe; note in docs that the arm64 override path now needs `chromium-sandbox` installed manually.
2. **Keep the flag but re-add the one package explicitly** — `apt-get install -y --no-install-recommends chromium chromium-sandbox` on the arm64 branch. Keeps the size win (drops every *other* Recommends) while preserving the override path. My preference.
3. **Drop `--no-install-recommends` from the arm64 `chromium` line only** — zero behavior change, loses the size win on arm64.

Happy to push whichever you pick (I'd default to #2). The amd64 branch is unaffected either way — it uses the downloaded Chrome-for-Testing binary, not the Debian `chromium` package, so its `Recommends` surface never included the sandbox.

ulises2k:
Pushed two more commits closing out the open items:

**`29e7113` — audit gate + arm64 sandbox**
- `ci.yml`: took your suggestion and bumped `npm audit --audit-level=critical` → `high` myself, so it's all in one PR (saves you the follow-up). Root-scoped; with the overrides here the root tree is 0 HIGH, so it now fences regressions. Dashboard keeps its own overrides + Dependabot for the react-router pair.
- `Dockerfile`: implemented option #2 from my arm64 findings — the arm64 branch now installs `chromium chromium-sandbox` explicitly, so `--no-install-recommends` still trims every *other* Recommends but the sandbox binary stays for the `PUPPETEER_ARGS`-override path. Zero regression, size win preserved. Went with #2 as the safe default; say the word if you'd rather have #1 or #3.

**Follow-up PR opened: #918** — `trivy image` in `release.yml`, gating `promote` on both arch scans, release-time only (no fork-permission issue). That's the OS-package coverage we actually wanted. Independent files, lands in either order.

I think that closes everything from your review here except your call on the arm64 approach. 🙏

rmyndharis (approved):
Thanks @ulises2k — this is all squared away. Approving.

The arm64 verification is exactly the kind of rigor I was hoping for, and the honesty about where you could and couldn't validate it (throwaway container on real Pi 5 hardware, no live session touched) is what made option #2 land cleanly. Keeping `chromium-sandbox` explicit while `--no-install-recommends` trims everything else is the right call — size win preserved, override path intact, and the comment in the Dockerfile will save the next person from undoing it. Verifying empirically rather than asserting from the docs is the difference between a change that's probably safe and one that's known safe.

Two things I appreciated beyond the original ask:

- Taking the audit threshold bump off my plate and getting it into one PR instead of a follow-up — and aligning `release.yml` to match, with the reasoning spelled out. "The release gate must not be laxer than the PR gate" is the correct framing; had you left it, it would've quietly become a bypass.
- Folding the DS-0002 rationale into a Dockerfile comment rather than parking it in `.trivyignore`. Once the workflow is gone, the ignore file had no anchor — the comment above `ENTRYPOINT` is where the next maintainer will actually look, and it reads as deliberate design, not a suppressed finding.

Merging this. I'll take #918 next.

On the broader note: your contributions on these last few PRs have been consistently high quality — well-scoped, independently verifiable, and you push back with evidence rather than deference when something's off (the Trivy rework being the clearest example). That's rare and genuinely useful to the project. Glad to have you on this.

PR #918, opened 2026-07-26
ci(release): scan release image for OS-package CVEs before promoting tags

ulises2k:
## Summary

Follow-up to #909, as discussed in review — the piece Trivy is genuinely best at and the source-tree scans can't do: **OS-package CVEs in the shipped image.**

Adds an `image-scan` job to `release.yml` that runs `trivy image` against the freshly-built **staging** image (before it gets release tags) on **both `linux/amd64` and `linux/arm64`**, and gates the `promote` step on it. A fixable HIGH/CRITICAL now stops the `X.Y.Z` / `X.Y` / `latest` tags from ever being applied.

## Why this and not `trivy fs` on every PR

`trivy fs` / `npm audit` read manifests in the source tree. The Debian packages baked into `node:22-slim` — and, **on arm64 only**, the `chromium` / `chromium-sandbox` packages the amd64 image never installs (amd64 uses a downloaded Chrome-for-Testing binary, not a `.deb`) — are visible only by scanning the built artifact. That's the same class of problem that broke the 0.10.3 image. Both architectures are scanned because their OS package sets genuinely diverge.

## Design notes

- **Placed at `promote`'s gate**, next to `boot-smoke`: an image that fails the scan never receives public release tags. Also added to the final Release job's `needs`.
- **Release-time only** (tag push), not on every PR — so it never hits the `security-events` / `packages` permission downgrade GitHub applies to fork PRs (the exact problem that would have made the push/PR workflow from #909 misbehave).
- **`ignore-unfixed`** — a CVE with no available fix does not block a release; only a fixable HIGH/CRITICAL does.
- **Static scan** — trivy reads image layers, it does not run the image, so no QEMU is needed for the arm64 variant, only a GHCR pull (authorized by the existing login).
- **`trivy-action` pinned to `v0.36.0`** (current release; tags carry the `v` prefix — the miss from #909).

## Test plan

- [ ] `release.yml` parses (validated locally with `yaml.safe_load`; job graph: `image-scan` needs `docker`, `promote` needs `[boot-smoke, image-scan]`).
- [ ] On the next tag, `image-scan (linux/amd64)` and `image-scan (linux/arm64)` both run after `docker` and before `promote`.
- [ ] A seeded fixable HIGH in the image fails the job and blocks promotion (confirms the gate bites).
- [ ] Clean image promotes as before (no false block from `ignore-unfixed`).

Depends on nothing in #909 — independent files (`release.yml` + `CHANGELOG.md`), can land in either order.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -209,7 +209,7 @@ jobs:
     # a matching container image, and a runtime boot check on both architectures (boot-smoke runs the
     # published image on amd64 + arm64). Trades a few minutes of buildx/boot latency for release
     # safety.
-    needs: [lint, test, test-postgres, dashboard, build, docker, boot-smoke, promote, verify-published]
+    needs: [lint, test, test-postgres, dashboard, build, docker, boot-smoke, image-scan, promote, verify-published]
     steps:
       - uses: actions/checkout@v7
         with:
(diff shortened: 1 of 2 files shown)


rmyndharis (approved):
Approving. This is exactly the piece we held back from #909 — `trivy image` against the built artifact, where the Debian packages baked into `node:22-slim` (and the arm64-only `chromium`/`chromium-sandbox`) are actually visible. The source-tree `npm audit` gate can't see those by construction.

A few things worth calling out, because they're easy to get wrong here and this gets them right:

**Gate placement is correct.** `promote.needs` now chains both `boot-smoke` and `image-scan`, and `release.needs` was updated to match — so a vulnerable image can't reach the public image tags *or* the GitHub Release. Both publication paths are fenced. Running it at release-time only sidesteps the fork-PR `security-events: write` downgrade that sank the push/PR workflow in the earlier draft of #909.

**The multi-arch mechanism is sound.** `TRIVY_PLATFORM` via env is the right approach — the action exposes no `platform` input, so the env var is the only path. I checked the old reports of `--platform` being silently ignored ([aquasecurity/trivy#4253](https://github.com/aquasecurity/trivy/issues/4253)); that's been closed-completed since May 2023, well before the v0.72 the action ships, so the arm64 variant is genuinely selected and scanned (no silent amd64 fallback masquerading as arm64 coverage). And the "no QEMU needed" note is correct — Trivy reads the manifest/layers statically, it doesn't run the image. `fail-fast: false` matches `boot-smoke`'s shape, so both arches report even when one fails.

**One non-blocking note for future maintainers** (not asking for a change): `vuln-type: os,library` means the scan also inspects the npm packages inside the image, which overlaps with the `npm audit` job earlier in the pipeline. The overlap is defensible — Trivy's DB and the GitHub Advisory DB that `npm audit` uses aren't identical, and a freshly-resolved `npm ci` tree in the image can diverge from the source lockfile — so it's defense-in-depth, not redundancy. The practical edge case to be aware of: the release could fail on a `library` HIGH that `npm audit` already passed because it's in Trivy's DB but not GHDB. If that ever bites, the fix is either a `.trivyignore` entry (now that there's a real workflow anchoring it) or narrowing `vuln-type` to `os`. Flagging so the next person doesn't treat it as a regression.

Merging.

PR #990, opened 2026-07-29
fix(engine): force document mode on whatsapp-web.js

### LINKED ISSUE #989: [Bug]: send-document does not force document mode on whatsapp-web.js
### Pre-flight

### OpenWA version

current `main` at `517a4dc`, after `v0.11.1`

### Deployment

Docker Compose

### Database

PostgreSQL

### WhatsApp engine

whatsapp-web.js

### What happened?

when i use the `send-document` endpoint with whatsapp-web.js, OpenWA calls the normal media send without telling whatsapp-web.js to send it as a document. in my setup i had to keep a custom build for documents to arrive with the correct type.

### Steps to reproduce

1. start a whatsapp-web.js session
2. call `POST /api/sessions/{sessionId}/messages/send-document` with a base64 file, MIME type and filename
3. check the attachment in WhatsApp

### Expected behavior

files sent through `send-document` should always arrive as WhatsApp documents.

### Relevant logs

```shell
the API accepts the request. there is no server error for this behavior.
```

i fixed it locally by changing `sendDocumentMessage` to pass the option that whatsapp-web.js already supports:

```ts
async sendDocumentMessage(chatId: string, media: MediaInput): Promise<MessageResult> {
  return this.sendMediaMessage(chatId, media, { sendMediaAsDocument: true });
}
```

the helper type also needs `sendMediaAsDocument?: boolean` next to `sendAudioAsVoice?: boolean`.

i built current `main` with this change, ran all 324 whatsapp-web.js adapter tests, and sent a `text/plain` file to my own account through the real REST endpoint. it returned HTTP 201 and arrived as a document.

cairodavila:
## summary

- force `sendDocumentMessage` to pass `sendMediaAsDocument: true` to whatsapp-web.js
- add a regression test for the forwarded option

## testing

- 324 tests passed in `whatsapp-web-js.adapter.spec.ts`
- a live `send-document` request arrived as a WhatsApp document

Fixes #989

### DIFF
--- a/src/engine/adapters/whatsapp-web-js.adapter.spec.ts
+++ b/src/engine/adapters/whatsapp-web-js.adapter.spec.ts
@@ -2635,6 +2635,32 @@ describe('outbound voice note (PTT)', () => {
   });
 });
 
+describe('outbound document mode', () => {
+  it('sendDocumentMessage forces whatsapp-web.js to preserve the document attachment type', async () => {
+    const sentMessage = { id: { _serialized: 'OUT1' }, timestamp: 1700000001 };
+    const sendMessage = jest.fn().mockResolvedValue(sentMessage);
+    const adapter = new WhatsAppWebJsAdapter({
+      sessionId: 's',
+      sessionDataPath: './data/sessions',
+      puppeteer: {},
+    });
+    (adapter as unknown as { status: EngineStatus }).status = EngineStatus.READY;
+    (adapter as unknown as { client: unknown }).client = { sendMessage };
+
+    await adapter.sendDocumentMessage('628@c.us', {
+      mimetype: 'application/pdf',
+      data: Buffer.from([1]).toString('base64'),
+      filename: 'document.pdf',
+    });
+
+    expect(sendMessage).toHaveBeenCalledWith(
+      '628@c.us',
+      expect.any(MessageMedia),
+      expect.objectContaining({ sendMediaAsDocument: true }),
+    );
+  });
+});
+
 describe('LID resolution for individual sends (#573 — WhatsApp @c.us → @lid migration)', () => {
   const ready = (client: unknown): WhatsAppWebJsAdapter => {
     const adapter = new WhatsAppWebJsAdapter({ sessionId: 's', sessionDataPath: './data/sessions', puppeteer: {} });
--- a/src/engine/adapters/whatsapp-web-js.adapter.ts
+++ b/src/engine/adapters/whatsapp-web-js.adapter.ts
@@ -1566,13 +1566,13 @@ export class WhatsAppWebJsAdapter extends EventEmitter implements IWhatsAppEngin
   }
 
   async sendDocumentMessage(chatId: string, media: MediaInput): Promise<MessageResult> {
-    return this.sendMediaMessage(chatId, media);
+    return this.sendMediaMessage(chatId, media, { sendMediaAsDocument: true });
   }
 
   private async sendMediaMessage(
     chatId: string,
     media: MediaInput,
-    extraOptions?: { sendAudioAsVoice?: boolean },
+    extraOptions?: { sendAudioAsVoice?: boolean; sendMediaAsDocument?: boolean },
   ): Promise<MessageResult> {
     this.ensureReady();
     this.ensureNotChannelRecipient(chatId);


rmyndharis (changes requested):
Thanks for the clean report in #989 and for following it up with a patch — the diagnosis is right, and `sendMediaAsDocument` is the correct lever. One blocker before this can go in, plus a note on the test.

## Blocker: the flag must not be set for `@broadcast` recipients

`whatsapp-web.js` refuses the send outright the moment `sendMediaAsDocument` is truthy and the recipient matches `/@\w*broadcast\b/` — and that regex covers **broadcast lists** (`<id>@broadcast`), not just `status@broadcast`:

```js
// node_modules/whatsapp-web.js/src/Client.js:1388
const isStatus = /@\w*broadcast\b/.test(chatId);
// :1413
} else if (isStatus && [options.sendMediaAsDocument, ...].includes(true)) {
    console.warn('The message type is currently not supported for sending in status broadcast…');
    return null;
}
```

Running the real method against an empty `this`, so only that guard can execute:

```
$ node -e "…Client.prototype.sendMessage.call({}, chatId, {}, { sendMediaAsDocument: true })…"
1234567890@broadcast   -> resolved: null
status@broadcast       -> resolved: null
628111@c.us            -> threw: Cannot read properties of undefined (reading …)
```

The `@c.us` case getting past the guard and failing later on the absent page state is what confirms the two `null`s come from the branch above rather than from the stub.

Our own guard does not cover this — `ensureNotChannelRecipient` is newsletter-only, and `wa-id.ts` says so explicitly ("Broadcast lists and `status@broadcast` resolve to a real `Chat` and are intentionally NOT matched here"). So the `null` flows into `toMessageResult`, which throws by design. The result for anyone sending a document to a broadcast list is a `201` today and an opaque `{"statusCode":500,"message":"Internal server error"}` after this change, plus a persisted FAILED row and a `message:failed` hook firing at their plugins.

That path is reachable from the shipped dashboard without a crafted request: `Chats.tsx` filters only `channel` and `status` out of the chat list, so a broadcast list stays clickable and its attachment picker defaults to `document`. `docs/04-security-design.md` also documents `status@broadcast` as an accepted `chatId`.

The fix is to withhold the flag for exactly those two kinds, using the helper we already have:

```ts
async sendDocumentMessage(chatId: string, media: MediaInput): Promise<MessageResult> {
  const kind = chatKind(chatId);
  const asDocument = kind !== 'status' && kind !== 'broadcast';
  return this.sendMediaMessage(chatId, media, asDocument ? { sendMediaAsDocument: true } : undefined);
}
```

`chatKind` is already imported in the adapter. This mirrors the shape `sendAudioMessage` uses for `sendAudioAsVoice`.

## The test does not exercise the defect

`application/pdf` already arrives as a document without the patch — WA Web classifies from the declared mimetype, and nothing about a PDF invites reclassification. The same applies to the `text/plain` file in your live check in #989: it would have arrived as a document either way, so that run does not actually demonstrate the bug.

I verified the real behaviour by sending from a whatsapp-web.js session on v0.11.1 and reading the message back from the receiving account:

| sent via `send-document` | arrives as | mimetype | filename |
| --- | --- | --- | --- |
| `image/png` | `image` | `image/jpeg` (re-encoded) | dropped |
| `application/pdf` | `document` | `application/pdf` | preserved |

So the payload is not merely mislabelled — an image is transcoded to JPEG and loses its filename.

Could you extend the test to cover:

1. an `image/*` mimetype asserting `sendMediaAsDocument: true` — the case that actually regresses;
2. `status@broadcast` and `1234567890@broadcast` asserting `expect.not.objectContaining({ sendMediaAsDocument: true })`.

The existing PTT pair just above yours (`sendAudioMessage with ptt` / `without ptt`) is the template.

## Not your problem — for context

While reviewing this I found some adjacent issues on the same path: URL sends discard the caller's `filename` and `mimetype`, and a document sent without a `filename` reaches the recipient labelled literally `undefined` once document mode is forced. I will handle those in a follow-up once this lands, so please keep this PR to the guard and the tests.

rmyndharis:
Thanks again for the report and the initial patch — the diagnosis was spot on. I've opened #996 to land this with the two follow-ups from my review folded in, and to rebase onto `main` (which has moved past this branch):

- the `sendMediaAsDocument` flag is withheld for `status@broadcast` and broadcast lists, since whatsapp-web.js returns `null` for any `@broadcast` recipient once it is set and `toMessageResult` turns that into a failed send;
- the test now uses an `image/png` mimetype (the case that actually regresses — a PDF already arrives as a document without the flag) and adds the two `@broadcast` cases.

Closing this in favour of #996. Attribution for the original one-line fix is preserved there.

rmyndharis:
Superseded by #996.

PR #1003, opened 2026-07-29
fix(engine): require real evidence before a session leaves ready or reports an unlink

rmyndharis:
## Summary

Three places in the session lifecycle were acting on something the gateway did not actually know. Each could take a working session down, or tell an operator something untrue. This corrects all three, plus the smaller inconsistencies that fell out of tracing them.

None of this has been released — everything here refines work still sitting under `[Unreleased]`, so the changelog entries are corrected in place rather than stacked with a second round of bullets describing the first.

## The onboarding-modal watcher could take a healthy session down

Detection keyed on the modal's heading text. An element's `textContent` includes all of its descendants, and WhatsApp Web renders every chat-list row as nested elements — so a conversation previewing the words *"what's new"*, an entirely ordinary English message, matched. A group named "What's New at Acme" matched. The session then moved to `action_required`, which `ensureReady()` refuses every send on, with a five-minute exposure window opening on every start and every reconnect.

The same pattern accepted only the ASCII apostrophe, so it could not match the typographic one WhatsApp Web actually renders. Over-inclusive against chat text and under-inclusive against the modal it was written for.

- Presence now keys on the **Continue button**, within a bounded number of levels of an element that also carries the heading. A control whose exact label is "Continue" sitting in that subtree is a shape the chat list does not produce; the walk is bounded so a match against `<body>` cannot qualify a button that is nowhere near the modal.
- The heading matcher accepts both apostrophes.
- A session leaves `ready` only after repeated clicks have failed to dismiss the modal — evidence the click is not landing and a human has to acknowledge it on the phone. A single click is the normal, successful case and moves nothing.
- A probe that cannot reach the page (reload, teardown, timeout) is logged and otherwise ignored. It says nothing about the modal, and blocking every send on a working session for a reason the operator cannot act on is worse than the problem being guarded against.

The matching moved into a self-contained exported function, so it is now executed directly against representative DOM shapes instead of only through a mocked `evaluate` — which exercised the loop but never the matching, the part that decides whether a session keeps working.

## An unlink could be reported without ever being sent

`engine.logout()` resolved when there was no live client or socket. The unlink never left the process, but the caller recorded a confirmed one and wrote a `SESSION_LOGGED_OUT` audit row for a device still listed under the account holder's Linked Devices.

The session-level "is it started?" check cannot catch this — it sees a registered engine, and an engine outlives its client: a stuck-auth recovery nulls it and then waits out the reconnect backoff, and a WhatsApp-side logout does the same on Baileys. Both engines now refuse, which surfaces as the 502 the route already documents. On Baileys that also preserves the credentials, which the success path would have wiped, leaving the device linked server-side with nothing left to retry with.

## `--best-effort` could wave a half-patched dependency through

`git apply` exits 128 for every fatal, and only some of those happen before it writes anything. Classifying the code alone as an untouched tree meant a fatal raised part-way through writing results — a read-only checkout, a full disk — was reported as a clean skip with exit 0, and a half-patched whatsapp-web.js reads as healthy from then on.

It is now matched on the refusals git emits while parsing, verified against the messages git actually produces for a CRLF patch (`corrupt patch at line N`) and a non-diff (`No valid patches in input`). The predicate is exported and tested from both directions.

## Smaller corrections

- **Stop and Unlink share one definition of "this session has a live engine."** A card can no longer offer an action the API rejects, nor offer Start to a session that is already started — which answers 400 and then leaves a QR dialog open that cannot resolve. `action_required` and `authenticating` previously had no working action at all; the latter is a state the whatsapp-web.js engine can hold for 90 seconds.
- **The unhandled-rejection handler passes its stack as log context**, not positionally. `warn`'s second parameter is the context, so a string there replaced the logger name for the whole line — burying a Puppeteer stack exactly where a reader looks for the scope.
- **The unlink error classifier separates the gateway's 502 from a reverse proxy's.** They mean opposite things: one stopped the session locally and wants a retry, the other never reached the gateway at all.
- **The `action_required` status pill uses the theme token** its siblings all use, instead of a literal that lost contrast on the dark card.
- **The Hebrew string for that status** said "regulation"/"installation" where every sibling string says session.

## Documentation

`action_required` is documented in the status table, the state diagram and both interface quotes. `lastError` is no longer described as failed-only. The 502 guidance now distinguishes a refused logout (nothing sent, nothing deleted, credentials survive) from one that ran past the teardown deadline — where the attempt is still running and, on whatsapp-web.js, ends in the same profile deletion a successful unlink performs, so that session does come back needing a QR. The send-document and remote-URL metadata changes get the changelog entries they were missing.

## Verification

Backend lint, `nest build`, full `tsc --noEmit` including specs, and the complete Jest suite — 3791 tests, 238 suites. Dashboard build, lint, typecheck and its 200 unit tests. The regenerated `openapi.json` is included.

The new coverage is deliberately aimed at what would have caught these: the in-page matcher executed against chat-preview text, a group name, both apostrophe forms, an offscreen button and an unrelated Continue button; the no-live-client unlink on both engines; the remote-URL metadata override in both directions; the exit-128 split; and `delete()` waiting on a teardown that lost its deadline race.


rmyndharis:
Follow-up: the remote-URL metadata change now excepts stickers.

`Util.formatToWebpSticker` dispatches on `mimetype.includes('image')`, but `formatImageToWebpSticker` then short-circuits:

```js
if (media.mimetype.includes('webp')) {
    return media;   // returned unconverted
}
```

So for a sticker the mimetype is an instruction, not a label — a URL sticker declared `image/webp` over bytes that are not webp was being sent unconverted. `toMessageMedia` grew a `trustDeclaredType` opt-out that only `sendStickerMessage` uses; the declared filename still wins everywhere, since nothing branches on it. Covered by a test that fails without the change.

PR #1049, opened 2026-08-02
Fix translation

albanobattistella:


rmyndharis (changes requested):
Thanks for keeping the Italian locale current — but this one changes the meaning, so I can't take it as-is.

The source string is "whatsapp-web.js uses your status-privacy **audience**" — `pubblico` is a noun here (the audience). The proposed "il tuo stato pubblico di privacy" turns it into an adjective attached to `stato`, dropping the audience concept and reading as "uses your public state of privacy" — which worse, implies the status is public. Every other locale keeps the audience noun.

The current string is indeed clunky (the double genitive "di privacy dello stato"), so a rewording is welcome — it just needs to keep `pubblico` as the audience. For example:

> Baileys richiede un elenco di destinatari; whatsapp-web.js utilizza il pubblico delle tue impostazioni sulla privacy dello stato.

or the minimal fix: `...utilizza il pubblico della privacy del tuo stato.`

rmyndharis:
One process ask alongside the string fix: please bring the PR itself in line with `CONTRIBUTING.md` and the PR template —

- **Title:** use a Conventional Commit, e.g. `fix(dashboard): keep the audience meaning in the it.json recipients hint`. The title becomes the commit message on squash merge, and "Fix translation" doesn't carry the scope or the why.
- **Body:** the description is empty. Please fill in the template (Description, Type of Change, Checklist) — even for a one-liner, the "what was wrong and why this wording" is what makes the change reviewable without opening the diff.

PR #1067, opened 2026-08-03
fix(scripts): prevent EALLOWSCRIPTS in nested dashboard install under npm 11

configurowebmax:
### Summary
Fixes a reproducible `EALLOWSCRIPTS` fatal error during the `postinstall` script execution under **npm 11** when `allow-scripts=true` is set in the user's `.npmrc`.

### Root Cause
1. When the root install (`npm ci`) reads `allow-scripts=true` from `.npmrc`, it exports this setting to the lifecycle environment as `npm_config_allow_scripts=true`.
2. Previously, `scripts/postinstall.js` executed `npm run dashboard:ci` (which runs `cd dashboard && npm ci`). That intermediate `npm run` process re-exported `.npmrc` settings to the nested `npm ci` command.
3. **npm 11** strictly forbids `--allow-scripts` originating from the CLI or environment variables (`npm_config_allow_scripts`) in project-scoped installs (`npm ci`, `npm i`), aborting the build with `EALLOWSCRIPTS`.

### Solution
- **`sanitizeEnv(env)`:** Added an environment sanitization helper in `scripts/postinstall.js` that strips any `npm_config_allow_scripts` variables (handling case and separator variations) from the inherited environment before spawning child processes.
- **Direct `npm ci` execution in `dashboard/`:** Replaced the indirect `'npm run dashboard:ci'` invocation with a direct `'npm ci'` call using `cwd: path.join(root, 'dashboard')` and the sanitized environment (`cleanEnv`), avoiding intermediate processes that re-export `.npmrc` options.
- **Project policy compliance:** Did NOT add any `"allowScripts": { "*": true }` entries to `package.json` nor modify user `.npmrc` configuration. Preserved fail-fast behavior for legitimate dashboard installation failures.

### Regression Tests
- Updated `scripts/postinstall.spec.js` assertions to match direct `'npm ci'` execution and explicit dashboard `cwd`.
- Added 3 dedicated unit regression tests for `sanitizeEnv`, environment stripping in `planSteps`, and `cleanEnv` injection in `run`.
- Verified cleanly against `npm run test:scripts` (20/20 passing), `npm run format:check`, and `npm run lint`.

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -4,11 +4,13 @@
  * Three conditional steps, each skipped when its target is absent so the hook is a no-op where the
  * piece is missing (the Docker builder stage copies package*.json long before any source):
  *
- *   1. `npm run dashboard:ci` when dashboard/ exists — the dashboard carries its own lockfile and
- *      the root install would otherwise leave it without dependencies. A failure here MUST abort
- *      the install: the old inline hook swallowed spawnSync's exit status, so a red dashboard
- *      install still reported `npm install` success and the breakage surfaced only at build/run
- *      time.
+ *   1. `npm ci` inside dashboard/ when dashboard/ exists — the dashboard carries its own lockfile and
+ *      the root install would otherwise leave it without dependencies. We run `npm ci` directly in
+ *      dashboard/ with any `npm_config_allow_scripts` environment variable stripped so that npm 11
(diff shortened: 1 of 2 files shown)


rmyndharis:
Thanks @configurowebmax — adopted in #1071 and merged to `main` as 5b14d406. Rebased onto the now five-step postinstall, and the env sanitisation was extended to the two whatsapp-web.js patcher steps added since your PR (locked by the env-strip test across all five steps). Credited in the CHANGELOG.

PR #1254, opened 2026-08-11
fix(engine): stop a whatsapp-web.js media burst shedding past the eighth

rmyndharis:
## Problem

This release removed the inbound-media queue cap on Baileys, because a cap equal to the active slots made admission a constant 8 whatever the batch size — one upsert of 40 lost the media of 32.

`whatsapp-web-js.adapter.ts` carried the identical construction and was not touched:

```ts
new ConcurrencyLimiter(
  inboundMediaConcurrency(),
  inboundMediaConcurrency(),   // queue cap == active slots
);
```

With the default concurrency of 4 that is the same active + queued = 8, so a burst on the whatsapp-web.js engine still emitted every message past the eighth without its media. The CHANGELOG entry describing the repair reads as engine-general; it was true of one engine.

## Why unbounding is safe here

It is not the same trade-off as on Baileys. There, `processInboundMessage` — the whole pipeline — runs inside the limiter, so parking delays every message including plain text. Here only `msg.downloadMedia()` is inside it, and each message awaits its own `capInboundMediaFor`, so a parked download delays that one message and a text message never enters the gate at all.

The slot is still held until the real download settles, so in-flight downloads remain bounded by `inboundMediaConcurrency()` — only the waiting room is unbounded.

## Also

With no queue cap the only rejection left is the limiter closing at teardown, so the warning no longer calls that saturation. Reporting an orderly teardown as saturation sends operators to look at concurrency settings for a problem that was never there — the same correction the Baileys side already got.

## Verification

`wwebjs-inbound-burst.spec.ts` pins the construction, mirroring `baileys-inbound-burst.spec.ts` including its anti-vacuity guard on the parser. Confirmed red before the change.

`lint`, `tsc --noEmit -p tsconfig.json`, `format:check` and `npm test -- --coverage` (5490 passing, no coverage threshold breached) all pass.


rmyndharis:
## Self-review correction — pushed as a second commit

Reviewing this branch again before merge, the change as written traded one defect for a worse one.

**The problem it introduced.** A slot is held until its download *really* settles — deliberately, so an abandoned download cannot let a fresh one exceed the concurrency bound — and a hung WA Web page never settles. `close()` is never called on this limiter (`webhook.service.ts:197` is the only `close()` in the codebase), and the limiter task swallows both download outcomes, so with the queue cap gone `run()` can no longer reject at all.

The rejected-on-full-queue path was doing real work: it unblocked everyone behind a stuck slot. Without it:

| | before this branch | after the first commit | after the fix |
|---|---|---|---|
| burst past the 8th | media lost, message emitted | media kept | media kept |
| burst behind stuck slots | message emitted without media | **never emitted at all** | emitted without media |

Losing the message is strictly worse than losing its media, and a hung page is exactly the condition this codebase has the most history with.

**The fix.** The inner timeout race only starts once a message is *admitted*, so it could never cover the queue wait. The caller's total wait is now bounded by the same `MEDIA_DOWNLOAD_TIMEOUT_MS` budget, which restores the degradation without shedding at a fixed batch size. The slot-holder `catch` is now a defensive log rather than a claim about a teardown nothing performs.

**Pinned three ways**, since the original spec asserted only the constructor and would have stayed green through all of this: the construction, the composition (a task that never gets a slot still resolves `null`), and the shape of the await. Reverting to `await boundedReady` turns the third red — verified by mutation, restored from a saved copy.

`lint`, `tsc --noEmit -p tsconfig.json`, `format:check` and `npm test -- --coverage` (5492 passing) all pass.

PR #1258, opened 2026-08-11
docs: correct four claims the code does not support

rmyndharis:
Round-2 review of this release found four statements that describe behaviour the repository does not have. None changes what runs; each would send a reader to look for something that is not there.

| Claim | Where | Reality |
|---|---|---|
| the tag path "publishes the chart" (×2) | `release.yml:90`, `:299` | no workflow packages or pushes the chart — no `helm package`, `helm push`, `chart-releaser` or `oci://` anywhere; operators install it from the ref |
| the service placement covers "the MCP tools" | `channel.service.ts`, `CHANGELOG` | true where it was copied from — `AgentToolDeps` injects `GroupService` — but there is no channel tool family, so nothing but the controller reaches `ChannelService` |
| the matrix covers "all five install-time patches" | `CHANGELOG` | there are **seven**: `postinstall.js` runs seven, the `Dockerfile` runs seven, and `docs/29` says seven in three places |
| a JSDoc block describing a 404 on an unknown call id | `whatsapp-web-js.adapter.ts` | `createCallLink` was inserted between `rejectCall`'s doc block and `rejectCall`, so the block documented the call-link builder |

Both `release.yml` sentences and the MCP justification were written in this session's own PRs (#1256 and #1257) — this corrects them rather than leaving them to be believed.

The channel guard's placement itself is unchanged and still right: it matches where the group guard sits and holds for any future non-HTTP caller. Only the borrowed justification was wrong.

`lint`, `tsc --noEmit -p tsconfig.json`, `format:check`, `actionlint` and the adapter suite (545 passing) all pass.


rmyndharis:
Added a fifth claim to this branch, found by the same pass.

The SDK entry for `chats/pin` / `chats/mute` said the gateway *"has published since v0.15.0 with no client counterpart"*. Neither route exists in `git show v0.15.0:openapi.json` — both were added in this same `[Unreleased]` section, so the routes and their five clients land together. The entry described a gap that never existed, which is the opposite of the situation `check:sdk-coverage` was written for.

PR #1306, opened 2026-08-12
fix(ci): excuse an unfixable advisory by id instead of lowering the audit bar

rmyndharis:
Unblocks the v0.18.0 release. `Security audit` went red on the release-cut commit (`19ad3f3a`) while the merge commit twenty minutes earlier was green — over an identical dependency tree. `GHSA-jmr9-qjv8-65gv` (`extract-zip` unvalidated symlink path traversal) had been published in between.

## Why it cannot simply be fixed

There is no patched version: the advisory covers `extract-zip *` and 2.0.1 is the latest publish. It reaches the tree through `@puppeteer/browsers <=2.13.2`, which `puppeteer-core@24.38.0` pins **exactly**, which `whatsapp-web.js@1.34.7` pins **exactly** in turn. Nothing expressible in `overrides` moves it, and `npm audit fix --force` proposes downgrading `whatsapp-web.js` to 1.34.2 — a downgrade of the default engine.

## The upgrade path was tried and measured

`overrides: { "@puppeteer/browsers": "^3.2.0" }` does remove the advisory (`extract-zip` leaves the tree entirely), and the functional risk I expected did not materialise — the Dockerfile's own `puppeteer browsers install 'chrome@146.0.7680.31'` succeeds against the 3.x API, and `puppeteer.launch()` drives a page on the result.

It fails elsewhere: 3.x replaces `extract-zip` with `yargs@18` and `modern-tar`, both `type: module`. Jest does not transform `node_modules`, so every suite reaching `whatsapp-web.js` dies at import — **60 suites fail to load, 5772 passing drops to 4152**. Reverting and reinstalling restored 5772 exactly, so the override was the cause. Making it work means changing `transformIgnorePatterns` for all 340 suites, which is not a release-time change.

## What this does instead

Keeps the `high` threshold and applies it **per advisory**. `npm run check:audit` reads `npm audit --json` and fails on any high/critical whose id is not explicitly allowlisted in `scripts/check-audit.mjs`, where each entry records its reason and what removes it.

The alternative lever — `--audit-level=critical` — would have un-gated every future high to accommodate one advisory, undoing what the audit job was extended to do in the previous release.

Two rules, and the second is what keeps it honest:

1. Any unallowlisted high/critical fails.
2. **An allowlist entry whose advisory no longer appears also fails**, so a carve-out cannot outlive its cause. When `whatsapp-web.js` ships a puppeteer carrying `@puppeteer/browsers` 3.x, this entry starts failing until it is deleted.

`dashboard/` keeps the plain `npm audit --audit-level=high` — it has nothing to excuse and stays the stricter of the two.

## Reachability, recorded in the allowlist entry

The vulnerable path is zip extraction inside `puppeteer browsers install`, which this repo calls once, at **image-build time**, on **amd64 only** (arm64 symlinks Debian's chromium and never reaches it), against a version-pinned Chrome for Testing build fetched from Google over HTTPS. Nothing in the shipped image extracts a zip through that path at runtime.

## Verification

Nine specs under `test:scripts`, including the load-bearing negative — the same advisory fails once it is not allowlisted — plus critical-blocks, moderate-does-not, and the stale-entry rule. Also mutation-checked end-to-end against the real audit output: emptying the allowlist makes `npm run check:audit` exit 1 naming the GHSA, and restoring it returns 0.

Full local set green: `check:audit`, `lint`, `format:check`, `tsc --noEmit`, `check:versions`, `openapi:check`, `check:chart`, `test:scripts` (91), `test` (5772), `build`.

Three documents described the old command and are corrected alongside it.


freethug-source (commented):
Great

PR #1328, opened 2026-08-16
fix(security): require an integrity pin for production plugin installs and document the plugin trust model

rmyndharis:
> Stacked on #1327 — retarget to `main` when it lands.

## Summary

- **Integrity pin required for production URL installs.** Installing a plugin is executing third-party code; HTTPS authenticates the channel, not the bytes-as-reviewed. `PLUGIN_INSTALL_REQUIRE_PIN` (production default, explicit opt-out documented in `.env.example`) makes a `#sha256=<hex>` pin mandatory on every URL install — plain `http` already required one, now in production so does `https`.
- **Plugin trust model documented in SECURITY.md.** The sandbox (worker isolate, capped heap, allowlisted env, deny-by-default network manifest, capability router) contains a buggy or runaway plugin — it is not a boundary against a malicious one. The section enumerates the install-path gates and points operators running untrusted fleets at a separate container with an OS-level sandbox.

## Tests

Five new cases pin the gate: unpinned https rejected in production, pinned accepted, lighter default outside production, `=true` enforces anywhere, `=false` lifts even in production. Malformed-marker sequencing preserved (post-download check keeps its error wording). Full lane, tsc, ESLint, prettier green.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -666,6 +666,9 @@ PLUGINS_DIR=./data/plugins          # Plugin directory (default: ./data/plugins)
 # Cap on a plugin .zip downloaded by install-from-URL (matches the 5 MB upload limit). A non-positive
 # or non-numeric value falls back to the default.
 # PLUGIN_DOWNLOAD_MAX_BYTES=5242880   # default 5 MiB
+# Require a #sha256=<hex> integrity pin on EVERY plugin install from a URL (production default;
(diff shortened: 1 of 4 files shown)


rmyndharis:
Retargeted to main after a history rewrite there — CI will run on the next push.

PR #1332, opened 2026-08-16
fix(config): forward every documented runtime knob through the bundled compose files

rmyndharis:
## What changed

The bundled `docker-compose.yml` and `docker-compose.dev.yml` `environment:` lists are an explicit allow-list: with no `env_file`, a variable missing from the list never reaches the container. About 75 knobs documented in `.env.example` had no line there, so setting them in `.env` silently did nothing. That included both v0.20.0 breaking-change opt-outs:

- `WEBHOOK_SSRF_REDIRECTS` (follow redirects with the SSRF guard off)
- `PLUGIN_INSTALL_REQUIRE_PIN=false` (lift the production install-pin requirement)

plus neighbors like `SSRF_ALLOWED_HOSTS`, `API_KEY_PEPPER`, `SHUTDOWN_DELAY_MS`, the WS/ingress limit families, and the DB/Redis pool and timeout knobs.

This PR forwards every documented knob the app reads, blank by default (`${KEY:-}`): a real host value still pins with top precedence, and the existing blank-clearing in `load-env` keeps `data/.env.generated` applying when nothing is set.

## The gate

The per-family and strict-boolean forwarding tests are replaced by one rule derived wholesale: every key `.env.example` documents that the app reads via `process.env` must be forwarded in **both** compose files, unless a `NOT_FORWARDED` entry names it with a reason (the dashboard-managed datastore toggles, `WEBHOOK_SSRF_PROTECT` which fails safe, `ALLOW_DEV_API_KEY` which production refuses, and the dev file's local-socket `DOCKER_HOST`). A new knob added to the app and `.env.example` now fails CI until it gets a forward or a documented decision.

## Consequential edits

- `BLANK_SHADOWED_ENV_KEYS` gains an entry for every new blank forward; `env-precedence.spec.ts` (which derives its expectation from the compose files) enforces this.
- Ten keys that shipped **uncommented** in `.env.example` with values equal to their defaults (`CORS_ORIGINS=*`, `WEBHOOK_SSRF_REDIRECTS=false`, the three DB timeouts, `REDIS_CONNECT_TIMEOUT_MS`, `BAILEYS_AUTH_DIR`, `MAX_CONCURRENT_SESSIONS`, `SEARCH_PROVIDER`, `SEARCH_LIMIT_MAX`) are now commented out. This is required by the existing rule that a blank-forwarded key must not ship uncommented (a copied `.env` would otherwise pin it), and it is behavior-neutral: each value equals the default the app uses when nothing is set.

## Verification

- Full unit suite: 332 suites, 5659 tests green (includes the rewritten parity spec and env-precedence).
- Mutation check: removing one forward line makes the new gate fail naming the exact key.
- `docker compose config` parses both files; `tsc --noEmit`, ESLint, and Prettier clean.


rmyndharis:

1. **PUPPETEER_EXECUTABLE_PATH no longer blank-forwarded.** A compose `environment:` entry overrides the image ENV even when it renders empty, so the blank forward deleted the Dockerfile's `/usr/local/bin/puppeteer-chrome` inside the container and puppeteer would have fallen back to a bundled Chromium the image does not ship (`PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true`). The forward now carries the image value as its default, and a new spec asserts the default stays in sync with the Dockerfile `ENV`.
2. **Ten more knobs forwarded** that the first pass missed because they are read through indirections rather than `process.env.KEY`: `METRICS_TOKEN`, `MESSAGE_REAPER_GRACE_MS`/`_BATCH_SIZE`, `INGRESS_RECONCILE_GRACE_MS`/`_BATCH_SIZE`/`_MAX_ATTEMPTS`, `MCP_RATE_LIMIT_MAX`/`_WINDOW_MS`, `MCP_IP_RATE_LIMIT_MAX`/`_WINDOW_MS`.
3. **The gate's read detection now covers those idioms** (`env.KEY`, `env['KEY']`, `xyzEnv('KEY')` helpers, `config.get('KEY')`), which also re-arms enforcement for the previously text-scanned families (`SEND_PACING_*`, `WS_RATE_LIMIT_*`, `STORAGE_*`, ...): 33 documented read keys were invisible to the first walk.

Full unit suite, `docker compose config` (both files), tsc, ESLint, Prettier re-verified.

PR #1333, opened 2026-08-16
fix(queue): hash ingress job ids BullMQ refuses instead of degrading to inline dispatch

rmyndharis:
## What changed

`IngressEnqueueService.enqueue` now passes the delivery id through `sanitizeIngressJobId` before `queue.add`.

BullMQ refuses two jobId shapes at `Job.validateOptions`:
- an exact integer string (`Custom Id cannot be integers`), which is what a provider with a numeric dedup header sends, and
- a colon id that does not split into exactly 3 parts (`Custom Id cannot contain :`), which is exactly what the redrive path mints (`redrive:<uuid>`).

Because `enqueue()`'s catch-all treats any `add()` throw as "Redis unreachable", those deliveries silently degraded to inline dispatch: no retry, no backoff, no queue ordering, and the admin redrive endpoint blocked up to the 5s dispatch timeout per row.

## Design

The refused shapes are hashed to a deterministic `ing-<sha256[:40]>` prefix. BullMQ's exactly-once enqueue dedup only needs the id to be stable per delivery, not recognizable, and the reconciler replays through this same function, so its dedup against an earlier successful enqueue is preserved. Safe ids (including 3-part colon ids, which BullMQ accepts) pass through unchanged.

## Verification

- New spec cases: pass-through for safe ids, deterministic hashing for the refused shapes, and three enqueue cases (numeric header, redrive uuid, two-part colon id) asserting `queued` with the sanitized id rather than the inline fallback.
- Full unit suite green (332 suites); `tsc --noEmit`, ESLint, Prettier clean.


rmyndharis:

1. **The hash is now namespaced by plugin/instance.** BullMQ dedups custom jobIds across the whole shared `ingress-queue`, while the database dedup is `(pluginId, instanceId, providerDeliveryId)`. Numeric provider ids are exactly the short per-account sequence style two instances of one provider can share, so a queue-global hash would have made the second instance's delivery collide with the first's job id and BullMQ would silently discard it (resolved as the existing job, no DLQ row). The hash input is now `pluginId \0 instanceId \0 jobId`; the reconciler replays through the same function so replay idempotency is unchanged.
2. **`0:`-led ids covered.** `Queue.addJob` (BullMQ 6.1.1) additionally rejects `'0'` and any id starting with `'0:'`, which the first version's predicate missed (`0:tenant:123` is a 3-part colon id, so it passed through raw and still threw into the inline-dispatch catch-all). Both shapes are hashed now, and non-string ids (a duplicated dedup header surfaces as `string[]`) are coerced instead of reaching BullMQ's own `startsWith` TypeError.

New spec cases pin the namespace divergence, the `0`/`0:` shapes, and the coercion. Full unit suite, tsc, ESLint, Prettier re-verified.

PR #1335, opened 2026-08-16
fix(security): tighten the SQLite database files to owner-only on boot

rmyndharis:
## What changed

`better-sqlite3` creates its database files with `0666 & umask`, so `data/openwa.sqlite` and `data/main.sqlite` ended up `0644` while every sibling secret in `data/` had been tightened to `0600`/`0700` (`.api-key`, `.env.generated`, the Baileys auth dirs). The databases are secret stores in their own right: `webhook.secret`, `plugin_instances.secret` (host-minted ingress HMAC keys) and `sessions.proxyUrl` are stored in plaintext inside them.

The bundled container deployment is insulated (named volume, dedicated non-root user), but bare-metal `npm run start:prod` and bind-mount setups shipped world-readable secrets.

`SqlitePermissionsBoot` is an `OnApplicationBootstrap` provider (runs after the TypeORM modules' init, which is where the files are created) that chmods each SQLite database plus its `-wal`/`-shm`/`-journal` sidecars to `0600`:

- the main connection is always SQLite;
- the data connection only when its configured type resolves to sqlite (Postgres deployments have no local file);
- best-effort per file: a failure (odd mount, immutable file) logs a warning and never blocks boot;
- re-applied on every start, so a file recreated by a restore or an external copy converges back, matching the credential-dir re-tighten posture.

## Verification

- New spec: tightening an existing file and all three sidecars; skipping absent paths; the best-effort path (forced `chmodSync` failure warns and does not throw); the boot hook resolving both files for sqlite, skipping the data file for postgres, and falling back to the documented default paths.


rmyndharis:

- Corrected the ordering rationale in the docblocks: the DataSources initialize eagerly inside their provider factories (not in an `onModuleInit`, which @nestjs/typeorm 11 does not define), and `onApplicationBootstrap` fires after every init hook either way. The guarantee holds a fortiori.
- The data-connection branch now mirrors the app.module factory's own condition (`!== 'postgres`) instead of `=== 'sqlite'`; equivalent today only because env validation confines `DATABASE_TYPE` to those two values, so the boot is faithful by construction rather than by validator coincidence.
- `restore.sh` chmods the two restored sqlite files to 0600 (matching what the app re-tightens on boot and what it already did for `.env.generated`/`.api-key`), closing the restore-to-next-boot window for files extracted under a foreign umask.

Full unit suite, tsc, ESLint, Prettier, `bash -n` re-verified.

PR #1336, opened 2026-08-16
fix(security): warn once when the throttler collapses onto a single proxy bucket

rmyndharis:
## What changed

`ProxyAwareThrottlerGuard` keys rate-limit buckets on `resolveClientIp(req, TRUSTED_PROXIES)`. With an empty `TRUSTED_PROXIES` that resolves to the direct socket address, which behind the documented reverse proxy is the proxy for every client: all traffic shares one bucket per tier (a single abuser rate-limits everyone), and per-key `allowedIps` see the proxy for every caller.

Refusing to trust the spoofable `X-Forwarded-For` header is the correct default, but the consequence was silent. The guard now logs a single warning per process at the first request that carries the header while the list is empty, naming the fix:

```
X-Forwarded-For is present but TRUSTED_PROXIES is empty: every client shares one
rate-limit bucket keyed on the proxy address (and per-key IP allowlists see the proxy
too). Set TRUSTED_PROXIES to the proxy address/subnet to key limits per client.
```

The nginx FAQ recipe (docs/12, the documented proxy setup path) previously set `X-Forwarded-For` with no `TRUSTED_PROXIES` guidance at all; it now tells operators to set it (`127.0.0.1` for host-level nginx, the compose subnet for a network peer).

## Verification

- New spec cases: the warning fires exactly once across requests on the guard instance, stays silent when the proxy is trusted, and stays silent for direct traffic without the header.


rmyndharis:

- The once-only comment now states the real scope: once per guard instance, which is the singleton APP_GUARD in practice (the only other instance, `InstanceThrottlerGuard`, reaches the warning solely through a defensive `super.getTracker` fallback that is unreachable on the real ingress route).
- Spec hardening: an empty-string `X-Forwarded-For` header (a stripped header still signals a proxy hop) now has a warning case, and the pre-existing describe silences the one stray console line its shared instance emits.
- docs/12 guidance corrected: for the bundled compose the published port traverses Docker NAT, so the container sees the bridge gateway, not `127.0.0.1`; the subnet example is now the canonical advice, with `127.0.0.1` reserved for a bare-metal nginx talking to the process directly.

Full unit suite, tsc, ESLint, Prettier re-verified.

PR #1338, opened 2026-08-16
perf(database): index messages.mediaPath for the chat-media orphan sweep

rmyndharis:
## What changed

The chat-media orphan sweep lists archived files and reconciles them against the messages table with a per-chunk `WHERE mediaPath IN (...)` query (`chat-media-archive.service.ts`), whose own comment already described it as "an indexed lookup over a bounded key list". No index existed on `mediaPath`: each hourly chunk was a full messages-table scan, so the sweep's cost grew with the whole message history instead of the archived set (archiving is opt-in and off by default).

This adds `IDX_messages_mediaPath` as a partial index (`WHERE "mediaPath" IS NOT NULL`):

- the column is NULL for every row while `CHAT_MEDIA_ARCHIVE_ENABLED` is off (the default), for non-media messages, and for over-cap media, so the partial form indexes only rows that can ever match;
- a hand-authored migration (`1786100000000`) creates it for migration-managed deployments, with the entity `@Index` carrying the same explicit name so the `synchronize` path converges on the identical index (same arrangement as `IDX_messages_createdAt`);
- `CREATE INDEX IF NOT EXISTS` makes `up` idempotent on both SQLite and Postgres; `down` drops it.

## Verification

- Migration spec (in-memory SQLite): index created, `up` re-run is a no-op, `down` drops it, and `EXPLAIN QUERY PLAN` for the sweep's `IN (...)` predicate shows a SEARCH via the index with no table SCAN.


rmyndharis:

- **The migration now lifts `statement_timeout` on Postgres** (`SET LOCAL statement_timeout = 0` behind a type guard), matching `AddMessagesCreatedAtIndex`. Without it, an upgrade boot where this is the only pending migration runs `CREATE INDEX` inside its own transaction with the pool's 30 s default `statement_timeout` in effect (earlier migrations' `SET LOCAL` no longer applies), so a large `messages` table gets the statement cancelled, the ledger-advancing transaction aborts, and the factory retry loop crash-loops the boot. Fresh-DB boots were masked by the single all-migrations transaction; CI cannot see the single-pending-migration path.
- docs/05's migration inventory gained the new file (the tree's tail entry was previously `1786000000000`).

Suite re-verified green.

PR #1340, opened 2026-08-16
fix(sdk): encode empty map fields as JSON objects in the PHP client

rmyndharis:
## What changed

PHP serializes an empty array as a JSON list, so two map-typed request fields reached the gateway as `[]` and were rejected by its object validation (`@IsObject` excludes arrays; the header-map validator rejects arrays outright):

- `headers` on webhook create/update (clearing custom headers, or defaulting them empty)
- `vars` on send-template (a template with no variables)

The three resource methods now cast an empty map to `stdClass` at the boundary where the field's wire type is known, so it encodes as `{}`. Non-empty maps are unaffected, and genuinely list-typed fields are untouched (the cast is per-field, not a body-wide conversion).

`MockBackend` gained a `rawBody()` accessor because its decoded view collapses both `{}` and `[]` to the same empty PHP array; the new test pins the exact encoded bodies.

## Verification

`phpunit` green (33 tests, 105 assertions in ResourcesTest, full suite run); `php -l` clean on all touched files.


rmyndharis:

- **`sendBulk` now casts empty per-item `variables` maps too**: `BulkMessageItemDto.variables` carries the same `@IsOptional() @IsObject()` validation as template `vars`, so an empty PHP array inside any bulk item had the identical `[]`-vs-`{}` failure the PR was opened for. New test pins both an empty and a non-empty item in one raw body.
- `MockBackend::lastCall()`'s docblock (displaced by the `rawBody()` insertion) is restored, and `rawBody()` carries a proper `@return string`.

Full PHP suite green (82 tests, 256 assertions).

PR #1342, opened 2026-08-16
fix(ci): run the contract-shape and docs gates on the release path

rmyndharis:
## What changed

A tag push starts `release.yml` alone (`ci.yml` triggers on branches), so any gate the tag path skips is a gate a release never ran. The workflow's own header states the invariant ("Mirror the CI workflow's gate jobs so a tag can never publish something the branch gate would have refused"; "the release gate must not be laxer than the PR gate"), but two lane gates had drifted to branches only:

- `check:contract-shapes` — the typed-SDK wire-shape check: an SDK whose mapped response types drifted from the OpenAPI contract could ride a tag to publication.
- `test:docs` — the repo-file drift specs (docs trees, gate tables, env examples).

Both now run on the tag path (`check:contract-shapes` at the end of the release lint job, `test:docs` in the release test job next to `test:scripts`).

To keep the invariant from drifting again, a new spec (`src/common/release-gate-parity.spec.ts`) derives the `npm`/`npx` gate commands of ci.yml's `lint` and `test` jobs and requires every one of them to appear in release.yml's counterparts. Adding the next lane gate to CI now fails the suite until the release path runs it too. The release jobs may run more (its lint also carries the audit gate); only the subset direction is asserted.

## Verification

- The parity spec passes (18 tests in the file group) and fails when either workflow gate is removed (the subset assertion enumerates the exact missing command).
- Full unit suite green; workflow YAML parses; `tsc --noEmit`, ESLint, Prettier clean.


rmyndharis:

- **The parity spec is now registered in the docs lane** (`test:docs` positional list) and mirrored in `testPathIgnorePatterns`, per the repo's lane-partition policy; it previously fell through both lists and ran in the unit lane under coverage. Lane totals confirm the move: docs 23 suites/259 tests (+1), unit 332 (−1); `test-lane-partition.spec.ts` green.
- **`executableLines` moved to a non-spec helper** (`src/common/workflow-lines.ts`) imported by both spec files, so importing it no longer drags the privilege-gate suite's describes into the importing suite's run.
- Spec doc comment now attributes the quoted invariant to where it actually lives (the workflow header plus the check:audit step comment).

PR #1389, opened 2026-08-17
fix(baileys): add resilient WhatsApp Web version resolution and fallback cache

giovanni-orciuolo:
### Summary

Fixes WhatsApp Web connection drops caused by upstream `fetchLatestBaileysVersion()` failures (such as GitHub raw content outages or rate limits).

### Changes

1. **`BaileysVersionResolver` (SRP Module)**:
   - Introduced a dedicated class `BaileysVersionResolver` (`src/engine/adapters/baileys-version-resolver.ts`) that manages WhatsApp Web version resolution with a 5-tier fallback strategy:
     - **Tier 1 (Operator Override)**: `BAILEYS_WA_VERSION` environment variable (e.g. `2.3000.1045340097` or `2,3000,1045340097`).
     - **Tier 2 (Direct WA Web Endpoint)**: Calls `b.fetchLatestWaWebVersion()` (`https://web.whatsapp.com/sw.js`) directly without any GitHub dependency. Persists successful versions to local disk cache.
     - **Tier 3 (Upstream Defaults)**: Calls `b.fetchLatestBaileysVersion()`. Persists successful versions to local disk cache.
     - **Tier 4 (Disk Cache)**: Reads `<authDir>/last_known_wa_version.json` if remote endpoints fail.
     - **Tier 5 (Safe Modern Fallback)**: Returns modern known working version (`[2, 3000, 1045340097]`).
2. **Lifecycle Integration**:
   - `BaileysLifecycle.connectInner` delegates version resolution cleanly to `BaileysVersionResolver`.
3. **Unit Tests**:
   - Added comprehensive tests in `src/engine/adapters/baileys-version-resolver.spec.ts` testing each tier using the AAA pattern.

### Verification

- Ran `npm test -- src/engine/adapters/baileys.adapter.spec.ts src/engine/adapters/baileys-version-resolver.spec.ts` (350/350 tests passed).


rmyndharis (changes requested):
Thanks for this, and for the shape of it: a dedicated resolver class, a documented tier order, and a test per tier is how I would want this structured. The problem is not the design, it is that the library does not behave the way the chain assumes, so most of it cannot run. Details below with the checks, so you can confirm each one yourself.

## Blocking: tiers 3, 4 and 5 are unreachable

Neither fetcher ever rejects. Both swallow every error and **resolve** with Baileys' frozen constant:

```js
// node_modules/@whiskeysockets/baileys/lib/Utils/generics.js  (both functions)
catch (error) {
    return { version: baileysVersion, isLatest: false, error };
}
// lib/Defaults/index.js
const version = [2, 3000, 1043857760];
```

`resolve()` only checks `isValidVersion(result?.version)` and ignores `isLatest`, so that stub always satisfies tier 2. `fetchLatestBaileysVersion` is never called, the disk cache is never read, and `DEFAULT_FALLBACK_WA_VERSION` is never returned.

Reproduced against the installed `@whiskeysockets/baileys@7.0.0-rc14` using the resolver from this branch, seeding the cache with a good value first, across three outage shapes (DNS failure, captive-portal HTML 200, HTTP 503). Identical every run:

```
seeded cache      : [2,3000,1045340097]
resolved (outage) : [2,3000,1043857760]   <- Baileys' frozen constant
tier 3 calls      : 0
cache after       : [2,3000,1043857760]   <- good value destroyed
```

Two consequences. The chain silently returns a stale protocol version with no log line, so an operator gets no signal that anything degraded. And `saveCachedVersion` writes that stub over a previously good cached value, which poisons the cache during exactly the outage tier 4 exists for.

Worth noting the same applies one level down: because `fetchLatestBaileysVersion` also resolves with the constant, checking `isLatest` on tier 2 alone would just move the dead end to tier 3. Both tiers need the check.

## Blocking: the fallback is still unreachable when an outage hangs

Neither call passes a `signal`, so each `fetch` runs to undici's default (roughly 300s) while the session init deadline is 60s (`resolveEngineInitTimeoutMs`, `src/engine/engine-init-timeout.ts`). If an endpoint hangs rather than failing fast, init times out before the chain advances, so the cache and default tiers are unreachable in that case too. This PR also adds one more network call ahead of the existing one, widening that window.

The two helpers differ in how much you can do about this, which is easy to miss:

```js
// fetchLatestWaWebVersion: spreads everything, so signal and dispatcher both reach fetch
const response = await fetch('https://web.whatsapp.com/sw.js', { ...options, method: 'GET', headers });

// fetchLatestBaileysVersion: only these three keys, so a signal you pass is silently dropped
const response = await fetch(URL, { dispatcher: options.dispatcher, method: 'GET', headers: options.headers });
```

So tier 2 can be bounded with `AbortSignal.timeout(...)`, but tier 3 cannot be bounded through options at all and needs an outer race or a dispatcher-level timeout.

## Please also address

**The `web.whatsapp.com` fetch bypasses the session proxy.** `connectInner` builds `proxyAgent` and passes it as both `agent` and `fetchAgent` specifically so a session with `proxyUrl` never connects direct (`baileys-lifecycle.ts:173`, "instead of silently connecting direct (#859)"). The resolver runs before that and calls `fetchLatestWaWebVersion()` with no options, so it egresses on the host's real IP on every connect and reconnect. Since that helper spreads `...options`, a dispatcher threads straight through.

**An existing reconnect test is disarmed.** The mock change in `test/__mocks__/@whiskeysockets/baileys.ts` makes a regression test in `baileys.adapter.spec.ts` pass vacuously. Please confirm it still fails when the behavior it names is broken.

**`BAILEYS_WA_VERSION` is not registered.** A new knob needs entries in `.env.example`, both compose files, `BLANK_SHADOWED_ENV_KEYS` (`src/config/env-precedence.ts`) and `src/config/env.validation.ts`. All twelve other engine knobs are registered, and without this the knob cannot be set at all under the bundled Docker stack. The env gates do not catch it because they derive their required set from `.env.example`.

**CI Lint is red**: ten errors across four files (Prettier, three unnecessary type assertions, one unsafe `any`). Nine clear with `npm run lint -- --fix` and `npm run format`.

## Smaller points

- The tier tests drive tiers 3 to 5 with `mockRejectedValue`, a shape the library provably never produces. Once `isLatest` is handled, please assert against the real resolve-with-stub shape, otherwise the suite stays green over the production behavior.
- The comma-format test passes even with comma support removed, so it is not measuring anything yet.
- Malformed values behave inconsistently: `2.3000` falls through quietly, `2.3000.abc` logs "Invalid" and continues, and `1.2.3` is accepted as a real override. Worth deciding whether an obviously wrong pin should fail loudly instead.
- The spec creates and recursively removes a directory inside the tracked `test/` tree. Please use `os.tmpdir()`.
- `DEFAULT_FALLBACK_WA_VERSION` will age. Consider documenting when it should be refreshed, or deriving it from the cache rather than a constant.

The underlying problem is real and the resolver is the right place to fix it, so I would like to land this once the chain can actually reach its lower tiers. Happy to help with the `isLatest` handling or the deadline wiring if you want a hand.

giovanni-orciuolo:
Thank you for the thorough review and insightful breakdown of the Baileys library internals!

I have addressed all the points in commit `e97ff9ac`:

1. **`isLatest: true` Check & Cache Protection**:
   - Both `resolveFromWaWeb` and `resolveFromBaileys` now explicitly check `result?.isLatest === true && this.isValidVersion(result?.version)`.
   - When `isLatest: false` (or an error) is returned, the resolver logs a warning, skips overwriting the disk cache with the frozen stub, and advances to the next tier.

2. **Timeout Bounds**:
   - Added bounded timeouts (defaulting to 5000ms):
     - `resolveFromWaWeb`: passes `signal: AbortSignal.timeout(this.timeoutMs)`.
     - `resolveFromBaileys`: raced via `Promise.race` against a timer with proper cleanup in `.finally()`.

3. **Proxy Egress Integration**:
   - `BaileysLifecycle.connectInner` now forwards `dispatcher: proxyAgent` to `this.versionResolver.resolve(b, { dispatcher: proxyAgent })` so remote version resolution does not bypass the configured session proxy.

4. **Registered `BAILEYS_WA_VERSION` across Configuration**:
   - Registered in `src/config/env-precedence.ts` (`BLANK_SHADOWED_ENV_KEYS`).
   - Registered in `docker-compose.yml` and `docker-compose.dev.yml` (`- BAILEYS_WA_VERSION=${BAILEYS_WA_VERSION:-}`).
   - Documented in `.env.example` under the Baileys engine section.
   - Strictly validated in `src/config/env.validation.ts` (`major === 2`, `minor >= 2000`, `patch >= 0`).
   - Validated against `docs-env-example.spec.ts` and `env-precedence.spec.ts`.

5. **Test Suite & Mock Cleanup**:
   - Reverted the `Browsers` mock in `test/__mocks__/@whiskeysockets/baileys.ts`.
   - Refactored `src/engine/adapters/baileys-version-resolver.spec.ts` to use `os.tmpdir()`.
   - Tier tests now assert against the real library behavior (`{ version: [2, 3000, 1043857760], isLatest: false }`), timeout races, proxy dispatcher forwarding, and comma/dot parsing.

6. **Documentation & Code Hygiene**:
   - Added JSDoc to `DEFAULT_FALLBACK_WA_VERSION` explaining semantic format and when to refresh.
   - Ran `npm run lint -- --fix` and `npm run format` (all checks passing, 0 errors / 0 warnings).

rmyndharis:
Verified all six points against the branch rather than the summary, and every one holds:

- `isLatest === true` is checked on **both** remote tiers, and `saveCachedVersion` now sits inside the success branch only, so a stale answer is neither returned nor written over a good cached one.
- The two tiers use different timeout mechanisms, which is the right call: `fetchLatestWaWebVersion` spreads `...options` so `AbortSignal.timeout` reaches fetch, while `fetchLatestBaileysVersion` maps only three keys and drops a signal, so racing it is the only option that works.
- `resolve(b, { dispatcher: proxyAgent })` is now called after `proxyAgent` is built, so the ordering is right and not just the argument.
- All five registration sites carry `BAILEYS_WA_VERSION`, and the validator rejects both a malformed string and an implausible one.
- The mock revert is clean: the diff against the base for `test/__mocks__` is empty, so the reconnect regression test is armed again.
- The specs assert the shape the library really produces. Local gate is green here too: lint, `tsc --noEmit`, `format:check`, 263 docs specs and the full unit suite.

I pushed two small follow-ups as a separate commit (9bc3f26) rather than asking for another round, since they are mechanical and the change is otherwise ready. Your commits are untouched.

**Abandoned fetch.** `Promise.race` drops the loser without cancelling it, so the timed-out call keeps running and its settlement is unobserved. It cannot reject today because it catches everything and resolves the frozen stub, so this is defence against the library changing rather than a live bug. `fetchPromise.catch(() => undefined)` defuses it, and the timer is `unref`'d for the same reason `withQueryDeadline` does it in `baileys-query-deadline.ts`, which is the existing pattern for this shape in the repo.

**CHANGELOG.** The repo asks for an `[Unreleased]` entry on user-visible changes, and this one had none. I described the chain from the operator's side rather than by tier number, since that is what a reader upgrading cares about.

One note for when this merges: it will be a merge commit, not a squash, so both of your commits keep their authorship in the history.

Thanks for the careful second pass. Reading the two fetchers' option handling closely enough to notice that only one forwards a signal is exactly what this needed, and the result is a chain whose lower tiers can actually be reached.

rmyndharis:
Heads up on a force-push: the branch is now rebased onto current `main`.

Adding the CHANGELOG entry is what made this necessary. The branch was 35 commits behind and had not touched that file, so nothing conflicted until an entry landed in the same `[Unreleased]` section that `main` had meanwhile filled. Rebasing was the cleaner fix, since it also clears the drift rather than papering over one file.

Your two commits keep their authorship and their content; only the parent changed, so the SHAs moved:

| before | after |
| --- | --- |
| `6f87fe3` | `1d6d1fe` |
| `e97ff9a` | `5b68034` |

If you have the branch checked out locally, `git fetch` then `git reset --hard origin/fix/baileys-version-resolution` will line it up. Nothing of yours was dropped or edited.

`mergeable` is back to true and CI is running.

rmyndharis (approved):
All eleven points from the first review are addressed, verified against the branch rather than the summary, and CI is green on the rebased head.

The chain now behaves as intended: `isLatest` gates both remote tiers so a stale answer is neither returned nor cached over a good one, each tier is bounded by the mechanism its own fetcher supports, the resolver rides the session proxy, and the knob is registered everywhere the repo requires with a validator that rejects an implausible pin.

Thanks for the second pass, and for reading the two fetchers closely enough to notice that only one of them forwards a signal. That detail is what made the timeouts correct rather than merely present.

PR #1396, opened 2026-08-18
fix(baileys): read the WA web version from WhatsApp, not the Baileys repo

m7fz7:
## Description

The Baileys engine resolved its protocol version with `fetchLatestBaileysVersion()`, which reads a version hardcoded in the Baileys GitHub repository. That pin lags what WhatsApp's servers actually accept, and a stale version fails the handshake silently: pairing links (and, per upstream reports, some ordinary connections) never complete and surface no reason.

This swaps it for `fetchLatestWaWebVersion()`, which reads the version straight out of `web.whatsapp.com/sw.js`, so it always matches what WhatsApp is currently serving. Both functions are exported by the pinned `@whiskeysockets/baileys@7.0.0-rc14` (`lib/Utils/generics.d.ts:54`) and return the same `{ version }` shape, so the call site is a drop-in replacement.

Scope is one line of engine code plus the test doubles that named the old function:

- `src/engine/adapters/baileys-lifecycle.ts`: the call, and a stale comment referring to it by name
- `test/__mocks__/@whiskeysockets/baileys.ts`, `test/baileys-engine.e2e-spec.ts`, `src/engine/adapters/baileys.adapter.spec.ts`: mocks renamed to match

Behaviour is otherwise unchanged: the reconnect path still treats a failed version lookup as non-terminal, so a registry outage retries rather than killing the session.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] Tests added/updated
- [x] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

Existing specs were updated rather than extended: the version source is a single library call the adapter spec already exercises through its mock, so there is no new assertion that meaningfully locks it down. Documentation is the `CHANGELOG.md` `[Unreleased]` entry; no file under `docs/` names either function.

Verified locally with `npm run build`, `npm test`, `npm run lint`, `npm run format`, and `npm --prefix dashboard run build`.

Not applicable, no user-facing surface changes.

## Related Issues

No existing issue tracks this specific cause.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- The Baileys engine reads the WhatsApp Web protocol version from `web.whatsapp.com/sw.js` instead of the version pinned in the Baileys repository. The pinned value lags what WhatsApp's servers accept, and a stale version fails the handshake silently, so pairing links (and some ordinary connections) never completed and gave no reason.
 - The Go and Java SDKs can @mention on an audio send. `SendAudioRequest` was flattened off the shared media type and lost the field, so the typed path could not set it while every other client could.
 - Three routes declare the `409` they can answer: a duplicate template name on create or rename, and an integration instance id that already exists. Clients generated from the contract modelled those calls as unable to conflict.
 - `DOMAIN` is dropped from `.env.example`. Nothing read it, so an operator setting it to their real hostname changed nothing.
--- a/src/engine/adapters/baileys-lifecycle.ts
+++ b/src/engine/adapters/baileys-lifecycle.ts
@@ -180,7 +180,12 @@ export class BaileysLifecycle {
     }
     const b = await this.loadLib();
     const { state, saveCreds } = await b.useMultiFileAuthState(this.host.authPath);
-    const { version } = await b.fetchLatestBaileysVersion();
+    // fetchLatestBaileysVersion() scrapes a hardcoded version out of the Baileys GitHub repo, which
+    // lags what WhatsApp's own servers actually expect — the confirmed root cause of pairing links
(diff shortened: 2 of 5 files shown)


rmyndharis:
Thanks for this, and for the clear write-up.

Your diagnosis is right. On the pinned `7.0.0-rc14`, `fetchLatestBaileysVersion()` currently
resolves `[2, 3000, 1043857760]`, which is byte-identical to the constant already bundled in the
package, while WhatsApp is serving `1045443729`. The round-trip was buying nothing, and the pin
has gone as long as 132 days between updates upstream.

I am closing this on timing rather than substance. #1389 was opened a day earlier against the same
call site and merged a few hours ago. It resolves the version through a fallback chain: an operator
pin (`BAILEYS_WA_VERSION`), then the two library endpoints, then a disk cache of the last known-good
value, then a built-in default. Each remote tier is bounded by a timeout and rides the session proxy,
and a stale answer is neither cached nor used.

Three things shaped that design, in case they are useful to you:

- Neither library function ever rejects. Both swallow every error and resolve with the bundled
  constant and `isLatest: false`, so a blocked or rate-limited request degrades silently to the very
  version the fix exists to avoid. Checking `isLatest` is what makes any fallback reachable.
- `web.whatsapp.com/sw.js` is not stable per request. Ten consecutive reads here returned three
  different `client_revision` values, non-monotonically, so it reflects a staged rollout bucket
  rather than one current version. The version is re-resolved on every connect and reconnect, so
  caching matters more than it first appears.
- `connectInner()` builds the egress proxy agent immediately above this call, and the request needs
  to ride it. Otherwise a session configured with a proxy reaches WhatsApp from the host address.

None of that takes anything away from your patch. It is a correct fix for a real problem that
nothing in this repository had reported, and you found it independently. Thank you for the care you
put into the description, and please do send more.

PR #1421, opened 2026-08-20
fix(dashboard): let the chat room shrink inside its flex row

rainerigius:
## Description

Fix a chat layout overflow issue where the chat area could extend beyond its parent container, causing horizontal overflow on narrower layouts.

The issue was in `.chats-page .chats-room`. Since it is a flex item, its default minimum width could be determined by its content, preventing it from shrinking enough to stay within the available space.

The fix adds:
```css
min-width: 0;
```
to .chats-page .chats-room, allowing the flex item to shrink correctly inside its container and preventing the chat layout from overflowing horizontally.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [ ] Tests added/updated
- [ ] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- The dashboard chat room shrinks within its layout instead of overflowing it, so a long contact or group name no longer pushes the send button out of reach. Thanks @rainerigius.
+
 ## [0.23.0] - 2026-08-20
 
 ### Added
--- a/dashboard/src/pages/Chats.css
+++ b/dashboard/src/pages/Chats.css
@@ -288,6 +288,7 @@
    ============================================================================= */
 .chats-page .chats-room {
   flex: 1;
+  min-width: 0;
   display: flex;
   flex-direction: column;
   background: var(--bg-light);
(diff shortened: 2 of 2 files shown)


rmyndharis:
Thanks for this, the diagnosis is exactly right. `.chats-room` is a row flex item next to a fixed 320px, non-shrinking sidebar, so its automatic minimum was pinned to the min-content of the `nowrap` header, and `.chats-layout { overflow: hidden }` clipped the excess with no scrollbar to recover it. The send button ended up outside the clip rect and unreachable by pointer.

Worth noting the reach is wider than "narrower layouts" suggests, since the threshold is set by header content rather than by viewport. With a 100 character group subject the send button is clipped by 84px at a 1440px viewport and only clears the edge above roughly 1525px, so this reaches ordinary laptop widths. The same line also fixes the raw JID row and the channel and status panes.

I pushed two things to the branch so it can land:

- `flex-shrink: 0` on the pane icon in `.chats-room-header`. Once the room can shrink, that icon became the only flexible item in the row, so an unbreakable channel or status title of roughly 26 characters or more collapsed it to zero width. Both of its siblings already opt out of shrinking.
- An `[Unreleased]` changelog entry.

Two neighbouring issues are out of scope here and are tracked separately: the composer's own minimum width still pushes the send button past the clip edge below a 1000px viewport, unchanged by this PR (#1422), and `.chats-room-header h2` lacks the truncation treatment its `h3` twin already has (#1423).

PR #1449, opened 2026-08-24
feat(session): warn when a linked session comes back asking for a QR

rmyndharis:
When WhatsApp stops accepting a session's stored credentials while the engine is down (the device was removed on the phone, the link expired offline, or the browser profile lost its state), the next start boots straight into the QR screen. Nothing marks that transition: no `LOGOUT`, no auth failure, no audit row, and `QR code generated` reads the same as a first pairing. Operators end up asking why a populated profile came back to QR with an empty unlink audit.

- `SessionEngineEventWiring.buildCallbacks` takes whether the session had completed a link before (`session.phone !== null` at start) and, on the first QR from such an engine, logs one warning with `action: 'relink_required'` naming the likely causes and the check to make on the phone. Periodic QR refreshes do not repeat it.
- The lifecycle passes that flag from the session row it already holds at `initializeEngine`.

Covered by two tests in `session.service.spec.ts` (warns once for a linked row, never for an unlinked one). Stacked on #1448 for the changelog placement; the two changes are independent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- A previously linked session that comes back asking for a QR now logs a warning (`relink_required`) naming the
+  likely causes, since an unlink that happened while the engine was down can leave no other trace.
+
 ### Changed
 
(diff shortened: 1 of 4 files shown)


rmyndharis:
Closed automatically when its base branch was deleted with #1448. The same commit, rebased onto main, continues in a replacement PR.

PR #1462, opened 2026-08-24
chore: migrate to bun as package manager

buluma:
Migrates OpenWA from npm to Bun. bun.lock replaces package-lock.json; postinstall now chains an npm rebuild of better-sqlite3 to fix a Bun-vs-Node native ABI mismatch that otherwise broke every DB-backed jest test. CI: lint/test/test-postgres/build moved to bun; dashboard subproject left on npm; audit job gets a package-lock-only step since npm audit can't read bun.lock. Dockerfile: bun copied in as a static binary from oven/bun, install swapped, runtime stays node:22-slim.


buluma:
Sorry — opened against the wrong repo by mistake (meant to target my own fork, not upstream). Closing.

PR #1307, opened 2026-08-15
Fixed chat loading issues using pagination and resolved session disco…

gujaratiprakash:
Fixed chat loading issues using pagination and resolved session disconnection through persistent session management.

Here is the complete Pull Request description ready to be copied into GitHub:

***
# Fixed chat loading issues using pagination and resolved session disconnection through persistent session management.

## Description

This PR addresses performance bottlenecks when loading large chat histories, prevents false transport disconnection errors during CDP command execution, catches unhandled media download rejections, and ensures full WhatsApp session persistence across Docker container rebuilds.

#### 1. Chat Pagination & Scroll UX (Dashboard Frontend & API)
* **API Pagination Support**: Updated `getChats` and `getChatMessages` in `dashboard/src/services/api.ts` to accept `limit` and `offset` query parameters.
* **Top-Scroll Infinite Loading**: Enhanced `useChatMessages.ts` with `fetchPreviousMessages()`, `hasMore`, and `isLoadingMore` states. In `ChatThread.tsx`, scrolling near the top (`scrollTop < 100px`) automatically fetches older message history pages and displays a loading indicator.
* **Scroll Position Lock**: Added `onOlderMessagesPrepended` in `useChatScrollPosition.ts` to calculate height delta (`scrollTop + (newScrollHeight - oldScrollHeight)`), keeping the viewport locked on the current message without jumpiness when older messages load.
* **Bottom-to-Top Conversation Flow**: Ensured chats open scrolled to the most recent messages at the bottom, maintaining pin alignment on media loads.
* **Sidebar Chat List Caching**: Cached sidebar chat queries per session via React Query in `Chats.tsx`, making tab switching (Chats / Channels / Status) instant with zero network delay.

#### 2. Session Disconnection & CDP Timeout Fixes (Backend & Engine)
* **Unlimited CDP Timeout**: Exposed `PUPPETEER_PROTOCOL_TIMEOUT` (defaulting to `0` / unlimited) in `src/config/configuration.ts`, `WhatsAppWebJsAdapter`, `WhatsAppWebJsPlugin`, and `EngineFactory` to prevent `ProtocolError: Runtime.callFunctionOn timed out` during heavy `Client.getChats()` operations.
* **Refined Transport Classifier**: Updated `isPageTransportError` in `whatsapp-web-js.adapter.ts` to return `false` on `timed out` errors so protocol timeouts no longer trigger false page death reports or disconnect live sessions.
[...]

Issue #628, opened 2026-07-05
[Bug]: Proxy with login pass does not work


gudge25:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm on the latest released version (or noted my version below)

### OpenWA version

0.8.7

### Deployment

Other (describe below)

### Database

PostgreSQL

### WhatsApp engine

whatsapp-web.js

### What happened?

Proxy (website webshare.io )with  login and password does not work
if IP auth ->  all good

proxy applied:
{"timestamp":"2026-07-05T11:53:55.432Z","level":"info","context":"WhatsAppWebJsAdapter","message":"Using proxy: socks5://socks5://jkxivowr-2:***@p.webshare.io:80"}

### Steps to reproduce

1- Use a proxy (login/pass)  from  webshare.io 
Free Proxy List

<img width="559" height="131" alt="Image" src="https://github.com/user-attachments/assets/b9bd4c31-5ee4-4d5c-8b0b-ae69788b7348" />

2- it is working from curl
curl --proxy "socks5://jkxivowr-2:***@p.webshare.io:80/" https://ipv4.webshare.io/
45.38.107.97

### Relevant logs

```shell
[Nest] 7  - 07/05/2026, 12:03:17 PM   ERROR [ExceptionsHandler] ProtocolError: Page.navigate timed out. Increase the 'protocolTimeout' setting in launch/connect calls for a higher timeout if needed.
    at <instance_members_initializer> (/app/node_modules/puppeteer-core/lib/cjs/puppeteer/common/CallbackRegistry.js:103:14)
    at new Callback (/app/node_modules/puppeteer-core/lib/cjs/puppeteer/common/CallbackRegistry.js:107:16)
    at CallbackRegistry.create (/app/node_modules/puppeteer-core/lib/cjs/puppeteer/common/CallbackRegistry.js:25:26)
    at Connection._rawSend (/app/node_modules/puppeteer-core/lib/cjs/puppeteer/cdp/Connection.js:108:26)
    at CdpCDPSession.send (/app/node_modules/puppeteer-core/lib/cjs/puppeteer/cdp/CdpSession.js:74:33)
    at navigate (/app/node_modules/puppeteer-core/lib/cjs/puppeteer/cdp/Frame.js:173:51)
    at CdpFrame.goto (/app/node_modules/puppeteer-core/lib/cjs/puppeteer/cdp/Frame.js:151:17)
[...]

Issue #368, opened 2026-06-19
[Bug]: The message isn't reaching some WhatsApp phone numbers.


yasielp:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm on the latest released version (or noted my version below)

### OpenWA version

0.4.3

### Deployment

Docker Compose

### Database

SQLite (default)

### WhatsApp engine

whatsapp-web.js

### What happened?

I am trying to send messages to several numbers using the OpenWA web interface's "message tester" option; it reports "success" for all of them, but the messages never reach the recipient.

### Steps to reproduce

I am trying to send messages to several numbers using the OpenWA web interface's "message tester" option; it reports "success" for all of them, but the messages never reach the recipient.

### Relevant logs

```shell

```

rmyndharis:
Hi @yasielp, thank you for the report and for filling in the version/deployment details!

Good news first: this is a well-understood pattern, and there are a few concrete things we can check together to pin it down.

A quick but important distinction in how the Message Tester reports results: a green "success" there means WhatsApp **accepted** the message and returned a message id -- it does **not** yet mean the message was **delivered**. Actual delivery is reflected by the delivery tick on the message in the **Chats** page (sent -> delivered -> read), and a `message.failed` webhook fires if WhatsApp later rejects it. So the first thing to check is: in the Chats view, does the message for one of the failing recipients ever advance past a single tick ("sent")? If it stays at "sent" and never reaches "delivered", that tells us WhatsApp accepted but silently dropped it -- which is almost always an account/engine-side condition rather than something OpenWA can force through.

We did a deep investigation of exactly this "200 + messageId but never delivered, for some recipients" symptom in #199 (worth a read). The common causes on the whatsapp-web.js engine are:
- **WhatsApp's LID migration** on the linked account (some recipients resolve to an `@lid` form the local store can't deliver to).
- A **session warm-up window**: whatsapp-web.js can emit `ready` slightly before WhatsApp Web finishes syncing and reloads once; sends in that brief window get a message id locally but are lost on the reload. This usually clears up shortly after linking.

To help us classify which one you're hitting, could you share:
1. Does it fail for **all** numbers, or only **some**? If only some, is there anything common about them (e.g., the account was recently LID-migrated, business vs personal, numbers you've never chatted with before)?
[...]

PR #1463, opened 2026-08-24
feat(engine): bound the whatsapp-web.js browser command timeout

JuanGalzerano:
greetings from Argentina!!!

This is the `protocolTimeout` piece from the #1307 review, split out on its own as suggested there. Nothing from the pagination side is included.

I rewrote this description after your review, so it describes the branch as it stands now rather than as I first opened it. The short version: the default is gone rather than raised, the value is bounded from above, and the two tests that could not fail are gone.

## The problem

Puppeteer drops any CDP command after 180s (`cdp/Connection.js`: `this.#timeout = timeout ?? 180_000`). One `client.getChats()` on an account with a few thousand chats can run past that, so the read fails with `Runtime.callFunctionOn timed out …` even though the renderer is fine and the session is still `ready`.

The part I can actually prove is the second half: there is no way to change it. `protocolTimeout` appears nowhere in `src/`, `.env.example` or `docs/`, so an operator who hits this has to fork or patch `node_modules`.

## About the missing measurement

You asked for one and I do not have it. You also found #199, #628 and #368, where the timeout was ending a hang caused by something else rather than failing a healthy slow read, and I could not turn up anything that contradicts that. So I am not going to claim a number from the mechanism alone.

That means the escape-hatch framing you offered is the one this takes. The default is not 300000, and it is not 180000 either. Unset leaves the launch option absent, so puppeteer-core's own `timeout ?? 180_000` applies. I did think about writing 180000 into `configuration.ts` for symmetry, but that pins today's figure in code that has no way to notice a bump, and it would make the parity `.env.example`, `docs/10` and the CHANGELOG all claim quietly false one release later.

There is a side effect worth calling out. It makes the launch-option presence check load-bearing. On the previous revision `configuration.ts` always returned a number, so the option was always spread, and `it('omits the option when unset')` passed with the entire spread deleted. It fails now.

Since nothing changes for anyone who leaves it unset, the CHANGELOG entry is a plain `### Added` rather than the behaviour change filed under it before.

## The knob

`PUPPETEER_PROTOCOL_TIMEOUT_MS`. Positive integer, at most 2147483647, rejected at boot on both ends.

`0` is refused because `common/CallbackRegistry.js` arms the timer under `if (timeout)`, so a falsy value arms nothing. I drove the installed registry to be sure rather than reading it off the source:

```
protocolTimeout=50  ->  rejected: Runtime.callFunctionOn timed out. Increase the 'protocolTimeout' …
protocolTimeout=0   ->  no timer, still pending
```

So `0` does not relax the bound, it removes it, and a wedged renderer would sit on the command with the HTTP request stuck behind it. That is the 503-on-a-dead-page contract turned off by an env var.

The ceiling is the one you asked for. `checkPositiveInt` only bounded from below, so anything above `2147483647` reached Puppeteer intact:

```
TimeoutOverflowWarning: 2147483648 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
```

Every command in the launch handshake then fails with `Target.setDiscoverTargets timed out. Increase the 'protocolTimeout' setting`, which is the error the new troubleshooting entry tells the operator to raise the value for. Your point about the docs forbidding `0` and steering people at a row of nines is the reason this matters, not the arithmetic. Both ends are rejected at boot rather than clamped, so they find out the value they wrote is not the value they would have got.

`configuration.ts` uses the guarded form the other positive-only `*_MS` knobs use. That removed the `typeof === 'number' && > 0` guard from the launch options and three of the five launch-option cases with it, since the range is enforced at the config layer now.

Plumbing follows the existing engine knobs (`configuration.ts` → factory / `engine/builtin/whatsapp-web-js` → adapter config → launch options), plus `.env.example`, both compose forwards, and the `BLANK_SHADOWED_ENV_KEYS` entry those need. The key also moved inside the `checkPositiveInt` array literal, so `docs-env-example.spec.ts`'s extractor picks it up from the validator instead of relying on `env-precedence.ts` to carry it.

## Docs

`docs/12` recommended 600000 and said nothing about what raising it costs. It uses 300000 now and adds the part that was missing: the budget is also how long a wedged browser holds a command before the gateway can give up on it. At 300000 that is five minutes instead of three, and for that whole window the stuck command holds the logout fence behind it, so `start` and `delete` answer `409 SESSION_NAME_TEARDOWN_PENDING` and `forceKill` refuses with `400` because the engine was already evicted. It also says outright that there is no measurement here, so anything above the default is something you reach for after seeing the error, not a value to set up front.

`docs/10` had it as a live assignment of 300000 inside the copy-paste production `.env` template, which told every operator to do exactly what `docs/12` says not to. It is commented out like the other optional knobs.

## The carve-out

You left this to me, so it stays, and the label from last time still applies: it changes nothing today. puppeteer-core 24.38.0 emits that message from one site with no `Protocol error (…)` prefix, none of the `PAGE_TRANSPORT_ERROR_PATTERN` signatures appear in it, and deleting the guard leaves the suite green. It earns its place by failing if the message drifts, not by covering a branch.

What changed is that it can now detect the drift at all. The spec provokes the message from the installed puppeteer-core (`new Callback(1, 'Runtime.callFunctionOn', 1)`) instead of the literal I had copied by hand. You were right that the old version could not fail on the one thing it existed to catch.

Two tests are gone. The falsy-budget probe exercised no code in this repo and survived every revert of the PR. The four death-signature cases duplicate the six in `whatsapp-web-js.adapter.spec.ts` -- yours has `TargetClosedError: page closed` and a second Target-closed shape on top of my four, and I checked none of the six matches the timeout pattern, so the guard cannot swallow them.

Six tests remain. I checked each against a revert of the code it guards: the launch-option spread, the boot rejection, the config range and the message-drift assertion all fail when their production code is removed.

## Testing

`wwebjs-protocol-timeout.spec.ts` passes, 6 tests.

`npm test` shows no change against the pre-change baseline here, and this time I baselined it instead of asserting it: the four suites that fail on my machine fail identically with the branch stashed, 9 failed / 605 passed either way, and they are the known Windows `ps` and POSIX-permission ones.

`npm run test:docs` passes 262 including `docs-env-example.spec.ts`. The one failing suite there, `docs-ci-jobs`, fails the same way on a clean `upstream/main`.

`tsc --noEmit` and `eslint` are clean, and `prettier --check` is clean on every touched file.

CI on this branch is yours to approve. I know the run I linked last time was on my fork and does not count as evidence here.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -127,11 +127,13 @@ STATUS_SEED_ON_READY=false
 # Path to a system Chromium/Chrome binary. Leave unset to use the bundled Puppeteer browser.
 # Required in the Docker image and on hosts without a bundled browser (e.g. Alpine/ARM).
 # PUPPETEER_EXECUTABLE_PATH=/usr/bin/chromium
-# How long a single browser command may take before it is abandoned, in milliseconds. Default 300000
-# (5 min), above Puppeteer's own 180000. Raise it if a large account (thousands of chats) fails a
-# read with "Runtime.callFunctionOn timed out"; the renderer is still working, the command budget
-# just ran out. Must be a POSITIVE integer — 0 does not mean "no limit later", it means no timer is
-# armed at all, so a wedged browser holds the request forever instead of failing it as 503.
+# How long a single browser command may take before it is abandoned, in milliseconds. Unset means
+# Puppeteer's own budget (180000 today), so leaving it unset changes nothing. Raise it if a large account
+# (thousands of chats) fails a read with "Runtime.callFunctionOn timed out"; the renderer is still
+# working, the command budget just ran out. Must be a POSITIVE integer no greater than 2147483647:
+# 0 does not mean "no limit later", it means no timer is armed at all, so a wedged browser holds the
+# request forever instead of failing it as 503; and above 2147483647 Node's timer overflows and
+# fires after 1 ms, so the browser never finishes launching. The gateway refuses to boot on either.
 # PUPPETEER_PROTOCOL_TIMEOUT_MS=300000
 # A freshly linked account is shown a "What's new on WhatsApp Web" modal that must be acknowledged or
 # WhatsApp unlinks the companion a few minutes later. The whatsapp-web.js engine clicks it
--- a/.env.example
+++ b/.env.example
@@ -132,7 +132,7 @@ STATUS_SEED_ON_READY=false
 # (thousands of chats) fails a read with "Runtime.callFunctionOn timed out"; the renderer is still
 # working, the command budget just ran out. Must be a POSITIVE integer no greater than 2147483647:
 # 0 does not mean "no limit later", it means no timer is armed at all, so a wedged browser holds the
-# request forever instead of failing it as 503; and above 2147483647 Node's timer overflows and
+# request forever instead of failing it; and above 2147483647 Node's timer overflows and
 # fires after 1 ms, so the browser never finishes launching. The gateway refuses to boot on either.
 # PUPPETEER_PROTOCOL_TIMEOUT_MS=300000
 # A freshly linked account is shown a "What's new on WhatsApp Web" modal that must be acknowledged or
--- a/.env.example
+++ b/.env.example
@@ -127,6 +127,12 @@ STATUS_SEED_ON_READY=false
 # Path to a system Chromium/Chrome binary. Leave unset to use the bundled Puppeteer browser.
 # Required in the Docker image and on hosts without a bundled browser (e.g. Alpine/ARM).
 # PUPPETEER_EXECUTABLE_PATH=/usr/bin/chromium
+# How long a single browser command may take before it is abandoned, in milliseconds. Default 300000
+# (5 min), above Puppeteer's own 180000. Raise it if a large account (thousands of chats) fails a
+# read with "Runtime.callFunctionOn timed out"; the renderer is still working, the command budget
+# just ran out. Must be a POSITIVE integer — 0 does not mean "no limit later", it means no timer is
+# armed at all, so a wedged browser holds the request forever instead of failing it as 503.
+# PUPPETEER_PROTOCOL_TIMEOUT_MS=300000
 # A freshly linked account is shown a "What's new on WhatsApp Web" modal that must be acknowledged or
 # WhatsApp unlinks the companion a few minutes later. The whatsapp-web.js engine clicks it
 # automatically, matching the English button label `Continue`. If your WhatsApp Web renders the modal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,26 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- `PUPPETEER_PROTOCOL_TIMEOUT_MS` sets how long a single browser command may take on the
+  whatsapp-web.js engine, defaulting to 300 000 ms, above Puppeteer's own 180 000 ms. A read that
+  walks the whole store, such as `GET /sessions/{sessionId}/chats` on an account with thousands of
+  chats, could run past the stock budget and fail with `Runtime.callFunctionOn timed out` while the
+  renderer was still working. The value must be positive: Puppeteer only arms its timer for a
+  truthy one, so `0` drops the bound instead of raising it and lets a wedged renderer hold the
+  request forever.
+
+### Changed
+
+- The whatsapp-web.js transport classifier now rules out a protocol timeout explicitly. That
+  message carries no transport-death signature on the current Puppeteer, so nothing was
+  misreporting it and behaviour is unchanged; the point is the spec, which pins the invariant in
+  both directions against the verbatim Puppeteer string so a future bump cannot start reading a
+  slow command as a dead page.
+
 ## [0.23.3] - 2026-08-24
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,27 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `PUPPETEER_PROTOCOL_TIMEOUT_MS` sets how long a single browser command may take on the
-  whatsapp-web.js engine. A read that walks the whole store, such as
-  `GET /sessions/{sessionId}/chats` on an account with thousands of chats, can run past Puppeteer's
-  budget and fail with `Runtime.callFunctionOn timed out` while the renderer is still working, and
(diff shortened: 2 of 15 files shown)


rmyndharis (changes requested):
Thanks for splitting this out, and for the probe work behind it. The plumbing is complete and the
reasoning about `0` is the part that matters most, so this is close. A few things before merge.

**Bound the value from above.** `checkPositiveInt` only bounds from below, so a value above
`2147483647` reaches Puppeteer intact and Node collapses the timer:

```
TimeoutOverflowWarning: 2147483648 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
```

The browser then never finishes launching and every command fails with
`Target.setDiscoverTargets timed out. Increase the 'protocolTimeout' setting ...`, which is the exact
error the new troubleshooting entry tells the operator to raise the value for. Since the docs
correctly forbid `0`, an operator who wants an effectively unlimited budget is steered straight at a
row of nines. Please reject anything over `2147483647` at boot, or clamp at `configuration.ts`.

**The default move needs to be in `### Changed`.** `### Added` currently carries the fact that every
whatsapp-web.js deployment goes from Puppeteer's 180000 to 300000 on upgrade, while `### Changed`
says behaviour is unchanged. That is true of the classifier and misleading as the section a reader
scans for behaviour changes. One line is enough: a wedged browser now holds a command, and the
logout fence behind it, for five minutes instead of three, during which start and delete answer
`409 SESSION_NAME_TEARDOWN_PENDING` and `forceKill` refuses with 400 because the engine was already
evicted.

**`docs/12` recommends 600000**, double the default this PR sets, with no mention of that cost.
Please use 300000 there and add the clause above.

**On the premise.** The mechanism checks out: `session.service.ts` issues an unbounded
`engine.getChats()` and paginates in Node afterwards, and whatsapp-web.js makes that a single
`pupPage.evaluate` over the whole store, so it is one CDP command under one budget. What is missing
is a measurement. Searching this repo for `protocolTimeout` turns up #199, #628 and #368, and in both
real occurrences the timeout was ending a hang caused by something else, not failing a healthy slow
read. If you have an account size, an endpoint and a wall-clock time, please add them. If you do not,
the honest framing is that this is an escape hatch and the default should stay at Puppeteer's 180000,
which also makes the CHANGELOG note above unnecessary.

**On the carve-out, your call.** Thank you for saying plainly that it changes nothing rather than
presenting it as a fix. That is worth more than the code. Having checked it here: deleting the guard
leaves all 11 of the new tests green, and the reason is unreachability rather than weak tests.
puppeteer-core 24.38.0 emits that message from exactly one site with no `Protocol error (...)`
prefix, and the prefix comes only from the mutually exclusive reject path. The four death-signature
cases also duplicate coverage that already exists in `whatsapp-web-js.adapter.spec.ts`. Keeping it
costs about twelve lines and is documented honestly, so either way is fine by me. If it stays, the
spec should derive the message from the installed library rather than pinning a hand-copied literal;
as written it cannot fail on the drift it exists to catch.

**One convention note.** Every other positive-only `*_MS` knob in `configuration.ts` uses the guarded
form (`Number.isFinite(n) && n > 0 ? n : default`). Using it here makes the parsed value always
positive, which removes the need for the `typeof === 'number' && > 0` guard in the launch options and
three of the five launch-option test cases with it.

CI has not run on this branch yet; the linked run is on a fork, so I will approve the workflow here
before merging.

Everything else looks right. The BLANK_SHADOWED_ENV_KEYS entry is not boilerplate, it is what stops
an empty compose forward from shadowing a hand-edited `data/.env.generated`, and the plumbing hits
every parity gate the repo enforces including `docs-env-example.spec.ts`, which `npm test` skips and
only `npm run test:docs` runs. The troubleshooting entry being filed under the verbatim error string
is the right instinct.

JuanGalzerano:
Thanks for the thorough read — pushed `ac0dd2fc` addressing all of it.

- **Upper bound**: `PUPPETEER_PROTOCOL_TIMEOUT_MS` now rejects anything above `2147483647` at boot, with the overflow explained in the error message. Also moved the key into `checkPositiveInt`'s array literal so `docs-env-example.spec.ts` can see it.
- **Default**: dropped rather than raised. Unset now leaves the launch option absent — `cdp/Connection.js` reads `timeout ?? 180_000`, so absence is how to say "Puppeteer's own number" without pinning `180000` in code that would silently go stale on a Puppeteer bump. You're right that I had no measurement, only a mechanism argument; the three issues you pointed at (#199, #628, #368) all read as the timeout catching a hang from something else, not a healthy slow read. Since nothing changes for a deployment that leaves the var unset, the CHANGELOG entry is a plain addition now, not a behaviour change filed as one.
- **`docs/12`**: 300000, with the cost note (five-minute hold on the logout fence, `409`/`400` for `start`/`delete`/`forceKill` during it).
- **Guarded parse form**: `configuration.ts` now matches the other `*_MS` knobs, which let me drop the redundant `typeof === 'number' && > 0` check from the launch options and three of the five spec cases that existed only to cover it.
- **Spec derives the message from the library**: it now imports `Callback` from the installed `puppeteer-core` and provokes the real rejection instead of pinning a hand-copied string. Confirmed it fails on a Puppeteer bump that reshapes the message — reverting the guard with this spec in place, it goes red.

On the carve-out itself — I'm keeping it. You're right that all 11 tests stayed green with it deleted, and now that the spec provokes the real message that's a meaningful signal, not a gap in coverage: the guard is provably unreachable on 24.38.0. But per the docstring, that's pinning intent for the day Puppeteer does emit the death signature in this message, and the 12 lines felt cheap for that. Open to being told that's not worth it.

CI: https://github.com/JuanGalzerano/wa-galzerano/actions/runs/32982508936

rmyndharis:
Thanks, this is in good shape. I pushed three small things straight to the branch rather than send it back for another round:

- **Bounds at the launch sink.** `configuration.ts` is not the only writer of that blob. A plugin config stored through `PUT /api/plugins/whatsapp-web.js/config` is validated only as an object and merged over the env values at every boot, so a `0` written there reached `puppeteer.launch` and armed no timer at all. The sink now range-checks instead of testing `!== undefined`, which is what the comment above it already promised.
- **Two test pins.** Deleting the classifier carve-out left the spec at 6/6, so it could not fail on the one thing it exists for. It now asserts the `Protocol error (label): ...` shape puppeteer actually emits. Same for the plugin's `protocolTimeoutMs` hop, which passed with the key present but undefined.
- **Three prose corrections.** The budget bounds a command that is slow but still running, so it is not what protects you from a wedged page: the liveness watchdog probes every 60 s with a 15 s timeout and calls it at two consecutive failures, whatever this value is. The logout fence only exists once a logout has been issued, not during the stuck read the entry documents. And the timing-out request is not a 503. The changelog entries are cut to the length the rest of the file uses.

Your measurement on `0` and on the overflow was right and is what the sink guard now encodes.

Full suite is green here: 345 suites / 5940 tests, `tsc`, `eslint`, `format:check`, `test:docs` 24/264, `build`, and the coverage ratchet with no threshold missed. The `9 failed` you saw locally does not reproduce on macOS, so I agree it is the Windows `ps` and POSIX-permission pair.

Merging this first. #1464 will rebase on top of it for the changelog.

PR #1464, opened 2026-08-24
feat(dashboard): load older chat history as you scroll up

JuanGalzerano:
greetings from Argentina!!!

The chat-history pagination from the #1307 review, on its own branch. Dashboard-side only; none of the engine, Docker or compose changes from that PR are here.

I rewrote this description after your review, so it matches the branch as it stands. The blocker is fixed, and so are the three things behind it. Two of your points changed the design rather than the code, so I have called those out below.

## The problem

The chat room fetched one fixed window and stopped: the newest 100 persisted rows, plus up to 100 rows of engine history merged in, so roughly 200 messages. Anything older had no way to be reached -- you scroll up and it just ends. (The original version of this description said 100, which you correctly pointed out is not what the `queryFn` does. Fixed here and in the CHANGELOG.)

The backend already does its half: `GET /sessions/{id}/messages` takes `limit` and `offset`. So this is frontend work only.

## The cursor

Your note was that the fetch-older offset has to count DB rows already fetched, not the length of the merged array. That drove the page shape: each page keeps `db` and `history` as separate arrays instead of pre-merging them. Merging first destroys the count, because the merged list adds engine-history items and drops the duplicates between the two sources.

Termination is the part you found wrong, and it was. `nextMessagePageParam` compared `dbRowsFetched`, which grows as `upsertIntoPages` merges live arrivals into page 0, against a `total` frozen at the last fetch. Your 105-row walkthrough is exactly right:

```
after first page      -> next = 100
after 1 live message  -> next = 101
after 5 live messages -> next = undefined   (dbRowsFetched 105, total 105)
```

The offset arithmetic stays, since a server insert really does shift the window by one. Only the stopping rule changes: the page records how many rows its fetch returned and paging ends when that comes back short of `MESSAGE_PAGE_SIZE`. `total` is gone from `MessagePage` along with the rule that read it.

## The blocker

You were right about all four handlers, and about why nothing caught it. `getQueriesData<T>` is an unchecked cast and there were no component tests, so the suite could not have seen it.

All four now read through `cachedSessionThreads` and write through `updateCachedMessages`, and the `<ChatMessageView[]>` generic is gone from those reads. That leaves one unchecked cast, inside the hook that owns the shape, instead of five at call sites that do not.

The edit handler needed the extra work you flagged. `editedIndex === list.length - 1` under a comment claiming the caches are chronological is false for a page, since `page.db` is stored raw at `createdAt DESC` and the last row of a page is its oldest message. Position is asked of the merged thread now, and the comment is fixed.

## The two things your review did not name, which came out of fixing it

Both are places where my first fix was wrong, so I would rather state them than have you find them again.

**The scroll restore rewrite regressed before it worked.** You asked for live measurement instead of the capture-and-return-a-closure protocol, which was right. But applying the delta on the first commit whose `scrollHeight` changed is not the commit that lands the page -- it is the commit that renders the spinner. `.messages-loading-older` is an in-flow child of `.room-messages` (`flex: 0 0 auto`), so it grows the container by about 46px one commit early, spends the correction on itself, and the real prepend then goes uncorrected. The view ends up thrown a full page backwards, which is worse than what was there before.

The height is read when the page is requested and the delta applied on the falling edge of the in-flight flag. The falling edge specifically, not `!isFetching`: the request happens in a scroll handler while the query only reports itself in flight a turn later, so any commit in between -- a live message, a media decode, another query resolving -- would otherwise consume the pending correction. `scrollTop` is still read live at apply time, so a deliberate reposition during the fetch survives, and the flag is cleared on a chat switch so it cannot fire against the wrong thread.

**The in-flight write replay was dead on arrival.** Your point about TanStack snapshotting `data.pages` in `onFetch` is correct, and the fix records those writes and replays them on the landed page. My first attempt put the replay and the leave-the-chat cleanup in one effect with `isFetching` in the deps. React runs the previous cleanup before the new body, so on the `true -> false` edge the cleanup deleted the queue and the replay then found nothing. The mechanism was inert and the suite did not notice, because nothing covered it.

The cleanup is its own effect keyed only on the chat now, declared first. Each queued write also records the cache value it produced, so a fetch that failed -- which leaves the cache exactly as the last write left it, meaning nothing was lost -- is told apart by identity from one that replaced it, and the queue is not applied twice.

## Media payload cap

I took the second option you offered and corrected the docstring rather than capping per page. Nothing that worked before regresses, and scaling `keep` with `pages.length` would remove the bound where it matters, which is the rendered set. The paragraph in `useChatMessages`' docstring that described this bound is back too -- you were right that I deleted it instead of updating it.

## Smaller items

`mapCachedMessages` is exported under the name the call site already imported, so the identical passthrough is gone. `cachedSessionThreads` takes a match predicate and tests it against the raw pages, so a realtime event only pays to merge the thread that actually holds the message rather than every open chat. `byMessageId` replaces the three hand-inlined copies of the two-candidate id test. The `.ts` import extensions are reverted.

The CHANGELOG conflict with #1463 has not materialised yet -- `upstream/main` is still at `3b7fbe7e` and both branches are clean fast-forwards. Whichever you take second, I will rebase.

## Testing

Four component tests in `pages/Chats.test.ts`, which is the gap you named. Each one I checked against a revert of the code it guards, since that is the whole reason they exist:

- scrolling to the top requests `offset=100` and renders the older bubbles, then stops instead of asking past the end
- a `message.ack` over a paged thread, on a row only the oldest page holds -- fails with the verbatim `list.findIndex is not a function` on the old handler
- the reading position across a prepend -- fails at 46, which is the spinner
- a message arriving while an older page is in flight -- gone from the DOM without the replay

The last three need commits the runner can tell apart, so the older page is gated open the way the media route already was, and the thread's `scrollHeight` is a getter over the live DOM instead of a pinned constant. A constant cannot distinguish the spinner commit from the landing commit, which is how the scroll bug hid. They also need a socket that never dials, so `socket.io-client` resolves to a double under the runner. That is inert for the other test files: only `Chats.tsx` imports `useWebSocket`, and `connect()` bails without an api key anyway.

Plus 9 unit tests across `messagePages.test.ts` and `scrollDecision.test.ts`, including the 105-row case above.

`npm --prefix dashboard run test` is 351 passing (332 before), `build`, `lint`, `typecheck` and `format:check` all clean. Locale parity passes, so `loadingOlderMessages` is in all 13 catalogs.

One limit worth stating, same as last time: the DOM heights in the scroll test are modelled, not laid out, since jsdom does no layout. The commit ordering and the arithmetic are pinned; the actual pixel geometry is not.

Also noted your point about `message.service.ts` ordering by `createdAt DESC` with no tiebreaker -- leaving that to your change.


rmyndharis (changes requested):
This is the piece I most wanted from #1307, and the design is right: the cursor counts DB rows rather
than rendered ones, engine history stays a first-page backfill, and the scope is clean (dashboard,
CHANGELOG and `docs/17` only, rebased on current main, none of the items I asked to be dropped).
Requesting changes for one blocker and a few things behind it.

## Blocker: four WebSocket handlers still read the messages cache as a flat array

The cache value changes from `ChatMessageView[]` to `InfiniteData<MessagePage>`, but only one of the
five places in `Chats.tsx` that touch it directly was converted. These four still call array methods
on it:

- `Chats.tsx:428` delivery ack, `list.findIndex` at :433
- `Chats.tsx:457` reaction, `list.findIndex` at :462
- `Chats.tsx:486` revoke, via `findRevokedIndex` in `utils/chatMessages.ts:266`
- `Chats.tsx:506` edit, via `applyMessageEdit` in `utils/chatMessages.ts:279`

Seeding a `QueryClient` with the page shape this PR writes and running the ack handler body against
it gives:

```
cached value type: [object Object] | keys: [ 'pages', 'pageParams' ]
THREW: TypeError: list.findIndex is not a function
```

So delivery ticks, reactions, revokes and edits stop working in the chat view, on ordinary use of any
open chat. There is no try/catch in `hooks/useWebSocket.ts`, so it fails silently with nothing
surfaced to the user.

Please route all four through the paged helpers, and drop the `<ChatMessageView[]>` generic from those
`getQueriesData` calls while you are there. That cast is the only reason `npm run typecheck` passes on
this; without removing it the same class of mistake reaches us again on the next cache change.

The edit handler needs more than a mechanical port. `Chats.tsx:518` relies on
`editedIndex === list.length - 1` with the comment "Message caches are chronological", and that is no
longer true: `page.db` is stored raw from the server, which orders `createdAt DESC`, so per-page
arrays are newest-first while the old flat cache was ascending. Compute that check against the
flattened merged view and fix the stale comment.

I want to be clear that the green CI is not the problem here. `getQueriesData<T>` is an unchecked
cast, and there are no `*.test.tsx` files in the dashboard at all, so no component renders anywhere in
the suite. Nothing we currently run could have caught this.

## Paging stops early once messages arrive

`nextMessagePageParam` compares a count that grows against a total that does not: `dbRowsFetched` sums
the rows the cache holds and `upsertIntoPages` appends every live arrival to page 0, while `total`
stays at whatever the last fetch reported. On a 105 row chat:

```
after first page      -> next = 100
after 1 live message  -> next = 101
after 5 live messages -> next = undefined   (dbRowsFetched 105, total 105)
```

The five oldest rows become unreachable until a reload, a socket reconnect or `gcTime`. The offset
arithmetic itself is drift-correct and worth keeping, since each server insert really does shift the
window by one. It is only the termination that is wrong. Record the fetched count on the page in the
`queryFn` and stop on `fetched < MESSAGE_PAGE_SIZE`, the usual limit/offset termination.

## Cache writes made during a fetch are lost when the page lands

TanStack v5 snapshots `data.pages` when the fetch starts (`infiniteQueryBehavior.ts`, `oldPages` is
read in `onFetch`), so anything written with `setQueryData` while an older page is in flight is
overwritten by the result. With `staleTime: Infinity` there is no refetch to recover it. An
optimistically sent message during that window can leave the DOM permanently. Please reconcile
in-flight writes, or at minimum re-apply the optimistic message after a page lands.

## Scroll restore

Two issues, both fixed by the same change. `restoreOlderScrollRef` is never cleared on a chat change,
so a switch made while an older page is in flight applies the previous chat's delta to the new one
(clearing it in the chat-switch effect at `Chats.tsx:216` will not work, that is a passive effect and
runs after the layout effect has written). And the restore writes an absolute `top + delta` from
geometry captured at trigger time, so any deliberate repositioning during the fetch is discarded.

Rather than patching both, replace the capture-before and return-a-closure protocol with a live
measurement: keep a `prevScrollHeightRef` updated in a layout effect on each commit, and apply
`el.scrollTop += el.scrollHeight - prev` on the commit that lands the page. That removes
`onOlderMessagesPrepended` and the closure-in-a-ref entirely, never reads a stale number, and cannot
fire against the wrong chat.

## Media payload cap no longer covers the fetch window

On main the value stored in the cache was the output of `mergeChatMessages`, so `capMediaPayloads`
bounded the cache itself. Now the cache holds raw pages and the cap runs only inside `select`.
`MEDIA_PAYLOAD_CACHE_LIMIT` stays at 100 while the window becomes pages times 100, so `select` strips
the oldest payloads inside the scrollable window: two 100 payload pages render 100 of 200. Nothing
that worked before regresses (the newest 100 always survive, and the omitted marker is still a working
download button), so this is not a blocker. Please either cap per page before the merge or leave the
behaviour and correct the docstring, which currently states the invariant in as many words. Do not
scale `keep` with `pages.length`; that removes the bound on the rendered set, which is where the
second `data:` URI copy and the decoded bitmap live. Note the paragraph in `useChatMessages`'s
docstring that described this bound was removed rather than updated.

## Smaller items

- `updateCachedMessages` is a passthrough to `mapCachedMessages` with an identical signature. Rename
  `mapCachedMessages` and export it; the call site in `Chats.tsx` keeps the name it already imports.
- Revert the `.ts` import extensions in `useChatMessages.ts`. `messagePages.ts` genuinely needs its
  extension because the node test runner resolves it, but nothing imports `useChatMessages.ts` from a
  test, and dashboard sources are 199 extensionless to 7 explicit. The same PR leaves
  `ChatComposer.tsx` and `Chats.tsx` importing it extensionless, so it is not consistent either.
- The CHANGELOG entry and the PR description both say the room stops at the newest 100 messages. The
  same `queryFn` also pulls up to 100 rows of live engine history and merges them, so the room shows
  roughly 200 today. Worth correcting so the entry does not overstate what changed.
- This will conflict with #1463 on `CHANGELOG.md`, since both open a new `[Unreleased]` section.
  Whichever lands second needs a rebase.

## Tests

My original ask for a spec covering the paging behaviour is half met: the two decision functions are
well covered (the case where 2 DB rows and 3 history items yield offset 2 is exactly the regression I
was worried about), but nothing covers the feature in situ, which is why the blocker above shipped
green. One component-level test in `pages/Chats.test.ts` would close it: three `Object.defineProperty`
lines for `scrollHeight`/`clientHeight`/`scrollTop` on `.room-messages` and one `fireEvent.scroll` is
enough to assert the `offset=100` request and the older bubbles rendering. A second test dispatching a
`message.ack` while a paged chat is open would have caught the blocker outright.

Separately and not your problem: `message.service.ts` orders by `createdAt DESC` with no tiebreaker,
and the inbound projector writes through `repository.insert()` so `createdAt` picks up SQLite's whole
second `datetime('now')` default. Ties are therefore common and offset paging can skip a row. I will
add an `id` tiebreaker in its own change.

CI has not run here yet; the linked run is on a fork, so I will approve the workflow on this branch
once the above is in.

To be explicit about what is good, because the list above is long: the cursor design is the hard part
and it is right. The asymmetric error handling, where page 0 may fall back to history alone but an
older page throws rather than resolving empty and reading as "no more messages", is easy to get wrong
and was got right. The `ChatComposer` change is a real simplification, collapsing 27 lines of
hand-rolled echo-race reconciliation into one call with both orderings pinned by tests. The layout
effect placement and the reason given for it are correct. And `mapCachedMessages` returning undefined
on a missing cache incidentally fixes a latent bug on main, where `(old = []) => ...` would seed a
phantom fresh empty slice for a never-opened chat. Thank you for stating the testing limits of the
stub in the description; that is what pointed me at both defects.

JuanGalzerano:
Thanks for actually running the ack handler against the paged shape — that's a real bug and I'm glad it didn't reach anyone. Pushed `faf2426b` addressing everything.

- **Blocker**: all four handlers (ack, reaction, revoke, edit) go through the paged helpers now. Dropped the `<ChatMessageView[]>` generic from `getQueriesData`, so the cast that hid this lives in one place — the hook that owns the shape — instead of at five call sites that don't. The edit handler's `editedIndex === list.length - 1` check is against the merged thread now; the "message caches are chronological" comment was wrong on the paged shape, since pages store server order (`createdAt DESC`), so a page's last row is its oldest message.
- **Paging stopping early**: confirmed the mechanism — `dbRowsFetched` grows with every live arrival merged into page 0, `total` doesn't, so on a chat with live traffic they meet before the real end. Termination is the ordinary short-page signal now (`fetched < pageSize`); the offset itself was fine, since it does need to count arrivals.
- **In-flight writes lost to a fetch**: recorded and replayed on top of the landed page, keyed by chat and cleared on chat-change in its own effect — hanging it off the replay effect would delete the queue right before that same run replays it.
- **Scroll restore**: rebuilt as you described, live height read at request time and delta applied on the falling edge of the in-flight flag. Two more edges under that shape, in case they're worth naming: the older-page spinner is an in-flow child of the same container, so its own height lands one commit before the messages and was eating the correction if I read it there instead of at request time; and the fetch is triggered from the scroll handler while the query only reports itself in flight a turn later, so "no longer fetching" alone would let a commit in between consume it too.
- **Media cap docstring**: corrected to say it bounds the rendered set via `select`, not the fetch window. Didn't scale it with page count — agreed that removes the bound where it costs a second `data:` copy and a decoded bitmap.
- **Smaller items**: `mapCachedMessages` is exported directly now, `updateCachedMessages` is gone. `.ts` import extensions reverted in `useChatMessages.ts`; kept in `messagePages.ts`, which the node test runner resolves directly.

`npm run typecheck` passes with the generic gone — it's clean because there's nowhere left for it to hide, not because I checked it manually.

CI: https://github.com/JuanGalzerano/wa-galzerano/actions/runs/32982700692

rmyndharis:
The design holds up: the offset-vs-rendered-length distinction, the replay queue's premise and the effect ordering are all correct, and the 105-row case is the right test to have written. Four things before I can take it.

**1. Scroll anchoring applies the correction a second time.** `.chats-page .room-messages` sets no `overflow-anchor`, so Chrome and Firefox already move `scrollTop` themselves when content is prepended. Reading `el.scrollHeight` in the layout effect forces the layout that applies that adjustment, so the `el.scrollTop` you read on the next line already contains it and `scrollTop + grew` adds the page height twice.

Chrome 146, your CSS and the hook's exact read/write order, 100 bubbles with 20 prepended (`grew` 1320):

```
start  topAtApply  final  ideal
0      0           1320   1320   correct; anchoring is suppressed at exactly 0
20     1340        2660   1340   off by 1320
99     1419        2739   1419   off by 1320
```

Paging fires at `scrollTop < 100`, so every offset a real wheel or trackpad gesture lands on is wrong, and the reader is thrown a full page toward the newest messages. Only a fling that clamps to exactly 0 hits the good path, which is the offset all four `makeScrollable` call sites pass, and jsdom models no anchoring, so the suite cannot see it. `overflow-anchor: none` on that rule fixes every offset. Keep the manual write, WebKit implements no scroll anchoring.

**2. A failed older page blanks the conversation.** `ChatThread.tsx` tests `messagesError` before `messages.length`, and TanStack v5 sets `status: 'error'` on any fetch failure while keeping `data`. One 500 or 429 on an older page replaces the loaded thread with the error placeholder, and the collapsed container can then never satisfy `scrollHeight > clientHeight`, so scrolling cannot retry. Wants `messagesError && messages.length === 0`, plus an inline failure where the spinner renders.

**3. A failed send shifts the cursor past a real row.** `dbRowsFetched` sums `page.db.length`, and the optimistic placeholder stays in page 0 with `status: 'failed'`; nothing removes it. The next page is requested at `offset=101`, the 101st-newest row is never fetched, and nothing in the UI shows the hole. Exclude the `temp_` ids from the count. Not rows without a `waMessageId`, which is nullable on genuine DB rows.

**4. `discardWritesLostToFetch` loses the write it is dropping.** The justification is that reopening the chat refetches a thread that already contains it, but at `staleTime: Infinity` it does not refetch, and the in-flight page is not cancelled, so it lands on its pre-write snapshot. Leaving a chat mid-fetch drops that message until `gcTime`. Either keep the queue, since the replay effect re-fires on remount, or `removeQueries` at discard.

Smaller, worth a pass while you are in there:

- Three comments state the opposite of the code: the spinner is an in-flow child and does move the list, nothing reads `total`, and the surviving per-page cap strips the newest payloads first on a `createdAt DESC` page.
- `'pulls exactly one older page, then stops'` passes with the termination rule removed. The third request goes to `offset=120` and nothing asserts on it.
- `MESSAGE_PAGE_SIZE` silently assumes the gateway clamps `limit` to exactly 100. Worth a comment at least, since a lower clamp would end paging after page one.

`capMediaPayloads` in `select` is fine as a follow-up, though note it drops the cross-page bound on cached base64. The `createdAt` ordering is mine and I will handle it separately: `addOrderBy('message.id')` over a v4 uuid measures worse, so it needs a keyset cursor.

Nothing here is a redesign. #1463 lands first, so rebase for the changelog.

JuanGalzerano:
Thanks for the depth here, especially the anchoring one — I'd never have found that without you running it in a real browser. Pushed `e55bdd21` addressing all four plus the smaller items.

- **Scroll anchoring**: `overflow-anchor: none` on `.room-messages`. Confirmed the mechanism matches what you measured — the manual correction stays as the only one that runs. Left untested; agreed jsdom can't see this.
- **Failed older page blanking the thread**: the full-screen error now only shows at `messagesError && messages.length === 0`. A failure with messages already loaded renders inline where the spinner sits, with a retry hint, so the container never collapses out from under a retry.
- **Cursor past a failed send**: `dbRowsFetched` excludes `temp_`-prefixed ids now. Matched on the prefix specifically, not a missing `waMessageId` — you're right that's nullable on genuine rows too.
- **`discardWritesLostToFetch` losing the write**: took the first option — removed the discard-on-unmount effect entirely. The queue lives until replay actually applies it; remounting the same chat re-runs that effect, and the identity check still catches the landed page's value differing from what was recorded.

Also found while fixing #3: the per-page merge was capping media payloads on `page.db`, which is `createdAt DESC` (server order) — `capMediaPayloads` strips from the front of an ASCENDING list, so on this order it was stripping the newest payloads first. Split `mergeOrAppend` into an uncapped core plus the capping wrapper; per-page upserts use the uncapped one now, since the real bound already runs once, correctly ordered, in `select`.

Smaller items:
- All three comments corrected — the spinner is an in-flow sibling that does push the list down, nothing reads `total` to decide pagination, and the per-page-cap direction is fixed alongside the code.
- The "pulls exactly one older page" test asserted a negative against one hardcoded offset (`2 * PAGE_SIZE`) that a broken termination rule would never actually request — it'd ask for `PAGE_SIZE + PAGED_OLDEST.length` instead, which the test didn't check for. Replaced with a count against any offset of the route.
- Added a comment on `MESSAGE_PAGE_SIZE` naming the clamp it assumes and what happens if the server's is lower.

New test for the discard fix (leave the chat mid-fetch, come back, message survives) — couldn't find a way to unit-test the anchoring math itself without a real layout engine, so that one's inspection-only.

CI: https://github.com/JuanGalzerano/wa-galzerano/actions/runs/33322120007

rmyndharis:
Third round is close. I checked the anchoring fix rather than taking it on inspection, with your CSS and the hook's exact read/write order, 100 bubbles and 20 prepended (Chrome 146):

```
variant                 start  grew  topAtApply  final  ideal  drift
anchoring left on           1  1020        1021   2041   1021   1020
anchoring left on          20  1020        1040   2060   1040   1020
anchoring left on          99  1020        1119   2139   1119   1020
overflow-anchor: none       1  1020           1   1021   1021      0
overflow-anchor: none      20  1020          20   1040   1040      0
overflow-anchor: none      99  1020          99   1119   1119      0
```

Zero drift at every offset, so item 1 is right. Items 3 and 4 are pinned by tests that go red when the fix is reverted, which is what I look for. The termination test you replaced is a genuine improvement too: the old suite passes 11/11 against a rule that never terminates, the new one catches it.

One blocker and a few things behind it.

## Blocker: the kept queue replays a stale write on top of newer state

Dropping the discard makes the queue outlive the fetch, but the replay still only fires from the mounted hook. Between the page landing and the user coming back, other writers keep mutating the same key: the composer's reconciliation holds the `queryClient` and writes after the composer is gone, and every socket handler goes through `cachedSessionThreads`, which walks every cached thread under the session whether or not it is open. Those writes apply directly, because the query is idle by then. On return, the identity check compares only the last queued write's `wrote` against the current value, which those later writes already changed, so it concludes the queue was lost and replays it over newer state.

Traced through the production functions:

```
after optimistic send : ["db-1/sent","temp_1700000000000/pending"]
after the page landed : ["db-1/sent"]
after reconciliation  : ["db-1/sent","wamid.real/sent"]
after returning       : ["db-1/sent","wamid.real/sent","temp_1700000000000/pending"]
```

Reachable this way: open a long chat, scroll to the top so an older page starts, send a message, switch chats before the page lands. The page read lands first and clobbers the placeholder, the send round trip then reconciles straight into the idle cache, and coming back re-adds the placeholder beside the real row. Nothing ever removes a `temp_` row, so the ghost is permanent under `staleTime: Infinity` and sits on the pending clock forever. The same shape resurrects a message revoked while the user was away: the revoke handler found nothing to patch because the landed page had already dropped it, and the replay puts the body back.

This is a trade rather than a straight regression, and the half you fixed is real: the same scenario on c3a495c9 ends clean, but the write your new test covers really was lost there. What closes both is replaying when the fetch SETTLES instead of when a consumer mounts. A one-shot cache subscription registered at queue time in `writeMessagesCache`, firing when that key's `fetchStatus` returns to idle, keeps the ordering correct whether or not the chat is open, needs no unmount cleanup, and leaves your effect as a safety net. `removeQueries` at discard closes it too, but note it makes the new test fail as written, since the stub's page 0 does not carry the socket-delivered message; that test would have to assert the refetch instead.

## Turning anchoring off also removed the compensation for late-decoding media

Inline images carry no reserved box (`.chat-image-media` sets only `max-width`/`max-height`), so one above the viewport is 0px tall until it decodes and then pushes the reader down. The browser used to absorb that; now nothing does, because `onMediaLoad` returns early when the reader is not pinned and no saved restore is pending. Three 200px images:

```
anchoring              media grew  scrollTop      viewport top      reader pushed by
left on                       600  500 -> 1100    message 7 stays   0px
overflow-anchor: none         600  500 ->  500    message 7 leaves  600px
```

This one is not limited to paging: scrolling up while media above you is still decoding is a path that exists on main today. An unpinned branch in `onMediaLoad` holding position the way the older-page effect does would close it. Reserving intrinsic size on `.chat-image-media` would too, for images whose dimensions we know up front.

## The corrected position never reaches the per-chat map

`writeScrollTop` sets `programmaticWriteRef` and the scroll listener returns before its `scrollMap.set`, so after the correction the map still holds whatever the user's last real scroll saved, which is under 100 because that is where paging fires. Leave without scrolling again, come back, and `decideRestoreTarget` takes the saved branch and restores that value, roughly a page above the row being read. One line after the correction write covers it.

## The new termination assertion does not catch the rule its docstring names

The stub answers every page with `total: PAGED_NEWEST.length + PAGED_OLDEST.length`, which is 120, and the thread serves exactly 120 rows, so the pre-fix rule evaluates `120 < 120` and terminates on its own. I put that rule back and `Chats.test.ts` passes 12/12. Raise the stub's total above the rows it serves and the assertion bites: with `total: 500` it fails with expected 2, actual 3. The docstring and the matching paragraph in the commit message need the same correction. The replacement is still better than what it replaced; only the justification is wrong.

## Item 2 has no test, and one fits the file you already extended

Nothing in the suite matches `messagesError` or `loadOlderMessagesError`, and the only `.messages-loading-older` assertions are spinner presence. An `olderPageFails` flag in the fetch stub, reset in `afterEach` beside `olderPageGate`, plus one test asserting the bubbles survive a 500 and the inline failure renders where the spinner sits, is about fifteen lines. This is exactly the class of defect the suite exists to catch, and the commit message says each item got a test where one was feasible.

## Page 0 lost its payload ceiling

Round 2 ran the capping merge there, so the cache held at most `MEDIA_PAYLOAD_CACHE_LIMIT` payloads. The cap now runs only in `select`, which is derived and never written back, and page 0 is the one page with no server bound, since every live arrival lands in it. A long session in a media-heavy chat accumulates raw base64 with nothing trimming it. You are right that a front-stripping cap on a `createdAt DESC` page was worse than no cap, so the direction is correct; what went missing is the ceiling. Either give `capMediaPayloads` a direction, or store `page.db` ascending at fetch time. Nothing reads it positionally, and that second option removes the need for the split at all.

## Smaller items

- `mergeOrAppend` has no production caller now. Everything goes through `mergeOrAppendUncapped`, and on main it had real ones. The risk is the next merge path picking the shorter, more obvious name and handing it a `page.db`, which is the bug you just removed. Storing pages ascending makes this go away with it.
- Three comments still name `mergeOrAppend` on paths that no longer call it: `useChatMessages.ts:182`, `ChatComposer.tsx:234`, `Chats.tsx:64`.
- With anchoring off, the spinner's own 43px is no longer absorbed either, so the list shifts while a page loads and settles when it lands. It self-corrects, and it is nothing like the page-sized error it replaced, but an absolutely positioned spinner would avoid it and would also remove the reason for reading the height at request time.

To be clear about where this stands: the rebase, the changelog correction and the thirteen locales are all in order, CI is green on the branch, and three of the four items came back correct with the reasoning behind them right. The blocker is one ordering question in code you were invited to keep, and it has a small fix.

JuanGalzerano:
Thanks — the anchoring measurement in particular is the kind of check I have no way to run myself. Pushed `f76756dd`.

## The blocker

Went with the settle-triggered subscription. `writeMessagesCache` now registers a one-shot `queryCache.subscribe` the moment a write first queues for a key, firing `replayWritesLostToFetch` the instant that key's `fetchStatus` leaves `'fetching'` — independent of whether anything is mounted. The mount effect stays, but only as a fallback for the case a fetch had already settled before anything ever queued against it; it's a no-op there now.

Traced through why this closes it: any write that happens once a key is idle applies directly (the `fetchStatus !== 'fetching'` guard in `writeMessagesCache` returns before queueing), so it's never part of what a replay could stomp. And since the settle-triggered replay fires synchronously off the cache-notify event, it necessarily runs at-or-before the moment `fetchStatus` flips — strictly before any write that only exists because the key was already idle. There's no window left for the ordering you traced to occur.

New test drives your exact scenario: send while an older page is in flight, leave before either settles, land the page first, resolve the send second, return. Reverting just the `scheduleReplayOnSettle` call and rerunning turns it red — 2 bubbles instead of 1, the ghost placeholder back beside the real message.

## Media growth while unpinned

Added the third branch you described. `onMediaLoad` now holds position there too, against a `lastCommittedHeightRef` synced every commit rather than a request-time snapshot — a decode has no request to hang one on the way `onOlderMessagesRequested` does. Same caveat as the anchoring fix: nothing in this harness drives an actual image decode, so it's inspection-only, not test-covered.

## Corrected position never reaching the map

Added the `scrollMap.current.set` right after the correction write, same as you said. New test: scroll once (saved at 0), let the correction land, leave without scrolling again, come back — asserts the restored `scrollTop` equals the corrected value, not 0. Reverting the one line turns it red (0 !== 600 in the run that caught it).

## Termination test's total

Changed the stub's `total` to 500 (well above the 120 rows actually served) and said why in a comment at the call site — the current rule doesn't read that field at all, only the test cared, and it needed the coincidence gone to bite. Reran with the old total→rule pairing reinstated to confirm: same failure shape you got, expected 2 actual 3.

## Item 2's missing test

Added `olderPageFails`, gated the same way `olderPageGate` is (reset in `afterEach`). New test: 500 on the older-page fetch, thread stays intact, inline failure renders where the spinner sits. Reverting the `messagesError && messages.length === 0` guard turns it red.

## Page 0's payload ceiling

You're right that the split cost the ceiling. Took the second option — `page.db` is stored ascending now, reversed from the server's `createdAt DESC` at fetch time, so `mergeOrAppend` runs on it directly again with no direction parameter needed. That removes `mergeOrAppendUncapped` entirely (its only two callers were the ones I'd added), restores the ceiling as a consequence rather than a separate patch, and the three stale comments you named are accurate again by construction — nothing left calling the function they didn't mean.

## CI

Dashboard, Test, Lint, Test (PostgreSQL), Helm chart, Shell scripts all green, twice. Security audit is failing on two HIGH advisories in `browserslist` (GHSA-c83g-rgw3-j3cx, GHSA-73wf-gq98-2v4g) — a transitive dev dependency via `@nestjs/cli`→webpack and `ts-jest`→`@babel/core`, nothing this branch touches. Confirmed it's not mine: same failure on a bare `upstream/main` checkout run through the same script. A patched version (4.28.8, current is 4.28.2) exists but bumping it is a lockfile change out of scope here.

Separately — not blocking, just worth flagging since it's how I initially misread this as unrelated noise rather than a real finding: `scripts/check-audit.mjs`'s self-invocation guard (`import.meta.url === \`file://${process.argv[1]}\``) never matches on Windows, since `process.argv[1]` is a native path (backslashes, drive letter) and the comparison assumes POSIX. The script silently no-ops there — exits 0 having never run the audit at all. That's why my own `npm run check:audit` came back clean locally while CI (Linux) correctly caught the real advisory. Happy to open that as its own tiny PR if useful, kept out of this one.

`Build` and `Docker Build` show `skipped` as a consequence of the audit gate, not anything in the diff.

PR #1468, opened 2026-08-26
fix(security): remove extract-zip via a narrow @puppeteer/browsers override

raoulmusci:
## Summary

- override `whatsapp-web.js`'s Puppeteer 24 pin with Puppeteer 25.9.0
- remove `extract-zip` and the `GHSA-jmr9-qjv8-65gv` audit allowlist entry
- install the system `unzip` binary used by `@puppeteer/browsers` 3 in the amd64 image
- keep CommonJS Jest suites isolated from Puppeteer 25's ESM entry point with a throwing, test-only mock

## Context

#1306 documented the advisory and allowlisted it because directly overriding only
`@puppeteer/browsers` 3 broke the Jest suite. Upgrading the complete Puppeteer package keeps its
dependency graph internally consistent; the Jest-only module boundary handles the separate ESM/CJS
test-runner issue without changing production behavior.

## Verification

- `npm run check:audit` — no unexcused high/critical advisories
- `npm audit --omit=dev` in the production image — 0 vulnerabilities
- `npm run test:scripts` — 115 passed
- `npm test -- --runInBand --silent` — 344 suites passed, 5,933 tests passed, 2 suites/6 tests skipped
- `npm run lint` — 0 errors (3 pre-existing warnings)
- Prettier check on all changed supported files — passed
- production Docker image build — passed
- Puppeteer launched the image's pinned Chrome successfully (`Chrome/146.0.7680.31`)
- runtime container health check and authenticated MCP `SessionFindAll` smoke test — passed

Related: #1306


rmyndharis (changes requested):
Thanks for digging into this, and for the Dockerfile work in particular. The `unzip` line is correct and not obvious: `@puppeteer/browsers` 3.x shells out to the system binary (`lib/fileUtil.js:251`, `execFileAsync('unzip', ['-o', ...])`) and otherwise throws `Extraction failed: no zip archiver is available`. The Jest module boundary is the right shape too, and matches the existing mocks under `test/__mocks__/`.

The blocker is the width of the override. Puppeteer 25 removed `Browser.isConnected()`:

```
puppeteer-core 24.38.0   lib/cjs/puppeteer/api/Browser.js:189    isConnected() { ... }
puppeteer-core 25.9.0    gone; only `get connected()` remains (cdp/Browser.js:427)
```

whatsapp-web.js 1.34.7 still calls it in two places, and both sit on paths this gateway runs constantly.

**`Client.destroy()`** (`Client.js:1290`) calls it optionally:

```js
const isConnected = browser?.isConnected?.();
if (isConnected) {
    await browser.close();
}
```

On Puppeteer 25 that evaluates to `undefined`, so `browser.close()` is skipped, `authStrategy.destroy()` still resolves, and nothing throws. Session stop reaches this through `session-engine-controls.ts:276` into `wwebjs-lifecycle.ts:813`. The escalation to `forceDestroy()` at `session-engine-controls.ts:285`, which is what actually kills the Chromium process, only fires when the graceful disconnect throws. Here it returns cleanly, so every stopped session leaves a live Chromium process behind while the API reports a clean stop. That is precisely the failure the comment at `:278` guards against, arriving silently instead of as an error.

**`Client.logout()`** (`Client.js:1307`) calls it unguarded:

```js
while (this.pupBrowser.isConnected() && maxDelay < 10) {
```

That raises `TypeError: this.pupBrowser.isConnected is not a function` after the in-page unlink has already been sent, so `authStrategy.logout()` and its `fs.rm` of the profile directory never run. `wwebjs-lifecycle.ts:807` catches and rethrows, which means the unlink succeeds on WhatsApp's side while the API reports failure and the session credentials stay on disk.

Neither shows up in the verification list: a health check and `SessionFindAll` never reach `client.destroy()`.

## Suggested narrower change

Override only the package that carries the advisory, and leave Puppeteer where whatsapp-web.js pins it:

```json
"overrides": { "@puppeteer/browsers": "^3.2.0" }
```

`extract-zip` enters the tree solely through `@puppeteer/browsers`, so this clears the advisory just as well, and Puppeteer stays at 24.38.0 with `isConnected()` intact. #1306 measured that `puppeteer browsers install 'chrome@146.0.7680.31'` and `puppeteer.launch()` both work against the 3.x API, so the functional side is already established, though it is worth re-confirming on this branch.

The Jest failure that stopped #1306 from taking this route is solved by your mock: mapping `^puppeteer$` to a stub cuts the chain before `@puppeteer/browsers` and its ESM dependencies are ever loaded. Both the mock and the `unzip` line stay exactly as they are.

## Minor

`scripts/check-audit.mjs` picked up two unrelated reformats (the `execFileSync` call and the `found.set` call). This repo runs Prettier at `printWidth: 120`, which keeps both on one line. `format:check` only globs `src/**/*.ts`, `test/**/*.ts` and `**/*.md`, so CI stays green either way, but please revert them to keep the diff on topic.

## On urgency

Worth stating plainly so the scope is clear: this advisory is not reachable at runtime here. `puppeteer browsers install` runs at image build time, on amd64 only (arm64 symlinks Debian chromium, `Dockerfile:217`), against a version-pinned Chrome for Testing build fetched over HTTPS. The allowlist rule from #1306 also fails the build the moment its entry stops matching a real advisory, so the carve-out cannot outlive its cause. That makes this a cleanup rather than a fix under time pressure, and a further reason to prefer the smaller override.

Happy to merge once the override is narrowed and a session stop plus a logout have been exercised against the change.

rmyndharis:
Following up on my review with three things that have changed or that I got wrong, since two of them make the narrower override I suggested more attractive and one of them means my suggestion was not the drop-in I implied.

**1. `main` now contains a spec that Puppeteer 25 cannot load at all.**

#1463 landed `src/engine/adapters/wwebjs-protocol-timeout.spec.ts`, whose first line is:

```ts
import { Callback } from 'puppeteer-core/lib/cjs/puppeteer/common/CallbackRegistry.js';
```

`puppeteer-core@25.9.0` ships no `lib/cjs` directory. The package is `"type": "module"` and its exports map resolves both conditions to the same ESM entry:

```
'.': { import: './lib/puppeteer/puppeteer-core.js', require: './lib/puppeteer/puppeteer-core.js' }
```

So this is not a matter of repointing the import: every internal module in 25.x is ESM, the unit lane here is CommonJS ts-jest, and Jest does not transform `node_modules`. There is no path that loads a puppeteer-core internal under Puppeteer 25. On rebase, this branch turns that spec red.

That is worth weighing against the runtime regressions I described: those are silent, this one fails the build. Both point the same way, toward leaving `puppeteer` where whatsapp-web.js pins it.

**2. My suggested override was incomplete, and the gap hits your version too.**

`test/jest-e2e.json` has no `moduleNameMapper` at all, so your `^puppeteer$` mapping does not reach the e2e lane. Every AppModule-booting e2e spec pulls `engine.factory.ts` into `WhatsAppWebJsAdapter` into `wwebjs-lifecycle.ts` into `whatsapp-web.js/src/Client.js`, which requires `puppeteer` at module load. `npm run test:e2e` breaks under either override until the same mapping is added there. The mock itself is the right shape, it just needs to be registered in both configs.

**3. The Dockerfile change may not be needed at all.**

`@puppeteer/browsers@3.2.1` declares `yauzl` as an *optional peer*:

```
peerDependencies      = { yauzl: '^2.10.0 || ^3.4.0', 'proxy-agent': '>=8.0.1' }
peerDependenciesMeta  = { yauzl: { optional: true }, 'proxy-agent': { optional: true } }
```

npm does not install optional peers, which is the only reason the system binary becomes load-bearing. Its own extraction path tries the CLI first and falls back to yauzl (`src/fileUtil.ts:206`), and the error you hit spells the alternative out:

```
Extraction failed: no zip archiver is available. Install `unzip` (or `tar.exe`/Powershell on
Windows), or add the optional `yauzl` dependency.
```

So `"yauzl": "^3.4.0"` in `dependencies` is worth trying before the apt line. It covers the builder stage as well as production, adds no Debian package to the image's scan surface, and does not reintroduce the advisory: that code path does its own symlink containment check (`isInsideDirectory`, used at `fileUtil.ts:385`), which is precisely what `extract-zip` 2.0.1 lacked. If you prefer to keep apt, `unzip` needs to be in both stages rather than just `production`.

**Where that leaves the ask.** Unchanged in shape, with one correction and one addition:

- `"overrides": { "@puppeteer/browsers": "3.2.1" }`, pinned exactly rather than with a caret, since `puppeteer-core` pins its own copy exactly and a range floats a transitive under a version it was never tested against.
- Keep the mock, and register it in `test/jest-e2e.json` as well as the unit config.
- Try `yauzl` in `dependencies` in place of the Dockerfile line.
- Keep `const ALLOWLIST = []`, and revert the two reformats in `scripts/check-audit.mjs`.
- Rebase. Only `CHANGELOG.md` conflicts, on the `## [Unreleased]` heading.

To restate the scope so this does not read as urgent: the advisory is not reachable at runtime here, the image scan does not flag it, and the stale-entry rule means the carve-out retires itself the moment it stops matching. This is cleanup, and the version above is a few lines that carry none of the risk.

rmyndharis:
Thanks for tackling this. The advisory is real: `extract-zip` carries GHSA-jmr9-qjv8-65gv and it reaches the tree only through `@puppeteer/browsers`, so moving that package off the vulnerable range is the right fix.

I pushed some changes to the branch and merged `main` in so it is current. The one adjustment worth calling out is the override. Bumping Puppeteer to 25 also removes `Browser.isConnected()`, which whatsapp-web.js 1.34.7 still calls during session teardown, so a 25 bump would break session shutdown. Narrowing the override to `@puppeteer/browsers@3.2.1` alone drops `extract-zip` while Puppeteer and `puppeteer-core` stay at 24.38.0.

Around that:

- Your Jest boundary mock is exactly what keeps `@puppeteer/browsers` 3 (ESM only, pulling yargs 18 and modern-tar) out of the CommonJS test run. I registered the same mapping in `test/jest-e2e.json` for the e2e suite.
- `puppeteer` 24.38.0 is now a direct dependency, so it and `puppeteer-core` hoist to the top level the way they did before the override.
- The builder image stage skips the Puppeteer browser download, since it never launches one, and the production stage keeps your `unzip` addition, which `@puppeteer/browsers` 3 shells out to when it extracts Chrome for Testing.

`npm audit` is clean with the allowlist entry removed, and the full CI run is green. Thanks again.

PR #1472, opened 2026-08-27
feat(engine): surface archived, pinned and muted state on ChatSummary

purnamcommunity:
## Description

`GET /sessions/{sessionId}/chats` returns `ChatSummary[]`, and the archive/pin/mute actions
(`POST /sessions/{sessionId}/chats/{archive,pin,mute}`) already let a caller change each of those
states — but the chat list never reported the resulting state back. A consumer building its own
chat list (e.g. a dashboard integration wanting to mirror WhatsApp's own "hide archived chats by
default" behaviour, order pinned chats first, or honour a mute) had no way to tell which chats
were archived, pinned or muted without re-deriving it. There is no event fallback either:
docs/06-api-specification.md:961 records that whatsapp-web.js's `chat_archived` event is
deliberately not wired.

Adds `archived`, `pinned` and `muted` to `ChatSummary`, both engine adapters' mappings, the DTO,
the OpenAPI contract, and docs. Also updates the 5 hand-maintained SDK type files (JS, Python, Go,
Java, dashboard) that `check:contract-shapes` compares against the DTO.

Engine mapping:

- **whatsapp-web.js** reads the library's native `Chat.archived`, `Chat.pinned` and `Chat.isMuted`.
  `isMuted` is already the current verdict, so there is no expiry comparison to make here.
- **Baileys** reads `archived`, `pinned` and `muteEndTime` from the `IConversation` its chat store
  already ingests. `pinned` is an ORDER rather than a flag, so it is read as one. `muteEndTime` is
  epoch milliseconds — measured, per `chat-mute.spec.ts:14-18`, and the same unit
  `mute-chat.dto.ts:19-24` documents on the way in — and is compared against now.

`muted` is the verdict rather than the expiry: that is what a caller needs to label a mute/unmute
control, and it is what both engines answer directly. The expiry instant itself is tracked
separately in #1473.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

No route or behaviour changes, but the shape grew rather than this being purely additive: the three
fields are required, so hand-built `ChatSummary` values (fixtures, mocks, stub gateways) have to
supply them. Recorded under `### Changed` in the CHANGELOG alongside the `### Added` entry.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

N/A — engine/contract-level change, no UI in this repo.

## Related Issues

Coordinates with #1477 / #1473 on the `muted` / `muteExpiration` split.


rmyndharis (changes requested):
The `archived` mapping is right on both engines, and the gap it closes is real: `POST /chats/archive` ships with no read counterpart, and docs/06-api-specification.md:961 records that whatsapp-web.js's `chat_archived` event is deliberately not wired, so there is no event fallback either. whatsapp-web.js derives `Chat.archived` from `data.archive` on a model built from the full chat collection, and Baileys carries `archived` on the `IConversation` the store already ingests. That part I would take as it stands.

Three things are in the way.

## The diff is wider than the title and the body

The body says "Purely additive, no existing field, route, or behavior changes." That does not match what is here:

- `ChatSummary` gains `pinned` and `muted` as well as `archived`.
- `setGroupDescription` on whatsapp-web.js becomes an unconditional `EngineNotSupportedError` (`wwebjs-groups.ts:328-344`), with the matrix cell flipped from `supported` to `not-available` (`engine-capability-matrix.ts:331-336`).

`pinned` is a fair rider: one line per producer, same shape, nothing extra to reason about. The `setGroupDescription` demotion is not. It turns `PUT /sessions/:sessionId/groups/:groupId/description` (`group.controller.ts:396`) from 200-or-`EngineRefusedError` into a flat 501 on one of two first-class engines, on a route the SDKs and the agent tool surface both expose. Nobody reading a PR titled "surface archived state on ChatSummary" will review that.

The measurement behind it is good and I want the change: testing both with `descId` undefined and defined, and using `setGroupSubject` as a control on the same group in the same session, is exactly how to establish it. Please move it to its own PR carrying, at minimum:

- the `docs/29-engine-capability-matrix.md` row at :310 and the wwjs exposure cell at :664
- the totals at :389 and :950, plus the 29.8 restatements
- an `@ApiResponse({ status: 501 })` on `group.controller.ts:396` and the matching docs/06 error line
- its own CHANGELOG entry

## Two jobs are red as pushed

`src/engine/docs-29-counts.spec.ts` recounts the figures from `engineCapabilityMatrix()` at runtime and matches them against every phrasing in docs/29. Flipping one cell without touching the document moves the count from the `199 ✅, 25 ❌` stated at docs/29:389 to 198/26, so the `test` job fails.

Making the three fields required is a breaking type change for anyone holding a `ChatSummary`, and the dashboard proves it inside this repo: `dashboard/src/pages/Chats.test.ts` builds two `Chat` literals (:32 and :45) that no longer typecheck once the fields are required, and the `dashboard` job runs `npm run typecheck`.

I am not asking you to make them optional. Required is the right call and it matches the file's own precedent. But those fixtures need the fields, and the CHANGELOG should say the shape grew rather than filing it under "Added" alone.

## `muted` overlaps #1477

#1477 adds the same fact as `isMuted?`, closing #1473. I do not want it landing twice under two names and two optionality rules, so the split is: this PR keeps `muted`, and #1477 reduces to `muteExpiration` on top of it. I am saying the same in that thread.

One change comes with owning the field. Drop the magnitude guess in the Baileys helper:

```
raw < 1e12 ? raw * 1000 : raw
```

The unit is settled, not ambiguous. Baileys fills `muteEndTime` from `toNumber(action.muteAction.muteEndTimestamp)`, and this gateway writes it through `chatModify({ mute: muteUntil })` where `muteUntil` is documented epoch milliseconds in `src/modules/session/dto/mute-chat.dto.ts:19-24`. A millisecond stamp below 1e12 is a date before 2001, so the first branch is unreachable and the comment justifying it points at a question the repo answers in two places.

## Rebase

`main` has moved twice since you branched. Only `CHANGELOG.md` conflicts, and only on the `## [Unreleased]` heading: drop yours and add bullets under the existing one. The code files merge clean.

purnamcommunity:
Force-pushed as `10e2f3d8`.

**`setGroupDescription`** — removed from this PR. It will follow separately, carrying the docs/29 row at :310 and the wwjs exposure cell at :664, the totals at :389 and :950 with the §29.8 restatements, an `@ApiResponse({ status: 501 })` on `group.controller.ts:396`, the matching docs/06 error line, and its own CHANGELOG entry. Removing it here restores the `199 ✅, 25 ❌` that `docs-29-counts.spec.ts` asserts, so that job passes without docs/29 changes in this PR.

**Mute unit** — the `< 1e12` branch is gone and `muteEndTime` is read as milliseconds. `chat-mute.spec.ts:14-18` records the measurement: a seconds-scale value sent through `chatModify({ mute })` left the chat unmuted, while the same instant in milliseconds muted it to the expected minute. `mute-chat.dto.ts:19-24` documents the same unit inbound. The claim appeared in four places — the `ChatSummary` interface doc, the DTO description, docs/06, and the generated `openapi.json` — all corrected. The test covering the removed branch is replaced by one pinning the millisecond reading.

**Dashboard fixtures** — both `Chat` literals in `dashboard/src/pages/Chats.test.ts` (:32, :45) now set the three fields, so `typecheck` passes. The fields stay required. The CHANGELOG records the shape change under `### Changed` alongside the `### Added` entry.

**`muted` / #1477** — this PR keeps `muted`. The DTO, interface and docs point at #1473 for the expiry instant, so `muteExpiration` landing on top continues rather than duplicates it.

**Rebase** — onto current `main`. Only the `## [Unreleased]` heading conflicted; the heading is dropped and the bullets sit under the existing one.

---

Added beyond the review: `src/engine/adapters/wwebjs-chat-state.spec.ts`. The whatsapp-web.js mapping had no test coverage — only the Baileys mappers did. The two wwjs summary builders are independent (`getChats` maps the library `Chat` model, `getChatsByLabel` maps what `getChatsByLabelId` returns), so a field added to one and missed in the other would not surface until a Business account listed a label. The spec also covers an unset field landing as `false` rather than escaping `undefined` into a required boolean, and mute reading `Chat.isMuted` rather than the `muteExpiration` stamp behind it.

Open question: #1477 states that Baileys uses a negative sentinel on `muteEndTime` for an indefinite mute. If that holds, `isMuted` currently reads an indefinite mute as expired, since a negative value fails the `> Date.now()` comparison. I found no evidence for it in the proto or in this repo's measured notes, so I left it unchanged rather than encode another assumption. It can be handled here or in #1477.

Locally: `test` (5950), `test:docs` (264), dashboard `typecheck` and tests (332), `lint`, `format:check`, `check:contract-shapes` (324 pairs), `openapi:check`. CI runs on this PR sit at `action_required`.

rmyndharis:
Thanks, this covers everything I asked for. Checking the branch at `10e2f3d8`: the `setGroupDescription` demotion is gone from the whole history, not just netted out at the tip (`wwebjs-groups.ts`, `engine-capability-matrix.ts`, `docs/29`, `group.controller.ts` and `engine-parity.spec.ts` all hash-match `main`, and no commit on the branch touches them), the magnitude guess is gone with a test that pins the unit, both dashboard fixtures supply the fields without a cast or a widening, and the rebase is linear under a single `[Unreleased]` heading. `test:docs` is back to 264 passing, so the capability-count gate is no longer in the way.

Adding `wwebjs-chat-state.spec.ts` unprompted was the right call. `getChatsByLabel` is exactly the builder that drifts.

Two things before this goes in.

## `LabelChatDto` did not follow the payload

`wwebjs-labels.ts:77-79` now emits `archived`, `pinned` and `muted`, but `LabelChatDto` (`src/modules/label/dto/label-response.dto.ts:27-48`) still ends at `lastMessage?`, and `label.controller.ts:51` declares the route as `type: [LabelChatDto]`. The committed `openapi.json` therefore publishes `LabelChatDto` with 7 properties against `ChatSummaryDto`'s 10, so `GET /sessions/{sessionId}/labels/{labelId}/chats` now returns three fields its own schema does not describe. `docs/06-api-specification.md:3868` says that route returns "the same shape `GET /sessions/:sessionId/chats` returns", which this diff makes untrue.

Nothing catches it: `check:contract-shapes` maps only `ChatSummary` to `ChatSummaryDto`, and `openapi:check` passes because the snapshot faithfully mirrors the un-widened DTO.

Either add the three fields to `LabelChatDto` and re-run `npm run openapi:export`, or point `label.controller.ts:51` at `ChatSummaryDto` and delete the duplicate. I would take the second: the DTO's own comment already calls it the chats-list projection. The CHANGELOG entry should name the label route as well, since its payload widened too.

## The new tests pass without testing the mapping

`wwebjs-chat-state.spec.ts:51` sets all three fields true together and `:59` sets all three false together, so no field is pinned to its own source. Swap lines 63 and 64 of `wwebjs-chats.ts`, making `archived` read `chat.pinned` and `pinned` read `chat.archived`, and all 1651 tests under `npx jest src/engine` stay green. On the Baileys side, replacing `archived: c.archived ?? false` at `baileys-session-store.ts:391` with a constant `false` leaves the full 5950-test suite green, because no Baileys case asserts `archived: true` at all.

One mixed-value case per mapper closes both: archived true with pinned false, and the reverse.

While you are in there, two of the new Baileys assertions are duplicates rather than the cases they read as. `chatFor` (`baileys-session-store.spec.ts:46-49`) reuses one store across the calls in a test, and `upsertChats` merges rather than replaces (`baileys-session-store.ts:138-139`), so `chatFor({}).pinned` at `:56` re-reads the `pinned: 0` set on the line above, and `chatFor({}).muted` at `:65` re-reads `muteEndTime: 0`. Neither reaches the absent-field path its placement implies. A fresh store per case fixes it.

## A question on the mute unit, since you raised the sentinel yourself

The millisecond reading is measured on the write path: `chat-mute.spec.ts` round-trips through `chatModify({ mute })`, which is app-state `MuteAction.muteEndTimestamp`. The read at `baileys-session-store.ts:394` is also fed by `messaging-history.set`, whose records carry `Conversation.muteEndTime`, a different field whose unit nothing in the repo establishes, and two lines above it the same message's `conversationTimestamp` is read as seconds. If history-sync stamps are seconds, every phone-muted chat reports `muted: false` after a reconnect, and the new test would pass anyway.

Your negative-sentinel question resolves the same way. One capture settles both: mute a chat from the phone, once with a duration and once indefinitely, reconnect the session so the history sync lands, and read the raw `muteEndTime` off the store. If you can run that, paste the raw values. If not, say so and I will run it on my side.

Neither answer has to block the merge, but I would rather know before the field ships than after, since `muted` is the whole point of the field.

One caveat on all of the above: CI has never run on this PR. There are no check runs on `10e2f3d8` at all, so everything green I quoted is a local run on my machine, not a passing build. Worth keeping in mind if anything here disagrees with what you see once it does run.

Issue #1474, opened 2026-08-28
[Feature]: Proxy option to route traffic through


vitusan:
### Pre-flight

- [x] I searched existing issues and this isn't already requested

### Problem / motivation

We all know how meta is, and for multiple projects running on the same server hosted on a big and cheap cloud provider, things get a little messy when dealing with the servers ip addresses. 

A simple proxy option to route every traffic through it suffice for most users. 

### Proposed solution

There is an example already built over at Evolution Api making this rather "easy" to port to this project since both are TS based codebases.

I can perform this modification myself, but sadly, i lack the time at this particular moment in my life. But if anyone here could implement this, things would be awesome and smooth for this awesome project.

### Alternatives considered

For large scale projects, the alternatives are using other projects, but they lack the beautiful integration of n8n that this project has and also lack the active update rhythm that this one has.

### Scope

- [x] I understand some features are limited by the underlying WhatsApp engine (e.g. interactive Buttons/List messages are not supported on whatsapp-web.js).

vitusan:
Update, this is for the UI, i know we can set the proxy configuration of the session through the api, but the UI option is a cool addition for the less tech savvy user.

vitusan:
Well, heck why not, here it is: https://github.com/rmyndharis/OpenWA/pull/1475.

PR #1475, opened 2026-08-28
feat(sessions): add per-session proxy configuration in dashboard UI

vitusan:
## Summary

- Add `GET`/`PATCH /api/sessions/:sessionId/proxy` for masked proxy read/update (credentials never returned; `proxyType` derived from URL scheme on read)
- `PATCH` accepts `proxyUrl` only — no inert `proxyType` write field; changes apply on the next session start
- Dashboard: optional proxy fields in create-session modal, Proxy button + edit modal on session cards (details from `GET /proxy`, not the list payload)
- Regenerate `openapi.json`, extend all five SDKs with `getProxy`/`updateProxy`, and align contract-shape gates

## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] Tests added/updated
- [x] Documentation updated
- [x] Lint passes (changed files)
- [x] Self-reviewed

## Test plan

- [x] `npx jest src/modules/session/dto/session-proxy.dto.spec.ts`
- [x] `npx jest src/modules/session/session.controller.spec.ts -t proxy`
- [x] `npx jest src/modules/session/session.service.spec.ts -t "session proxy"`
- [x] `cd dashboard && node --test src/pages/Sessions.test.ts`
- [x] `npm run openapi:check`
- [x] `npm run check:contract-shapes`
- [x] Manual: create session with proxy, edit via Proxy modal, verify settings apply on next start

## Related Issues

#1474 (https://github.com/rmyndharis/OpenWA/issues/1474)


rmyndharis (changes requested):
The gap here is real and there is no way around it from the API today. `proxyUrl` is write-once: `create-session.dto.ts` accepts it, and `session.controller.ts` then offers no proxy route and no general session-update route, so changing or clearing a proxy means deleting the session and pairing the WhatsApp account again, or editing the database by hand. A read-and-edit pair is the right answer, the masking approach is sound, and the backend logic reads correctly.

Please link #1474 in the body. You filed it half an hour before this PR and it is the motivation; "Related Issues: N/A" hides that.

Four things before this can land.

## The OpenAPI snapshot was not regenerated

This adds two routes with `@ApiOperation`/`@ApiResponse`, two DTO classes (`SessionProxyResponseDto`, `UpdateSessionProxyDto`) and three `@ApiProperty` fields on `SessionResponseDto`, but `openapi.json` is untouched. That fails `npm run openapi:check`, which diffs the committed snapshot against a fresh export.

It fails a second gate for the same reason. You added `proxyEnabled`/`proxyType`/`proxyHost` to all five hand-written wire types, but `Session`/`SessionResponse` to `SessionResponseDto` is a live mapped pair in `scripts/check-contract-shapes.mjs`, and that gate compares in both directions. With the fields in the hand types and absent from the contract, it reports five failures.

`npm run openapi:export` and committing the result clears both.

## `proxyType` does not do anything

Nothing reads it. Baileys picks its agent from the URL scheme (`baileys-lifecycle.ts:44-50`, `createProxyAgent(proxyUrl)`), and whatsapp-web.js goes through `buildProxyLaunchConfig(url)` in `wwebjs-proxy.ts:31`, which also takes only the URL. `engine.factory.ts:240` and `builtin/whatsapp-web-js/index.ts:72` set `type: proxyType ?? 'http'` on the adapter config, and no code anywhere reads `config.proxy.type`.

That has two consequences here. The dropdown in the modal is an inert control that returns 200 and repaints the card. And `updateProxy` writes a guess into the column:

```ts
if (nextUrl) {
  if (!nextType) nextType = 'http';
}
```

The create path deliberately stores `dto.proxyType || null` (`session.service.ts:275`), so after a PATCH a session with `socks5://host:1080` and no explicit type is persisted and then displayed as `http`. That is a wrong label on a card, for a field that changes no behavior.

My preference is to drop `proxyType` from the new surface entirely and derive the display label from the URL scheme in `projectSessionProxy`. That removes the inert control, the guessed default, one field from `openapi.json`, and one field from five SDK type files. Worth settling now: once the snapshot ships the field across five clients, removing it is a breaking change.

## The three fields on `SessionResponseDto` need an explicit decision

`session-response.dto.ts:97-100` states a flat invariant: `fromEntity` strips `config`, `proxyUrl` and `proxyType` as "sensitive engine config fields that must not appear in any API response." This PR puts `proxyType` and a derived `proxyHost` back onto that response.

A masked projection is arguably a different thing from the raw value, and I am open to that argument. But it cannot stay implicit while the docblock two lines up says the opposite. Either narrow the invariant in that comment to the credential-bearing values it is really about, or drop the three fields and let the card read from `GET /proxy` the way the modal already does. Note that the modal fetches on open regardless, so today those fields buy exactly one card row.

## Minor

- Fifteen files changed mode from `100644` to `100755`: every touched file under `docs/` and `sdk/`. Please revert.
- No CHANGELOG entry, and this is the largest user-visible change of the open set.
- Please run `npm run format` in both roots and `npm run lint` before the next push, so the Lint job has a clean run to report.

This merges clean against `main` as it stands, so no rebase is needed yet.

rmyndharis:
You addressed all four items and the hard part came back right: the masking is sound. `projectSessionProxy` reads `parsed.host`, which drops userinfo, the validator message does not echo what was typed, the audit row carries only the masked triple, and nothing logs the raw URL on any path. `updateProxy` writing `proxyType: null` rather than guessing `http` is better than what I asked for. The security row you added to docs/04 matches what the code actually does.

Rather than send this back for a fourth round I have pushed to your branch. Everything below is on it. Push again on top; if you disagree with any of it, say so and I will revert that piece.

## Why you never saw CI

Not your fault, and worth knowing. This PR has had **zero** workflow runs since 28 August. Not queued, not failing: none created. Once #1463 and #1480 landed, the branch conflicted with `main` on `CHANGELOG.md`, and GitHub cannot build the merge ref for a conflicting PR, so the `pull_request` event produced no run at all. Your 31 August push was invisible to CI for that reason.

I merged current `main` in rather than rebasing, so nothing you have locally is invalidated. Runs appeared immediately afterwards and I have approved them.

## Two ways the modal could destroy a proxy

Both are the same shape: the form defaults to "off" and Save treats "off" as "clear it".

**A failed `GET /proxy` rendered as "no proxy configured".** The effect resets `proxyInfo` to null and `proxyEnabled` to false before fetching, and the catch restored neither, so a read failure looked identical to a session that has no proxy. Pressing Save then sent `proxyUrl: null` and destroyed a URL and its stored credentials that the operator was never shown. Reproduced against your own test harness:

```
toggle shows proxy as enabled? false   (stored proxy: enabled true, hasCredentials true)
PATCH sent on Save: {"proxyUrl":null}
```

The modal now shows a load failure with no form and no Save button when the read fails.

**Any toast reset the open modal.** The effect listed `t` and `toast` as dependencies. `ToastContext` hands out a fresh object literal on every provider render, and toasts auto-dismiss on a 4 second timer, so a toast expiring while the modal was open re-ran the effect and wiped the URL mid-typing. The deps are now `[proxySessionId]` with the same eslint-disable the file already uses.

Both are pinned by tests that go red when the fix is reverted; I checked that rather than assuming.

## The gates that were red

- **Dashboard typecheck and build.** `createProxyInvalid` produced `boolean | ""`. `isValidProxyUrl` already rejects an empty string, so the emptiness check was redundant and removing it fixes the type and the expression together.
- **Dashboard i18n, all twelve non-English locales.** `en.json` renamed `dashboard.columns.proxy*` to `sessions.card.proxy*` and the others kept the old names. Chasing the rename would have been wrong: nothing reads either spelling and `.session-proxy` matches no element, so the card indicator those strings were for was never built. I deleted the dead strings and the dead CSS rule instead.
- **Python SDK mypy.** `SessionProxy.proxyType` was annotated `ProxyType`, a name the Python SDK never defines. It now spells the union the way the same file spells it for `CreateSessionRequest`.
- **Go SDK gofmt.** The struct's name column was padded one space wide. I have no Go toolchain here, so this was the one change I could not check locally; the Go job on the new head confirms it, along with the marshalling change below.

## Contract and clients

`openapi.json` published both new string fields as `type: object`, because neither `@ApiProperty` passed an explicit `type` and `string | null` reflects as `Object`. That misleads every generated client, and it also disarmed `check:contract-shapes`, which compares field types across all five: the gate was green because the contract claimed a shape nothing could disagree with. With the types corrected the gate immediately found a real mismatch in the Go client, which is the next item.

Go and Java both documented `proxyUrl: null` as the way to clear a proxy and neither could send it. Go had `omitempty` on the pointer, which omits a nil instead of emitting null; it now follows `UpdateSessionConfigRequest`, whose comment twelve lines below describes this exact trap. Java routed only two body types through the `serializeNulls` Gson. The PHP docblock advertised a `proxyType` field the DTO no longer accepts.

## Two calls I made that are worth your objection

**`PATCH /proxy` now requires an unscoped key.** Before this route existed, `proxyUrl` could only be set through `POST /api/sessions`, which carries `@RequireUnscopedKey`, so a key restricted to specific sessions could never configure a proxy. Matching `PATCH /config` instead would have widened that reach as a side effect of adding a read and edit pair. Redirecting a session's entire egress through a chosen host is a deployment-level act, so I kept the reach where it was and documented it in docs/06.

**The proxy type dropdown is gone from the create modal.** Nothing reads the column it wrote, and the first PATCH through your new endpoint nulls it, so it could only ever mislabel a socks5 proxy as http. The create API still accepts `proxyType`, so this is not a breaking change.

Smaller things while I was in there: the proxy validity check moved into `handleCreate`, because the name field's Enter key calls it directly and a button-only guard let Enter create a session with no proxy while the toggle said on; proxy fields reset whenever the create modal closes, so a cancelled create no longer keeps a credentialed URL in memory; the load failure toast says load rather than save; and the changelog entry now mentions the dashboard, which is the headline of your PR and was missing from it.

One thing I deliberately left alone: the card still renders `proxyType ?? 'http'`, which would mislabel a proxy whose scheme is not one of the four. Both validators reject those, so it is only reachable through a hand-edited database, and I would rather you decide how to show it than have me guess.

Thanks for working through three rounds of this. The endpoint pair is a real gap closed, and the write-once proxy has bothered people for a while.

rmyndharis (approved):
All sixteen checks green on 53be47c1. The four items from the first round came back correct, and the rest is on the branch with the reasoning in the comment above.

PR #1477, opened 2026-08-29
feat(session): expose chat mute state in the chats list (#1473)

### LINKED ISSUE #1473: [Feature]: Expose chat mute state (isMuted / muteExpiration) in the chat DTO
### Pre-flight

- [x] I searched existing issues and this isn't already requested

### Problem / motivation

The API can mute a chat (`POST /sessions/{id}/chats/mute`) but never
reports whether a chat *is* muted. `GET /sessions/{id}/chats` returns:

id, name, isGroup, kind, unreadCount, timestamp, lastMessage

There is no way to read mute state back.

This blocks a common automation rule: "never auto-reply in a chat the
user has muted." Muting is how a user says "I don't want noise from
here" — an assistant that answers there defeats the mute. Today that
rule cannot be enforced, only promised.

Version: OpenWA v0.23.3, ENGINE_TYPE=whatsapp-web.js

### Proposed solution

Add two optional fields to the chat DTO returned by
`GET /sessions/{id}/chats` and `GET /sessions/{id}/chats/{chatId}`:

    isMuted?: boolean
    muteExpiration?: number   // epoch seconds, 0 = muted indefinitely

Optional keeps it non-breaking, and lets an engine that can't report it
return `undefined` rather than a misleading `false`.

The data is already available — whatsapp-web.js exposes it on the Chat
object. In `node_modules/whatsapp-web.js/index.d.ts`:

    1873:  isMuted: boolean;
    1908:  Promise<{ isMuted: boolean; muteExpiration: number }>

So this is a mapping in the wwebjs adapter plus two fields on
`ChatSummaryDto`, not new engine work.

Distinguishing "unknown" from "not muted" matters for the use case: a
consumer enforcing "never reply in muted chats" should be able to fail
closed when the engine can't answer.

### Alternatives considered

1. **Call `POST /chats/mute` and read the response.** Non-starter: it's a
   write. You'd have to change the state to learn it.

2. **Keep a mute list on the client side.** Drifts immediately — the user
   mutes from their phone and the external list never finds out. It also
   puts the burden of a WhatsApp-owned fact on every consumer.

3. **Reach into the engine adapter directly, bypassing the API.** Breaks
   the abstraction and doesn't work when the API runs in its own
   container, which is the documented deployment.

4. **A dedicated `GET /chats/{chatId}/mute` endpoint instead of DTO
   fields.** This would work and I'd take it. I proposed the DTO fields
   because the main use case scans a chat list to decide where an
   automation may act — with a separate endpoint that's one extra round
   trip per chat.

### Scope

- [x] I understand some features are limited by the underlying WhatsApp engine (e.g. interactive Buttons/List messages are not supported on whatsapp-web.js).

usmancynosure:
Closes #1473.

## What

Adds two optional fields to the chat summary returned by `GET /sessions/{sessionId}/chats`:

- `isMuted?: boolean` — current mute state; left `undefined` when the active engine cannot report it, so "unknown" stays distinct from "not muted" (a consumer enforcing "never reply in a muted chat" can fail closed).
- `muteExpiration?: number` — Unix **seconds** the mute ends (same unit as `timestamp`), `0` = muted indefinitely; only set when muted.

Optional + additive, so it is non-breaking.

## Why

The API could mute a chat (`POST /chats/mute`) but never report whether a chat *is* muted, which blocks the common automation rule "never auto-reply in a chat the user muted".

## Engine mapping

Both engines normalise to the same seconds / 0-indefinite contract:

- **whatsapp-web.js**: from `Chat.isMuted` / `Chat.muteExpiration` (already epoch seconds; the `-1` "forever" sentinel is normalised to `0`).
- **baileys**: from `Chat.muteEndTime`, which is epoch **milliseconds** (per the measured note in `chat-mute.spec.ts`) with a negative sentinel for indefinite — converted to seconds, and treated as unmuted once the end instant has elapsed.

Controller and service pass the shape through unchanged.

## Tests / checks

Unit tests added for both adapters' mappings; OpenAPI snapshot regenerated. Locally green: `tsc --noEmit`, full unit suite (5940), e2e (181), `test:docs` (engine-parity / changelog / OpenAPI contract), lint, `openapi:check`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- `GET /sessions/{sessionId}/chats` now reports each chat's mute state: `isMuted` and, when muted,
+  `muteExpiration` (Unix seconds, `0` = muted indefinitely). Both fields are optional and absent when the
+  active engine cannot report mute state, so a consumer can tell "unknown" apart from "not muted" — e.g. an
+  automation rule that must never reply in a muted chat. Mapped on both engines. Closes #1473.
+
 ## [0.23.3] - 2026-08-24
 
 ### Added
--- a/openapi.json
+++ b/openapi.json
@@ -10246,6 +10246,16 @@
           "lastMessage": {
             "type": "string",
             "example": "hi"
+          },
+          "isMuted": {
+            "type": "boolean",
+            "description": "Whether the chat is currently muted. Absent when the active engine cannot report mute state (so a consumer can tell \"unknown\" apart from \"not muted\").",
+            "example": true
+          },
+          "muteExpiration": {
+            "type": "number",
+            "description": "Unix seconds at which the mute expires (same unit as `timestamp`). `0` means muted indefinitely. Only present when `isMuted` is true.",
+            "example": 1700003600
           }
         },
         "required": [
--- a/src/engine/adapters/baileys-session-store.spec.ts
+++ b/src/engine/adapters/baileys-session-store.spec.ts
@@ -64,6 +64,7 @@ describe('BaileysSessionStore', () => {
         unreadCount: 2,
         timestamp: 200,
         lastMessage: 'newest',
+        isMuted: false, // no muteEndTime on the chat -> not muted
       },
     ]);
     expect(store.lastMessage('628111@s.whatsapp.net')).toEqual({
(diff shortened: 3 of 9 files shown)


rmyndharis:
Thanks for this, and for claiming #1473 in the open before starting. The work is sound: the mapping is clean on both engines, the OpenAPI snapshot is genuinely regenerated rather than hand-edited, and the tests cover the two cases that actually matter (the `-1` sentinel and a mute whose end has already passed). I want those tests.

I am holding this on a decision that is mine, not on anything you did wrong, so let me state it plainly rather than leave you guessing.

## The collision

#1472 also adds mute state to `ChatSummary`, as a required `muted: boolean`, alongside `archived` and `pinned`. Two open PRs are adding the same fact under two names and two optionality rules, and whichever lands second rewrites the same hunk in `whatsapp-engine.interface.ts`, `chat-summary.dto.ts`, `wwebjs-chats.ts`, `baileys-session-store.ts` and `openapi.json`.

I am landing #1472 first. Not on merit between the two, but because it already carries the five hand-written SDK type files that `check:contract-shapes` requires and covers all three `ChatSummary` producers, so it is the cheaper of the two to rebase onto.

That leaves this PR the piece #1472 explicitly declined to do: `muteExpiration`. Drop `isMuted` (it arrives as `muted`), keep your two adapter mappings for the expiry, keep your tests. That is roughly 40 lines instead of 156, and the part that is genuinely yours stays yours. I am asking #1472 to drop its magnitude guess on the Baileys unit and read plain milliseconds, which is your reading and the correct one.

## Three things to fix while you are in there

**The optionality does not do what the description promises.** The stated reason for `isMuted?` is that a consumer can tell "engine cannot report" from "not muted" and fail closed. But neither producer can ever omit it: `wwebjs-chats.ts` sets `const isMuted = Boolean(chat.isMuted)` unconditionally, and `neutralMuteState` returns `isMuted` on all four branches. The `undefined` state the JSDoc, the DTO description, the OpenAPI entry and the CHANGELOG all advertise cannot occur, so anyone writing the documented fail-closed check is writing dead code. This resolves itself once `muted` is required on #1472, but it is worth knowing why.

**A third producer is missed.** There are three: `baileys-session-store.ts:377`, `wwebjs-chats.ts:26`, and `wwebjs-labels.ts:42` (`getChatsByLabel`, behind `GET /sessions/{id}/labels/{labelId}/chats`). This PR updates two. Because the field is optional, `tsc` does not complain, so that route quietly returns chats with the field absent, which under this PR's own doctrine means "the engine cannot report mute state" for whatsapp-web.js. That is exactly the cost of the optional convention, and it is why #1472's required fields caught the same producer automatically.

**The unit should be milliseconds, not seconds.** I know `timestamp` on the same object is seconds, and that is a fair argument. The stronger one runs the other way: this is the read half of `POST /sessions/{id}/chats/mute`, whose `muteUntil` is epoch milliseconds, and `mute-chat.dto.ts:19-24` spells out the failure mode in the schema description, that a seconds-scale value is an instant in 1970 so the mute expires immediately while the request still answers 200. A caller that reads an expiry and writes it back hits precisely that trap. Please switch to milliseconds, keep `0` for indefinite, and name the unit in the field description with a cross-reference to the mute endpoint. I will add the reciprocal clause on `MuteChatDto`.

## Gates

`check:contract-shapes` fails today: `ChatSummary` to `ChatSummaryDto` is a mapped pair on all five clients and the gate compares in both directions, so the fields need to reach `sdk/javascript/src/types.ts`, `sdk/python/openwa/types.py`, `sdk/go/types_chat.go`, `sdk/java/.../ChatSummary.java` and `dashboard/src/services/api.ts`. After #1472 lands those files will already have the pattern to copy.

You will also need a rebase. Only `CHANGELOG.md` conflicts, on the `## [Unreleased]` heading.

I will comment here once #1472 is in so you are not rebasing against a moving target.

rmyndharis:
Thanks for this, and for claiming #1473 in the open. The mute-state feature has landed, just not through this branch: #1472 added `muted` (with `archived` and `pinned`), #1531 made it survive a reconnect (WhatsApp does not re-deliver mute state on reconnect, so it is now persisted per chat and rehydrated on boot), and #1533 adds `muteExpiration` on top, in milliseconds so it round-trips with `POST /chats/mute`. Your reading of the Baileys unit shaped that work and #1533's changelog credits you. `main` has moved a long way under this branch (the chat summary, the label route and the store were all rewritten), so I am closing this rather than ask you to rebase onto a diff that no longer overlaps. Thank you.

Issue #1476, opened 2026-08-29
[Bug]: Retrieving the contact list is not working at all. 500 - Internal server error


onepay-ye:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm on the latest released version (or noted my version below)

### OpenWA version

0.23.3

### Deployment

Docker Compose

### Database

SQLite (default)

### WhatsApp engine

whatsapp-web.js

### What happened?

Hello,

I have noticed that in recent versions, from 10 onwards up to the current release, we can no longer fetch contact lists via the endpoint:

**/api/sessions/{sessionId}/contacts**

When fetching the contact list, it returns a 500 - Internal server error.

Any Help ,

### Steps to reproduce

returns a 500 - Internal server error

### Relevant logs

```shell

```

rmyndharis:
Thanks for the report. I need two things to place this, because a 500 from that route can come from several different causes and the HTTP body will not distinguish them.

**1. The container log around the failure.** The response body carries nothing useful, but the gateway logs the underlying error. Please run the request and then:

```bash
docker compose logs --tail=200 <service-name>
```

and paste the lines around the 500. The first line of the stack is what separates the possibilities.

**2. Roughly how many contacts are on that account**, and whether `GET /api/sessions/{sessionId}/chats` works on the same session.

## The most likely cause, and something to try now

`GET /contacts` reads the entire contact store in a single browser command; the `limit`/`offset` parameters trim the response afterwards, in the gateway, not in the browser. Puppeteer gives every browser command a fixed budget, 180 000 ms by default, and on a large account that read can run past it. The command is dropped while the renderer is still working, which is why the session stays `ready` and everything else keeps working, and why it reproduces every single time on the affected account while smaller accounts on the same deployment are fine. That matches "not working at all" for this one endpoint.

If that is what is happening, the log line will name a CDP method and the setting, something like:

```
Runtime.callFunctionOn timed out. Increase the 'protocolTimeout' setting in launch/connect
calls for a higher timeout if needed.
```

The budget is now configurable:

```bash
PUPPETEER_PROTOCOL_TIMEOUT_MS=300000
```

One caveat on availability: this landed on `main` and is not in 0.23.3, so it will reach you in the next release. If you are able to run from `main` before then, setting it is the fastest way to confirm or rule out this theory.

If the log shows something else, paste it and I will take it from there. Please keep this open either way, since the diagnosis is not settled yet.

[...]

PR #1478, opened 2026-08-30
fix(contacts): read ids via readWid and skip unreadable entries with a count

Deyvis17GY:
## What

Rework the whatsapp-web.js contacts mapping along the lines suggested in review.

- Fold getContacts and getContactById (identical six-field mapping) into one
  `toContact(c)` helper that reads the id through `readWid`, so the
  `_serialized`->`$1` rename lands on both reads; returns null for an unreadable id.
- getContacts skips unreadable entries, counted and logged
  (`Skipped N contact(s) without a serialized id`), matching wwebjs-chats.ts,
  instead of dropping the address book silently.
- Fixes getContactById returning 404 for a contact whose id only differs by the rename.
- Convert getContacts' dead-page rethrow to `EngineTransportError` (503) to match
  the sibling read in wwebjs-chats.ts.

## Note

Reframed to "Refs #1476": the page-level cause isn't established yet. Happy to split
the transport-error change into its own PR or add a unit test if preferred.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - A whatsapp-web.js protocol timeout is no longer eligible to be classified as a dead page.
   Behaviour is unchanged on the current Puppeteer; the guard keeps a future bump from reporting a
   slow command as a transport death.
+- `GET /sessions/{sessionId}/contacts` declares `503` in the contract and answers it when the
+  whatsapp-web.js page dies mid-read, instead of a bare `500`. Thanks @Deyvis17GY.
 
 ### Fixed
 
@@ -37,6 +39,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and the library still calls it positionally, so `widToGroupJid` threw inside the page. A new
   install-time patch (🔧⁹, docs/29) sends the options object; an empty description still clears.
   `setGroupSubject` was never affected and Baileys is unchanged. Thanks @purnamcommunity.
+- whatsapp-web.js contact reads resolve the renamed `$1` serialized-id field, so contacts keep their
+  `id` on a WhatsApp Web build that renamed it; an entry with no readable id is skipped and counted
+  in the log. Thanks @Deyvis17GY.
 - Inbound media whose download fails now keeps the `media` envelope with `omitted: true` and the declared
   size, on both engines, instead of dropping the field and looking like a message that never had media.
 - Webhook filters and automation rules gated on `hasMedia` now match those messages.
--- a/docs/06-api-specification.md
+++ b/docs/06-api-specification.md
@@ -2278,7 +2278,7 @@ List all contacts for a session, returned as an in-memory paginated window.
 ]
 ```
 
-**Errors:** `400` session is not started · `401` missing/invalid API key, or key not scoped to this session · `409` conflict or engine not ready (retryable)
+**Errors:** `400` session is not started · `401` missing/invalid API key, or key not scoped to this session · `409` conflict or engine not ready (retryable) · `503` the WhatsApp page died while reading contacts (retry shortly)
 
 #### GET /api/sessions/:sessionId/contacts/blocked
 
--- a/openapi.json
+++ b/openapi.json
@@ -4398,6 +4398,9 @@
           },
           "409": {
             "description": "The session is not connected — an engine exists for it but is not `ready`: disconnected, reconnecting, or still initializing, so the request never reached WhatsApp. Wait for `ready` and retry. A session that was never started answers `400` instead, and the session lifecycle routes answer `409` for a conflicting state rather than this. One window answers this while the session still reads `ready`: WhatsApp Web periodically reloads its own page and the engine re-injects into it — for those few seconds the answer is a `409` naming the reload; retry shortly."
+          },
+          "503": {
+            "description": "The WhatsApp page died while reading contacts, retry shortly"
           }
         },
         "summary": "Get all contacts for a session",
--- a/src/engine/adapters/whatsapp-web-js.adapter.spec.ts
+++ b/src/engine/adapters/whatsapp-web-js.adapter.spec.ts
@@ -5624,15 +5624,92 @@ describe('WhatsAppWebJsAdapter page transport error detection (wedged page fast-
     },
   );
 
-  it('detects a transport error from a getter too (getContacts)', async () => {
+  it('converts a transport error from a getter into a 503 (getContacts)', async () => {
     const getContacts = jest.fn().mockRejectedValue(new Error('Protocol error: Target closed'));
     const { adapter, onDisconnected } = readyAdapter({ getContacts });
 
-    await expect(adapter.getContacts()).rejects.toThrow('Protocol error: Target closed');
-
-    expect(onDisconnected).toHaveBeenCalledTimes(1);
+    await expect(adapter.getContacts()).rejects.toBeInstanceOf(EngineTransportError);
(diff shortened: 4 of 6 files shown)


rmyndharis:
Thanks for picking this up, and for the write-up. I cannot take it as it stands, and the reason is upstream of the code, so let me lay it out.

## The mechanism cannot produce the reported symptom

`Client.getContacts()` is a single `pupPage.evaluate`. The contact models are built inside the page, where `getContactModel` calls `createWidFromWidLike(contact.id)` and dereferences the result unguarded, across the whole collection in one `Promise.all`. A contact with a null `id` therefore throws in the page and rejects the evaluate. Nothing crosses back to Node.

The filter added here runs after that `await`, in `WwebjsContacts.getContacts`. On the input it is written for, it never executes: the request has already failed one frame earlier. Merging it would close #1476 while leaving it broken.

Worth saying that #1476 does not establish this cause either. It has no log, no stack trace and no reproduction, and "from version 10 onwards" is hard to square with a mapping line that has not changed in a long time. I have asked the reporter for the container log and for one thing to try, so we should know more shortly.

## Where the filter does bite, it makes things worse

`readWid` is already imported in this file, at line 5. It exists because a WhatsApp Web build renamed the serialized-id property from `_serialized` to `$1`, and its docstring says every raw-Wid read goes through it so the rename lands on all of them at once. A guard that reads `c.id?._serialized` directly is exactly the pattern it was written to replace: under that rename the expression is `undefined` for every contact, so the filter drops the whole address book and the endpoint answers `200 []`. A silently empty contact list is a worse failure than the 500, because nothing surfaces it.

Two smaller points in the same area. The patch cites `wwebjs-chats.ts` as precedent, but that guard counts what it skips and logs `Skipped N chat(s) without a serialized id` (:41, :65-67); this one drops silently, so an operator losing contacts has nothing to go on. And `getContactById` at :45 has the identical unguarded dereference, where a mapping throw falls into the broad catch and returns `null`, which the service turns into a 404 "Contact not found" for a contact that exists.

## What I would take

There is a real change here, and it is smaller than the one proposed. `getContacts` (:26-32) and `getContactById` (:44-51) map an identical six-field shape twice. Fold them into one private `toContact(c)` that reads the id through `readWid`, drop unreadable entries with a count and a `logger.warn` the way the chats path does, and both call sites are fixed at once. One note so it does not surprise you: `readWid` returns `string | undefined` while `Contact.id` is a required `string`, so this needs a filter or a `flatMap`, not a bare `id: readWid(c.id)`, which will not typecheck.

Separately, and worth doing whatever #1476 turns out to be: the `getContacts` catch rethrows a dead page raw, so a transport death answers 500 there while the sibling read in `wwebjs-chats.ts` converts it to `EngineTransportError` and the documented 503. That is a few lines and a genuine defect.

Either of those would be welcome as a PR. If you would rather wait for the reporter's log before spending the time, that is entirely reasonable too, and I will update #1476 as soon as I have it.

Deyvis17GY:
You're right on all counts — thanks for the detailed review. The original guard
sat after the `await`, so on the reported input it never ran (the throw is inside
the page evaluate), and reading `._serialized` directly would have silently emptied
the list under the `$1` rename. Both worse than the 500.

Reworked along the lines you laid out: a single `toContact(c)` helper reads the id
through `readWid` and returns null for an unreadable one, shared by both reads, so
the rename lands on both and getContactById no longer 404s a contact that only
differs by the rename. Skipped entries are counted and logged like the chats path.
Separately, getContacts now converts a dead-page error to `EngineTransportError`
(503) instead of a raw 500.

Dropped "Fixes" to "Refs #1476" since the page-level cause isn't established — happy
to hold for the reporter's log, split the transport change, or add a unit test if
you prefer.

rmyndharis:
This is what I asked for, and both halves landed: one `toContact` reading the id through `readWid` and shared by both call sites, skipped entries counted and logged the way the chats path does, and the dead-page conversion so a transport death answers the documented 503. Thank you for taking the diagnosis on board, and for dropping Fixes to Refs on your own. That was the right call, for reasons I come back to at the end.

I have approved the workflow run, so this branch gets CI for the first time. Two jobs go red, both small.

## Test

`whatsapp-web-js.adapter.spec.ts` still expects the raw Puppeteer message from the getter case, so the conversion you added trips it. It is the only failure in the suite: 1 failed, 1501 passed in the adapter lane. The case now checks conversion rather than propagation, so it wants a new name and the sibling's assertion shape:

```ts
it('converts a transport error from a getter into a 503 (getContacts)', async () => {
  const getContacts = jest.fn().mockRejectedValue(new Error('Protocol error: Target closed'));
  const { adapter, onDisconnected } = readyAdapter({ getContacts });

  await expect(adapter.getContacts()).rejects.toBeInstanceOf(EngineTransportError);

  expect(onDisconnected).toHaveBeenCalledWith('Page transport error during getContacts');
});
```

`EngineTransportError` is already imported in that file for the joinGroupViaInviteCode case, and the onDisconnected assertion still holds unchanged, because reportIfPageTransportError still runs before the throw.

## Lint

`readWid(c.id as unknown as SerializedWid)` at `wwebjs-contacts.ts:31` is an error under `@typescript-eslint/no-unnecessary-type-assertion`, so `npm run lint` exits non-zero. I can see where it came from: `getBlockedContacts` at :162 needs that cast, because the blocklist element type genuinely does not line up. In `toContact` it does. `readWid(c.id)` compiles and lints clean. Keep the `SerializedWid` import, :162 still uses it.

## The 503 needs declaring, and the decorator alone is not enough

The route advertises 200/400/409 today, so the status you added is one the published contract says it cannot produce. Add it to `findAll` next to the 409:

```ts
@ApiResponse({ status: 503, description: 'The WhatsApp page died while reading contacts, retry shortly' })
```

then run `npm run openapi:export` and commit the snapshot, or `openapi:check` fails.

The trap worth naming: `test:docs` cross-checks docs/06 against the contract, and with only the decorator added it fails with `GET /api/sessions/:sessionId/contacts misses 503`. So `docs/06-api-specification.md:2228` needs the code appended to its Errors line too, in the same style the blocklist route uses at :2245.

## Tests for the new behaviour

Nothing exercises `toContact`, the skip counter, the `$1` shape at either call site, or `getContactById` returning null. The file drops from 79.7% to 64.9% line coverage. It stays above the adapters floor, so no gate stops it, which is exactly why it is worth asking for rather than leaving to CI. Five short cases in the same spec cover all of it, using the `readyAdapter` helper already there:

- the happy path with one `_serialized` contact and one `$1` contact, proving the rename lands at the list call site
- one unreadable entry skipped while the rest survive, which is the shape #1476 actually reported
- a non-transport rejection still propagating untouched, with the session left READY
- `getContactById` mapping a `$1` id
- `getContactById` answering null for an unreadable one

## On #1476

You were right to drop Fixes. The reporter has since posted their log, and it points somewhere else entirely: the liveness watchdog declares the session dead thirteen seconds before the contacts call fails, and the read only rejects because our own reconnect closed the browser under it. The contacts read appears to be starving the watchdog's probe rather than dying with the page. That is our defect, not yours, and it is now #1501.

None of that changes what this PR needs. The readWid dedupe is good hygiene that protects the contacts path from the rename, and the 503 is the honest answer for a dead page whatever kills it. Worth taking on its own merits.

Deyvis17GY:
All four points are in (fdd90784): the getter case now asserts the EngineTransportError conversion, the unnecessary cast in toContact is gone (the one in getBlockedContacts stays), the 503 is declared on findAll with the OpenAPI snapshot and docs/06 updated, and five new cases cover toContact, the skip counter, both id shapes, and getContactById's null path. Adapter spec 573/573, lint and openapi:check clean, test:docs green locally.

Thanks for the thorough review — and for running down what #1476 actually was.

PR #1484, opened 2026-08-30
Feat/template chat composer

AbdoHussein1998:
## Description

Added support for sending message templates to specific chats.
Added template message support in the Message Tester for testing and verification.
Added template message display and sending directly within the chat.
## Type of Change

- [ ] Bug fix
- [ ] New feature

image
image

### DIFF
--- a/dashboard/src/components/chats/ChatComposer.tsx
+++ b/dashboard/src/components/chats/ChatComposer.tsx
@@ -1,9 +1,14 @@
 import { useEffect, useRef, useState, type Dispatch, type SetStateAction } from 'react';
 import { useQueryClient } from '@tanstack/react-query';
 import { useTranslation } from 'react-i18next';
-import { Loader2, Paperclip, Send, Smile, X } from 'lucide-react';
-import { messageApi, type Chat, type MessageType } from '../../services/api';
-import { mergeOrAppend, type ChatMessageView } from '../../utils/chatMessages';
+import { Loader2, Paperclip, Send, Smile, X,FileText  } from 'lucide-react';
+import {
+  messageApi,
+  templateApi,
+  type Chat,
+  type MessageType,
+  type MessageTemplate,
+} from '../../services/api';import { mergeOrAppend, type ChatMessageView } from '../../utils/chatMessages';
 import { promoteChatWithSnippet } from '../../utils/chatList';
 import { buildMediaSendPayload, buildOptimisticMetadata, quotedIdOf } from '../../utils/composerSend';
 import { messagesQueryKey, useChatMessagesActions } from '../../hooks/useChatMessages';
@@ -75,6 +80,65 @@ function ChatComposer({
   const [sending, setSending] = useState<boolean>(false);
 
   const [showEmojiPicker, setShowEmojiPicker] = useState<boolean>(false);
+
+  const [showTemplatePicker, setShowTemplatePicker] = useState<boolean>(false);
+  const [templates, setTemplates] = useState<MessageTemplate[]>([]);
+  const [loadingTemplates, setLoadingTemplates] = useState<boolean>(false);
+
+
+
+    /**
+   * Loads the message templates belonging to the currently selected WhatsApp session.
+   *
+   * The composer receives `selectedSessionId` from Chats.tsx, so templates are always
+   * scoped to the same session as the open conversation.
+   *
+   * A small cancellation guard prevents an old request from updating state after the
+   * user switches sessions or the composer is unmounted.
+   */
+  useEffect(() => {
+    let cancelled = false;
+
+    const loadTemplates = async () => {
+      if (!selectedSessionId) {
+        setTemplates([]);
+        setLoadingTemplates(false);
+        return;
+      }
+
+      setLoadingTemplates(true);
+
+      try {
+        const result = await templateApi.list(selectedSessionId);
+
+        if (!cancelled) {
+          setTemplates(result);
+        }
+      } catch (err) {
+        if (!cancelled) {
+          setTemplates([]);
+          showErrorToast(
+            t('chats.errors.loadTemplates'),
+            err instanceof Error ? err.message : undefined,
+          );
+        }
+      } finally {
+        if (!cancelled) {
+          setLoadingTemplates(false);
+        }
+      }
+    };
+
+    void loadTemplates();
+
+    return () => {
+      cancelled = true;
+    };
+  }, [selectedSessionId, t, showErrorToast]);
+
+
+
+
   // Monotonic token invalidating an in-flight attachment FileReader: picking a second file (or
   // removing the attachment) before `onload` fires must win over the late-arriving bytes —
   // otherwise the slower read overwrites the newer pick. Same pattern as composeImageReadSeq.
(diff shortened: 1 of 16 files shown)


rmyndharis (changes requested):
Thanks for this, and welcome. Reaching a template from the composer is a real gap and people have asked for it. The shape you chose for the composer is also the right one: filling the input rather than sending immediately means the user sees the text, and can edit it, before anything reaches a customer. I want to take this. It needs a round first, and I have tried to make every item below something you can act on directly rather than a matter of taste.

CI has never run on this branch, so none of this reached you automatically. Two of the items are red jobs.

## Enter no longer sends a message

This is the one to fix first, because it changes existing behaviour for every user and the diff does not read as if it were intended.

The message box changed from `<input type="text">` to `<textarea>` (ChatComposer.tsx:477). The form has always relied on implicit submission: a single-line text input inside a form submits it when you press Enter, and that is what sent the message, since there is no key handler anywhere in the file on either version. A textarea does not do that. Measured in Chromium:

```
<input type="text"> in a form: Enter submitted = true   value = "hello"
<textarea> in a form:          Enter submitted = false  value = "hello\n"
```

So on this branch, pressing Enter in the chat composer inserts a newline instead of sending. Either keep the `<input type="text">` (a template with newlines still pastes in fine, it just is not multi-line to edit), or keep the textarea and restore the behaviour on purpose with an `onKeyDown` that sends on Enter and lets Shift+Enter insert a newline. Either is fine by me; the second is nicer if you want multi-line drafting, and it is worth a line in the PR description because it is a behaviour change people will notice.

## Four strings show up as raw key names, in every language

The picker calls `t('chats.loadingTemplates')`, `t('chats.noTemplates')`, `t('chats.errors.loadTemplates')` and `t('Select Tempate')`, but no `chats.*` key was added to any locale file. When i18next cannot find a key it renders the key itself. Running it against this branch's own en.json:

```
t("chats.loadingTemplates")     -> "chats.loadingTemplates"
t("chats.noTemplates")          -> "chats.noTemplates"
t("chats.errors.loadTemplates") -> "chats.errors.loadTemplates"
t("Select Tempate")             -> "Select Tempate"
```

So the empty state literally reads `chats.noTemplates` on screen, and the dropdown's first option reads `Select Tempate`, misspelling included, in all thirteen languages. The last one is also passing an English sentence as a key; everything else in the dashboard uses a namespaced key and looks the text up.

Worth knowing why nothing caught it: `i18n:check` compares each locale against `en.json`, so a key missing from all of them at once is invisible to it. The gate is a parity check, not a completeness check.

The Message Tester block has the same issue in a milder form: its label and its three select states are hardcoded English rather than `t()` calls, while every other string on that page is translated.

## The dropdown has no accessible name, and it fails the suite

`npm test` in `dashboard/` has exactly one failure:

```
not ok 6 - every native control is named, or is hidden from assistive tech
  actual: [ 'components/chats/ChatComposer.tsx:391 <select>' ]
```

`src/a11y-controls.test.ts` scans for native controls with no accessible name. One `aria-label` on that `<select>` turns it green. The new Templates button has the same gap in a different way: its `title` and `aria-label` are the literal string "Templates" while the attach and emoji buttons beside it use `t()`.

## Formatting fails the lint job

`npm --prefix dashboard run format:check` rejects ChatComposer.tsx, MessageTester.tsx and services/api.ts. `npm --prefix dashboard run format` fixes all of it in one pass. Worth looking at the result afterwards, because one of the things it will fix is ChatComposer.tsx:11, where the end of the `services/api` import and the start of the `chatMessages` import ended up on the same line. It compiles, but it hides an import from anyone reading the header.

## he.json was skipped

`messageTester.types.template` went into twelve locales and not into `he.json`, which fails `i18n:check`. While you are in those files: five of the twelve got the English word rather than a translation, and a few use a plural where every neighbouring type label is singular.

## Sending a template with placeholders sends the placeholders

A template body can contain `{{name}}` tokens, and the Templates page treats them as a first-class feature. The Message Tester posts `{ chatId, templateId }` with no `vars`, and the server deliberately leaves unmatched placeholders literal so that a missing variable stays visible. So a template reading `Hi {{name}}, your order {{id}} shipped` reaches the recipient's phone with the braces intact.

The composer path avoids this by construction, since the text lands in the input where the user can see and replace the tokens before sending. The tester path has no such step, and `messageApi.sendTemplate`'s type has no `vars` field, so no caller can supply one even though the API has always accepted it. The cheapest honest fix is to widen the type to include `vars?: Record<string, string>` and render an input per placeholder, reusing `extractPlaceholders` and `renderPreview`, which already exist in Templates.tsx. If you would rather keep this PR small, showing a preview of the rendered text under the dropdown is enough to make the problem visible to the user, and I would take that.

## Smaller things

- **A stale selection survives a session switch.** In MessageTester, `setSelectedTemplate('')` runs only when the session is cleared (:120) or the fetch throws (:138), never on a normal switch from one session to another. The previous session's id stays selected with Send enabled, and that request always 404s. Resetting at the top of the effect covers it.
- **`useTemplatesQuery` already exists** in `dashboard/src/hooks/queries.ts:134`, and MessageTester already imports from that file. Both new fetches are about forty lines of `useState` plus `useEffect` plus a cancellation flag that the hook does for you, with caching shared with the Templates page. Replacing them is a net deletion.
- **The picker has no styling.** `chats-template-picker`, `template-picker-loading`, `template-picker-empty` and `template-select` appear only in the JSX; no stylesheet defines them, so the panel renders as a bare browser dropdown. Note `src/styles.scope.test.ts` requires rules in a page stylesheet to be scoped under that page's root class.
- **Viewer keys get an error toast.** `GET /sessions/:id/templates` is `@RequireRole(OPERATOR)` (template.controller.ts:24-25), but the composer fetches on every session change regardless of role, so a viewer sees a failure toast on the Chats page for a picker they cannot use.
- **Selecting a template replaces the draft** rather than appending, with no undo. The emoji handler in the same file appends; matching it is one line.
- **Opening the emoji panel does not close the template picker**, so both can be open at once. Your Templates button already closes the emoji one; the reverse was not added.
- **`package-lock.json` carries one unrelated line**, `"hasInstallScript": true` on better-sqlite3, from an incidental `npm install`. `git checkout origin/main -- package-lock.json` drops it.
- **No CHANGELOG entry.** Two lines under `### Added` in `[Unreleased]` covers it. Entries here close with a `Thanks @user.` sentence, so please add your own.

One process note: this repo merges pull requests rather than squashing, so every commit on the branch lands on main as-is. Squashing the five into one or two commits with conventional subjects would keep "testing templete" and the branch-into-itself merge out of the history.

None of this is a redesign, and the two paths you built are both sound ideas. Ping me when you have pushed and I will approve the workflow run so CI reports on it properly.

rmyndharis:
Checking in. The review above still stands, and the branch now conflicts with `main`, so a rebase is the first step either way. If there is no update within two weeks I will close this to keep the queue tidy; the composer gap is real and a fresh PR is welcome at any time.

AbdoHussein1998:
I’m really grateful to you 😁! I’m currently developing more features and adding additional levels of authorization.

I’ve also added more connection statuses, and I’ve integrated a real Brave browser instance with session support instead of using the testing Chrome version. There are many more improvements and features coming as well.

Once I’ve finished everything, I’ll share all the features together in one bulk update.

I’m truly grateful for the opportunity to contribute to such a great project, and I’d be very happy to be part of its development!

rmyndharis:
Closing this one as contained in #1623, which carries these commits plus the later work, and is closed there with the detail.

The composer gap this opened is still real: `main` has no way to reach a saved template from the Chats page. The version in #1623 is the better answer, because it sends the stored template through the send route instead of concatenating header, body and footer into the text box, and it handles placeholders instead of leaving them literal. That is item 1 of the list on #1623, and it lands on its own against current `main` in about eighty lines plus the locale entries.

PR #1485, opened 2026-08-31
feat(automation): gate autoreply rules on the conversation, not just …

alexandremartinxx:
…the message

Two optional rule fields answer questions about the CHAT rather than about the inbound message, which is why they are rule columns and not `conditions` entries: the webhook filter registry resolves every field synchronously out of the message payload, and neither question can be answered without reading the messages table.

- `newContactOnly` replies only in a chat with no earlier stored message, in either direction — a first-contact greeting. The projector commits the inbound row before dispatching, so the probe excludes the current message by its engine id, and counts id-less rows explicitly (SQL `<>` never matches NULL). Reads under both user-id dialects, so a contact known as `@s.whatsapp.net` is not greeted again as `@c.us`.
- `pauseOnHumanReply` silences the rule in a chat for good once a human sends into it, from the API, the dashboard, or a linked phone. Permanent by design: a conversation an operator took over does not become the bot's again on a timer.

`messages.automated` is what makes the second one possible. An autoreply and an operator's send are otherwise byte-identical OUTGOING rows written by different writers (`saveOutgoingMessage` for API sends, `handleOwnSendEcho` for phone- composed ones), so a naive probe would read the rule's own first reply as a human takeover and silence it forever after one message. The marker rides both echo-race merges, or losing that race would drop it.

Both gates fail CLOSED: if the probe cannot run, the rule does not reply. The ambiguity is not symmetric — talking into a chat the operator fenced off is worse than losing one autoreply to a transient DB fault. They also run only for a rule that declares one, after the cooldown pair rather than between its check and its set, so an ungated rule pays no extra round-trip per inbound message and a burst still collapses to one reply.

`dialectVariants` moves to a shared util: the send-pacing cold-contact probe and these gates ask the same question of the same table.

## Description

Brief description of changes

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] Lint passes
- [ ] Self-reviewed

## Related Issues

Closes #


rmyndharis (changes requested):
Thanks for the detailed write-up on this one. The core idea is sound: neither question can be answered from the message payload, so a rule column is the right shape rather than a `conditions` entry. The migration is careful about both dialects, and keeping `automated` out of the request DTO so a REST client cannot mark its own send invisible is the right call.

Requesting changes on scope first, then correctness.

## Scope

**Drop `RETOMAR-FLUXOS-CHATBOT.txt`.** It is 408 of the 1123 added lines, it is personal working material rather than project documentation, and it is unrelated to the gates. Please rewrite the branch so neither the file nor the commit that adds it appears in its history.

**Split the two gates.** `newContactOnly` is self contained: one column, one probe, nothing outside the automation module. `pauseOnHumanReply` is what brings in the `messages.automated` column on a hot table, the `SendOrigin` argument, the widened `PluginMessagePort`, and the marker carries at both echo race sites. I would like to land the first on its own and settle the second's semantics separately.

## Blocking: `newContactOnly` fails open on whatsapp-web.js

Both probes read the local `messages` table, and only Baileys ever backfills prior conversations into it. `onHistoryMessages` is optional on the callback interface (`src/engine/interfaces/whatsapp-engine.interface.ts:740`) and its only producer is the Baileys adapter (`src/engine/adapters/baileys.adapter.ts:168`). There is no equivalent on the whatsapp-web.js path.

So on wwebjs, an account linked with existing conversations starts from an empty `messages` table, every one of those chats reads as a new contact, and the rule greets all of them as their next inbound message arrives. The 300 second freshness gate spreads that out, it does not prevent it. The same hole reopens on either engine after a database restore.

whatsapp-web.js is the default engine, so this needs an answer before the gate can ship. How would you source "does this chat have history" in a way that does not depend on which adapter is running?

## Blocking: two opening messages produce no greeting at all

The probe excludes only the current row, by engine id (`src/modules/automation/automation-rules.service.ts:241-251`). Nothing serializes the inbound path per chat: Baileys pushes every message of an upsert batch through a concurrency limiter without awaiting, and the projector commits the row before its fire and forget dispatch. Two messages arriving together therefore each see the other as prior history, and the chat is never greeted. The gate is monotonic, so that is permanent.

"hi" followed immediately by "are you there?" is the most common opening there is, which makes this the feature's headline case rather than an edge one.

## Blocking: `pauseOnHumanReply` is retroactive and unbounded

The probe asks whether any OUTGOING row with `automated = false` exists in the chat, with no time bound and no relation to when the rule became responsible for it (`automation-rules.service.ts:260-267`). Two consequences:

- Replies sent before the upgrade read `automated = false`, since the column defaults false with no backfill. Rules have been replying through this same send path since 0.14.0, so an operator who enables the flag finds the rule inert in every chat it has already answered.
- Any account that messages first (send-bulk, send-template, outreach) has an outbound row before the customer ever writes in, so the rule is dead in that chat from the start.

Three comments state the opposite and need correcting either way: the migration header at line 15, `automation-rules.service.ts:258-259`, and the `automated` docblock in `message.entity.ts`. All three say nothing wrote automated replies before the column existed.

I think the answer is a time bound rather than more marking: count only human sends after the point the rule took the chat. Worth agreeing on that before you build it.

## Blocking: group behaviour is the inverse of what is documented

The human reply probe filters `direction: OUTGOING`, but every group participant's message persists as INCOMING (`src/modules/session/message-projector.service.ts:264`). A group full of people talking never closes the gate; the only thing that closes it is the account's own prior post there. And `newContactOnly` keys on `chatId` alone, so in a group it fires once for the room and never per participant, even though `author` is already stored for exactly that distinction.

Either scope both gates to 1:1 chats explicitly, or handle groups and say what they mean. The docs do not mention groups at all right now.

## CI is red

`npm test -- --coverage` fails on this branch. All tests pass, but `./src/database/` branch coverage lands at 68.03% against a 69% floor, because the migration ships without a spec. The 18 migrations before it each have one under `src/database/migrations/__tests__/`; `1786300000000-AddWebhookDeliveryFailureLookupIndex.spec.ts` is the pattern. One spec exercising `up()` and `down()` across both dialect branches puts the directory back over the line.

## Smaller items

- `MessageListItemDto` was not updated, so `GET /sessions/:id/messages` now returns an `automated` field that its declared 200 schema does not contain. `mediaPath` and `mediaMimetype` were handled there when they were added.
- The entity comment at `automation-rule.entity.ts:71` and the matching sentence in `docs/06` tell the operator that retention re-arms the gate. There is no message retention in this project; the only retention purge is the chat media archive's, which deletes blobs and not rows. As written it points at a recovery path that does not exist.
- Nothing asserts the persisted `automated` value on a real send. The automation spec writes the row by hand, so deleting both marker carries in `message-send.service.ts` leaves the suite green.

## What I would like to see next

A branch with the unrelated file gone from its history, scoped to `newContactOnly`, with the migration spec, an answer on the engine asymmetry, and group behaviour either excluded or specified. `pauseOnHumanReply` can follow as its own PR once we have agreed on the time bound.

rmyndharis:
One time-sensitive addition to the earlier request to drop `RETOMAR-FLUXOS-CHATBOT.txt`: that file ends with a live `owa_k1_` API key.

Rewriting the branch history removes it from your branch, but a pushed pull-request ref stays on GitHub permanently (`refs/pull/1485/head` is served even after a force-push or once the PR is closed), so the key is already public and cannot be un-published. Please rotate it now, separately from the history rewrite: issue a fresh API key on your OpenWA deployment and revoke this one. I have not quoted the key here.

The rest of the standing review is unchanged.

rmyndharis:
Closing this one. The branch still carries `RETOMAR-FLUXOS-CHATBOT.txt` with a live API key two days after the rotation request, the latest commit reads `feat incompleta`, and the review has had no response. The conversation-gated autoreply idea is worth having: a fresh PR without the stray file, on a rotated key, is welcome at any time. To repeat the urgent part: the key stays reachable through the pull-request ref even after this closes, so rotating it is the step that matters.

PR #1486, opened 2026-08-31
fix(groups): refuse setGroupDescription on whatsapp-web.js instead of failing with a 500

purnamcommunity:
## Description

Split out of #1472 at review request.

`PUT /sessions/{sessionId}/groups/{groupId}/description` fails on the whatsapp-web.js engine and
has been surfacing as a bare `500`. whatsapp-web.js types `GroupChat.setDescription` as
`Promise<boolean>`, but its injected evaluate calls the page's
`WAWebGroupModifyInfoJob.setGroupDescription(chatWid, description, newId, descId)`, which reaches
`widToGroupJid` with an undefined Wid and throws
`TypeError: Cannot read properties of undefined (reading 'toJid')`. The argument list the library
sends no longer lands the Wid where the page reads it.

The adapter now throws `EngineNotSupportedError` (`501`) instead. The description was never applied
either way; this makes the route say so rather than reading like an outage.

### Measurement

Measured live on wwjs 1.34.7, the latest published release:

- with `descId` undefined (group had no description) and with it defined (description added from
  the phone first) — identical throw, so the first-description case is not the variable
- `setGroupSubject` as a control on the same group in the same session succeeds, and it builds its
  Wid through the same `WAWebWidFactory.createWid`, so `createWid` is not at fault
- Baileys sets descriptions normally on the same account

The signature lives in the minified page bundle rather than in the library, so there is nothing in
`whatsapp-web.js` to patch around for this.

### Contract

- `docs/29` §29.4.6 row, and the totals it feeds: **199→198 ✅**, **25→26 ❌**, library-limitations
  **23→24**, methods carrying a ❌ **24→25**, REST-neutral **90→89**, Baileys-only **12→13**.
  `setGroupDescription` also drops out of the `getChatById` exposure list, since the adapter now
  refuses before reaching the client.
- `@ApiResponse({ status: 501 })` on `group.controller.ts:396`, with the matching docs/06 errors
  line, engines note and callout, and a regenerated `openapi.json`.
- CHANGELOG under `### Changed`, since a caller on whatsapp-web.js sees a different status than
  before.

Two figures in the §29.4 totals were already wrong before this change and are corrected in passing:
`89 fully supported` should have read 88, which contradicted §29.8's own `88 + 2`, and
`11 are Baileys-only` should have read 12. Neither is covered by `docs-29-counts.spec.ts` — it
checks the bolded totals, not these parentheticals.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

A caller on whatsapp-web.js gets `501` where it previously got `500`. Both are failures and the
description was applied in neither, so no working call changes behaviour — but an integration that
matched on `500` will see a different status.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

Locally: `test` (5939), `test:docs` (264), `lint`, `format:check`, `openapi:check`,
`check:contract-shapes` (324 pairs), `check:sdk-routes`, `check:sdk-docs`.

N/A — engine/contract-level change, no UI in this repo.

## Related Issues

Split out of #1472.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `PUT /sessions/{sessionId}/groups/{groupId}/description` now answers `501` on the whatsapp-web.js
+  engine instead of failing with a bare `500`. The library types `GroupChat.setDescription` as
+  supported, but the page job it calls (`WAWebGroupModifyInfoJob.setGroupDescription`) no longer
+  accepts the argument list it sends and throws inside the page. The description was never applied
+  either way; the route now says so. Measured on wwjs 1.34.7 with the group's description both
+  unset and set, with `setGroupSubject` as a control on the same group in the same session. Baileys
+  is unaffected and remains the engine for this route. The capability matrix and docs/29 record it
+  as a library limitation.
 - A whatsapp-web.js protocol timeout is no longer eligible to be classified as a dead page.
   Behaviour is unchanged on the current Puppeteer; the guard keeps a future bump from reporting a
   slow command as a transport death.
--- a/docs/06-api-specification.md
+++ b/docs/06-api-specification.md
@@ -2858,7 +2858,12 @@ No `@HttpCode`; PUT default is `200`.
 
 Change the group description. An empty string clears the description.
 
-**Auth:** API key (OPERATOR)
+**Auth:** API key (OPERATOR) · **Engines:** Baileys only — whatsapp-web.js returns `501`
+
+> **whatsapp-web.js returns `501`.** The library types `GroupChat.setDescription` as supported, but
+> the page job it calls (`WAWebGroupModifyInfoJob.setGroupDescription`) no longer accepts the
+> argument list it sends, so the description is never applied. This was previously a bare `500`.
+> See docs/29 §29.4.6 for the measurement.
 
 **Path parameters**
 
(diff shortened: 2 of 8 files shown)


purnamcommunity:
Closing in favour of #1487, which fixes this rather than declaring it unavailable.

The premise here was wrong. This PR argued that because the signature lives in the minified page bundle rather than in the library, there was nothing to patch around. The first half holds; the conclusion does not. The bundle is readable, and the shape it wants is:

```js
function* (t) {
  var desc = t.desc, groupWid = t.groupWid, newDescId = t.newDescId, prevDescId = t.prevDescId;
  ... sendSetDescriptionRPC({ iqTo: widToGroupJid(groupWid), descriptionId: newDescId,
                              descriptionPrev: prevDescId,
                              bodyArgs: desc != null ? { bodyElementValue: desc } : undefined,
                              hasDescriptionDeleteTrue: desc == null }) ...
}
```

A single options object. wwjs passes four positional arguments, so `t` is the Wid, every field reads back undefined, and `widToGroupJid(undefined)` throws — which is exactly the `toJid` error measured here. `setGroupSubject` in the same module is still positional (`function* (wid, subject)`), which is why subjects kept working and made the Wid look innocent.

#1487 sends the options object as an install-time patch, in the same place as the other `patch-wwebjs-*` fixes. Measured on wwjs 1.34.7 against a test group: setting a description answers 200 and reads back, an empty string clears it, no page throw. So the capability is restored rather than demoted, and the matrix cell stays `supported` carrying 🔧⁹ instead of flipping to `not-available`.

Everything this PR carried that is still true — the failure analysis, the `descId`-undefined-and-defined measurement, `setGroupSubject` as the control — is in #1487.

PR #1487, opened 2026-08-31
fix(groups): realign the wwjs group-description job call with the page bundle

purnamcommunity:
## Description

Supersedes #1486, which demoted this capability to `501` on the strength of "the signature lives in the minified page bundle, so there is nothing here to patch around". The first half is right; the conclusion was not. The signature is recoverable, and the fix belongs in the same place every other `patch-wwebjs-*` fix lives.

`GroupChat.setDescription()` calls the page's `WAWebGroupModifyInfoJob.setGroupDescription(chatWid, description, newId, descId)` positionally. That job now takes a single options object, deminified from the page bundle:

```js
function* (t) {
  var desc = t.desc, groupWid = t.groupWid, newDescId = t.newDescId, prevDescId = t.prevDescId;
  ... sendSetDescriptionRPC({ iqTo: widToGroupJid(groupWid), descriptionId: newDescId,
                              descriptionPrev: prevDescId,
                              bodyArgs: desc != null ? { bodyElementValue: desc } : undefined,
                              hasDescriptionDeleteTrue: desc == null }) ...
}
```

So `t` is the Wid, every field reads back undefined, and `widToGroupJid(undefined)` throws `Cannot read properties of undefined (reading 'toJid')` — the bare 500. `setGroupSubject` in the same module is still positional (`function* (wid, subject)`), which is why subjects kept working and made the Wid look innocent: the same `createWid` feeds both, and only the description job's parameter shape drifted.

## The patch

`scripts/patch-wwebjs-group-description.js` sends the options object. Three details come from WhatsApp Web's own call site rather than guesswork:

- `newDescId` is `WARandomHex.randomHex(8)`, not `WAWebMsgKey.newId()`.
- An empty description is sent as `desc: null`, which selects the job's delete branch. Passing `''` would send an empty body element instead — and `''` is a documented input here (docs/06: "An empty string clears the description"), so this one is a silent behaviour change rather than a visible failure if it is got wrong.
- `prevDescId` keeps the existing `descId` lookup.

The `ServerStatusCodeError` rejection is untouched, so the admin-rights refusal that `EngineRefusedError` depends on still arrives the same way.

## Verification

Measured on wwjs 1.34.7 against a test group:

| Call | Result |
| --- | --- |
| `PUT .../description` with text | `200`, and the new text reads back from the group |
| `PUT .../description` with `""` | `200`, and the description reads back empty |
| restore original | `200`, group returned to its prior description |

No `toJid` throw in the page for any of them. The gateway ran the patched tree (`WhatsAppWebJsAdapter`, patch applied before start).

Not exercised: the admin-refusal `403` path, which needs a group where the account is not an admin. The transform does not touch it, but that is an argument, not a measurement.

## Contract

- `docs/29`: the 🔧⁹ row in §29.3.1, the dependency entry in §29.3.2, the row-level mark on `setGroupDescription`, and every patch count those feed (9 patches, 7 on wwjs; the wwjs patch-dependency tally 9 → 10).
- Wired into `postinstall.js`, the Dockerfile copy and fatal-run lists, and `test:scripts` — `dockerfile-patchers.spec.js` enforces all three.
- CHANGELOG under `### Fixed`.

`docs-29-counts.spec.ts` matched `✅🔧` marks with a `[¹²³⁴⁵⁶⁷]` class, so a new mark was invisible to it and marking this row read as drift rather than agreement. The class now spans every superscript a patcher can carry.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] Tests added/updated
- [x] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

Locally: `test` (5941), `test:docs` (264), `test:scripts` (123), `lint`, `format:check`, `openapi:check`, `check:contract-shapes` (324 pairs).

N/A — engine/patch-level change, no UI in this repo.

## Related Issues

Supersedes #1486. Split originally out of #1472.


rmyndharis:
Thanks for splitting this out, and for chasing the actual page signature instead of stopping at the 500. The transform is right: it matches the pinned 1.34.7 byte for byte, and mapping an empty description to `null` is better than the open upstream fix, which would regress the empty-string clear that `docs/06-api-specification.md` documents.

Nothing needed sending back, so I pushed four small things onto your branch instead. None of them touch the transform.

- `docs/29-engine-capability-matrix.md`: the 29.3.1 heading still counted eight patches over a nine row table.
- The 29.3.1 row and the patcher header now name the open upstream fix, wwebjs/whatsapp-web.js#201895, the way the #201832 and #201816 entries already do. It carries more weight than a citation usually would. Upstream keeps `await WAWebMsgKey.newId()` and passes the description through verbatim, so on the release that ships it this patcher matches neither its find nor its replace: it stops the image build rather than standing down. Whoever retires it then has to re-cut the transform to the `'' -> null` delta on top of the upstream shape, because the documented clear lives only in that mapping. I also folded your live measurement into the header so it survives outside the PR body.
- `CHANGELOG.md`: contributor credit, per `docs/20-community-guidelines.md`.
- `scripts/postinstall.spec.js`: the "all patchers run in stable order" test has a hand-listed fixture, so it never wrote your patcher into the temp root, never planned it, and `assert.equal(steps.length, 8)` stayed green while the test's own comment claimed to cover every patcher. `patch-wwebjs-block.js` had already slipped through the same way. The fixture is now derived from the `patch-*.js` files that ship and cross-checked against an explicit order list, so patcher ten cannot repeat it.

One question, and it is not a blocker: what led to `WARandomHex.randomHex(8)` in place of `await WAWebMsgKey.newId()`? The id is independent of the argument-shape bug, and it is the second reason this patcher can never stand down on an upstream release. If `newId()` was tried under the options shape and rejected, say so and I will record that in the header as deliberate. If it was not tried, I would still keep `randomHex(8)`, since that is the variant you measured.

Local gates on the pushed head: `test` 5941, `test:docs` 264, `test:scripts` 122, `lint`, `format:check`. CI is approved and running now, for the first time on this PR.

purnamcommunity:
Thanks, and thanks for pushing those four rather than sending it back. The postinstall fixture one is the good catch: I added a patcher and never noticed the order test wasn't planning it.

On the id: it wasn't tried. I took `randomHex(8)` straight off WhatsApp Web's own call site while I was reading the deminified job, and never went back to check whether `newId()` also works under the options shape. So record it as copied, not chosen. Happy for the header to say that plainly.

One thing you can add to it though, since it's a measurement and not an argument. The third row of my table, restoring the original description, exercised the id's round trip: that second edit reads `prevDescId` from the group's own `descId`, which is the id the first call minted. So the id `randomHex(8)` produces is stored by WhatsApp, read back, and accepted as `prevDescId` on a later edit. That's evidence for the variant that ships, independent of whether `newId()` would work too.

Issue #894, opened 2026-07-24
[Feature]: API key should be restricted to particular session


hayskytech:
### Pre-flight

- [x] I searched existing issues and this isn't already requested

### Problem / motivation

Suppose I have two sessions and I want to give access to only one session, then currently we don't have that feature. Operator key can access all sessions.

### Proposed solution

While creating API keys we should be able to select session. That API key is restricted to selected session. For this we should block all sessions endpoint for that API key which has session level scope.

### Scope

- [x] I understand some features are limited by the underlying WhatsApp engine (e.g. interactive Buttons/List messages are not supported on whatsapp-web.js).

rmyndharis:
Hi @hayskytech, thanks for the suggestion — good news: session-scoped API keys are already supported by the REST API. We just verified the behavior end-to-end on the current release (0.10.8).

When creating an API key via `POST /api/auth/api-keys`, pass an `allowedSessions` array with the session UUIDs the key may access:

```bash
curl -X POST http://<host>:<port>/api/auth/api-keys \
  -H "X-API-Key: <admin-key>" \
  -H "Content-Type: application/json" \
  -d '{
    "name": "session-1-only",
    "role": "operator",
    "allowedSessions": ["<session-uuid>"]
  }'
```

Verified behavior of such a scoped key:

- `GET /api/sessions/:id` for an in-scope session → `200`; for any other session → `401 API key not authorized for this session`.
- Per-session resources (messages, contacts, groups, webhooks, …) follow the same rule.
- `GET /api/sessions` returns only the sessions within the key's scope — list endpoints are **filtered, not blocked**, so scoped clients can still discover the sessions they own. The cross-session webhook, audit-log, and search endpoints behave the same way.
- The scope can be changed later via `PUT /api/auth/api-keys/:id`; the change takes effect immediately.

The part that is genuinely missing is **dashboard support** — the "Create API Key" form in the web dashboard only exposes `name` and `role` today, so the scope can only be set through the REST API. We're treating this issue as the request for that UI: a session multi-select in the create/edit form plus a scope indicator in the keys table.

While verifying this we also spotted one inconsistency: a scoped key can currently still call `POST /api/sessions` and create a new session — which it then can't even access, since the new id isn't in its allowlist. We'll tighten that as part of this issue (either rejecting session creation for scoped keys, or auto-adding the newly created session to the key's scope).

[...]

PR #1492, opened 2026-08-31
Scope operator and viewer API keys to chosen sessions

sebathi:
## Summary
- Operator and viewer (reader) API keys can be limited to specific WhatsApp sessions from the dashboard.
- If no sessions are selected, the key keeps access to every session, including ones created later.
- The session list stays hidden behind a **Choose sessions** link. **Leave for all sessions** clears the allowlist and hides the list again.
- Translations are included for every dashboard locale.

The backend already enforced `allowedSessions`; this exposes that choice in the API Keys UI and adds update support for existing keys.

## Test plan
- [ ] Log in as admin and open **API Keys**
- [ ] Create an operator key: confirm the session list is hidden and only **Choose sessions** is shown
- [ ] Open the list, select one session, create the key: the table shows that session
- [ ] Click **Leave for all sessions**, save (or create with none selected): the table shows **All sessions**
- [ ] Create a viewer key with no sessions selected: it can list every session
- [ ] A key scoped to session A gets 401 on session B
- [ ] Edit a scoped key with the pencil: the list starts open; leaving for all sessions and saving restores full access
- [ ] `cd dashboard && npm run i18n:check && npm test`

Refs #894


sebathi:
The logic for session-focused api keys was already inplace.

rmyndharis (approved):
Thanks, this is a clean contribution and it lands the half of #894 that was still missing. Two details that matter and that you got right: the picker binds `session.id` rather than `session.name`, which is the one mistake that would have silently scoped keys to nothing, and the create form clears the selection when the role changes to admin. Your rewrite of the mobile table rules to `:has()` also fixes a latent bug beyond the new column: the old `nth-child` selectors pointed at the wrong cells between 480 and 640px, once the key column is hidden.

Rather than send it back for small things, I pushed two commits to your branch.

**9df65bd8** corrects two behaviours in the scope editor.

An unchanged Save was still sent. `PUT /auth/api-keys/:id` writes `allowedSessions` whenever the field is present, so opening the pencil on a key that was never scoped and pressing Save stored `[]` over `null`. `AuthService.update` compares the authorization fields before and after, treats that as a change, and disconnects every live `/events` socket holding the key. The client does not reconnect after a server-initiated close, so those tabs stay dead until someone retries by hand. It also wrote an audit row saying the scope had moved. The editor now compares membership and closes without a request when nothing changed.

Nothing prunes a key's allowlist when a session is deleted, so an orphaned id is a steady state rather than an edge case. The picker listed live sessions only, which left such an id invisible, impossible to untick, and re-persisted on every save. A selected id with no live session now renders as the raw id so it can be removed, and the empty state keys off the row count so it no longer swallows those rows.

The scope group also picked up the accessible name that `apiKeys.sessions.label` was written for. That string is translated in all 13 locales and asserted by your new test, but nothing rendered it.

Both helpers went into `utils/sessionScope.ts` with tests, since the dashboard test runner globs `src/**/*.test.ts` and does not pick up `.tsx`.

**8d6934c2** corrects the README and adds the changelog entry.

A scoped key is refused in three shapes and the README named one. A request naming another session is `401`, but the key-management routes and the queue dashboard name no session at all and answer `403`, while session-filtered lists return that key's rows rather than an error.

Two things I left alone, since they are language calls rather than mechanical ones:

- The zh-HK strings use 密鑰 for the API key, where the rest of that catalogue uses that word for a secret. Both Chinese catalogues also introduce 全部 for "all" where they otherwise use 所有 throughout.
- `restricted` ships no plural forms, including in ar and he, where the catalogue carries CLDR categories elsewhere.

Neither blocks this. Could you add a `Refs #894` line to the description? That issue is where the scope of this UI was written down, and it should carry a pointer to the PR that delivered it.

PR #1498, opened 2026-09-01
feat(dashboard): add CSV/TXT file upload for bulk recipients

harry0x:
## Description
Added a new feature to the Message Tester Dashboard that allows users to upload a `.csv` or `.txt` file for bulk message recipients. 

This enhancement reads the uploaded file, extracts phone numbers (handling comma or newline separation), removes duplicate entries, and automatically populates the recipients text area. This improves the user experience by eliminating the need for manual copy-pasting of large recipient lists.

Also added the necessary UI translations for the `Upload CSV/TXT` button across all supported locale JSON files.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

![Upload CSV/TXT Feature Preview]
Screenshot 2026-09-01 at 11 24 40 AM

*(Note: I have attached the screenshot showing the new "Upload CSV/TXT" button placed cleanly on the right side of the Recipients label).*

## Related Issues
Closes #


rmyndharis (changes requested):
Thanks for this, and welcome. The idea is sound, the async wiring is more careful than most first patches in this area, and updating all 13 locales without being asked is the right instinct. There is one real problem, and fixing it the short way also removes several of the smaller notes below.

First, on CI, so the list does not read as "your build is red": no check has ever run on `700747ea`. Fork PRs sit at `action_required` until a maintainer approves the workflow, so GitHub has validated nothing here. I ran the gates locally against your head and they all pass, including all six steps of the `dashboard` job (lint, format:check, typecheck, i18n:check, build, 332/332 unit tests) and the parity gate at 781 keys across 13 locales. Your "Lint passes" tick is accurate. What that green does not cover is worth knowing: nothing in the repo executes a line of the new handler, there is no jsx-a11y plugin in `dashboard/eslint.config.js`, and the parity gate only warns about untranslated values of 20 characters or more (`dashboard/scripts/check-i18n-parity.mjs:24`), so a 14-character label can be wrong and still pass. None of the findings below would have been caught by automation.

## The main change: reuse `parseBulkRecipients` instead of parsing inline

`MessageTester.tsx:15` already imports `parseBulkRecipients`, and `:269` already uses it to build the actual request. The block at `:202-216` re-implements trimming, blank-dropping and dedup right next to it, and the one thing it adds on top, the comma in `text.split(/[\r\n,]+/)` at `:203`, is where the bug lives.

That split has no notion of a row or a column, so it flattens the file into a bag of cells, and `parseBulkRecipients` then normalises any cell containing digits into `<digits>@c.us` (`dashboard/src/utils/bulkRecipients.ts:19-20`). Nothing downstream stops it: `chatId` is a bare `@IsString()` at `src/modules/message/dto/bulk-message.dto.ts:109-112` and goes straight to `engine.sendTextMessage(chatId, ...)` at `src/modules/message/bulk-message.service.ts:772-773`. Running the real path:

```
id,name,phone,joined
1,Ada,628111222333,2024-01-15
2,Alan,447700900111,2024-03-02
```

sends `1@c.us`, `628111222333@c.us`, `20240115@c.us`, `2@c.us`, `447700900111@c.us`, `20240302@c.us`. Six recipients from two contacts, four of them fabricated out of an id column and a date column, and the counter at `:879` reports six rather than warning. A header row on its own is enough: `phone1,phone2` produces `1@c.us` and `2@c.us`. An email column survives verbatim as a chat ID through `bulkRecipients.ts:15-17`, and a quoted address is torn mid-value, so `628111222333,"221B Baker St, London, NW1 6XE"` also sends `221@c.us` and `16@c.us`.

The fix is a deletion. `parseBulkRecipients` already trims, drops blanks and de-dupes on the normalised chat ID, which is strictly stronger than the raw-string `Set` at `:214`, and `.trim()` absorbs CRLF. So the handler only needs to append the file's text:

```tsx
reader.onload = () => {
  const text = reader.result;
  if (typeof text !== 'string' || !text.trim()) return;
  setBulkRecipients(prev => (prev.trim() ? `${prev.trimEnd()}\n` : '') + text.trim());
};
reader.onerror = () => {
  setResponse({ success: false, timestamp: new Date().toISOString(), error: t('messageTester.fileReadError') });
};
```

I checked that against CRLF files, padded and blank lines, merging into an already-populated textarea, and re-picking the same file; behaviour is identical to yours minus the comma rule. It is about twenty lines shorter, and it also removes the dedup mismatch (where the textarea shows three lines while the count says one), the quoted-field tearing, and the `as string` cast. A single-column CSV still works because it is byte-identical to a TXT file, so `accept=".txt,.csv"` at `:867` stays honest. A multi-column row then fails exactly the way pasting one already fails today, which means the upload path adds no new failure mode.

If you would rather keep real multi-column CSV support, that is a larger feature: it needs a quote-aware cell split (the mirror of `escapeCsvCell` in `dashboard/src/utils/csv.ts:14-18`) and a rule for choosing one phone-shaped cell per row. Say the word and we can scope it, but it should not ride along here.

## Reuse `.browse-btn` for the control itself

Three separate problems in `MessageTester.tsx:848-868`, all fixed by one edit:

- `var(--bg-secondary)` at `:857` and `var(--border-color)` at `:858` are not defined anywhere in the dashboard. The tokens are `--bg-white` / `--bg-light` and `--border` (`dashboard/src/App.css:13-19`). An unresolvable `var()` with no fallback makes the declaration invalid at computed-value time, so the background falls to transparent and the `border` shorthand resets `border-style` to `none`. Rendered against the real stylesheets it computes to `rgba(0, 0, 0, 0)` with `border-style: none` in both themes, and inherits `--text-secondary` from `.message-tester .form-group label`, so it comes out as bare grey text the same colour as the RECIPIENTS label next to it. Not your invention, incidentally: `ApiKeys.tsx:290` already uses the same undefined token on main.
- A `<label>` wrapping a `display: none` input has no focusable element, so the picker is mouse-only and a screen reader sees static text with no role. There is precedent for this shape in the repo (`Plugins.tsx:1068-1074`, `Infrastructure.tsx:353-366`), which is why I am not treating it as a blocker, but the better pattern is 224 lines above your change in the same file.
- `className="upload-btn"` at `:849` matches no rule in any of the 21 dashboard CSS files, which is why the 12-property inline style has to carry everything.

`MessageTester.tsx:624-634` already solves all three:

```tsx
<button type="button" className="browse-btn" onClick={() => bulkFileInputRef.current?.click()}>
  <Upload size={14} /> {t('messageTester.bulkRecipientsUpload')}
</button>
<input ref={bulkFileInputRef} type="file" accept=".txt,.csv" style={{ display: 'none' }} onChange={handleBulkFileChange} />
```

`.message-tester .browse-btn` (`MessageTester.css:220-238`) already gives you inline-flex, `gap: 0.375rem`, `background: var(--bg-white)`, `border: 1px solid var(--border)`, the radius and a hover state, with tokens that exist. That deletes the whole inline object. It also fixes something you would not have seen: `marginRight: '6px'` on the icon at `:865` is a physical margin and does not flip for the two RTL locales, so in Hebrew and Arabic the icon ends up flush against the text with the 6px stranded on the outer edge. `gap` is direction-agnostic. If `.browse-btn`'s padding is too tall for the label row, add a small modifier class in the stylesheet rather than inline overrides; `MessageTester.css:362-373` shows the house pattern for resetting the `.form-group label` cascade in CSS, which is what your `textTransform` / `fontWeight` / `letterSpacing` overrides were correctly diagnosing.

## Two locale values

Eleven of the thirteen are good and match the vocabulary the catalogue already uses. Two do not:

- `fr.json:684` is "Télécharger CSV/TXT", but this file uses `télécharger` for download (`fr.json:91`, `fr.json:795`) and other verbs for upload (`fr.json:621` "Envoyer un fichier", `fr.json:853` "Téléverser .zip"). Suggest "Téléverser CSV/TXT".
- `ar.json:699` is "تحميل CSV/TXT", where every existing Arabic upload string uses رفع (`ar.json:636`, `ar.json:868`). `messageTester.uploadFile` renders on this same page, so two verbs for the same action appear a few controls apart. Suggest "رفع CSV/TXT".

## CHANGELOG and tests

`CONTRIBUTING.md:49-50` asks for an `[Unreleased]` entry when a change is user-visible. One line under the existing `### Added` at `CHANGELOG.md:10` covers it; `CHANGELOG.md:790` is the closest precedent in style. No `docs/` change is owed, since Message Tester has no wireframe in `docs/17-dashboard-design.md`, so leaving that box unticked was right.

On `CONTRIBUTING.md:47` and tests: if you take the reuse route above, the handler becomes plumbing and every rule worth asserting is already covered by the six cases in `dashboard/src/utils/bulkRecipients.test.ts`. Note that in the PR and I am happy. If any parsing survives in your version, put it in `dashboard/src/utils/bulkRecipients.ts` as an exported function and extend that existing test file with a header row, a multi-column row, a quoted field containing a comma, CRLF, and a merge into non-empty text. Our own docs point contributors at `*.spec.ts`, which is the backend convention; dashboard tests are `*.test.ts` under `node --test`. That is our documentation bug, not yours, and I will fix it separately.

## Smaller notes, none blocking

- No `reader.onerror`, and silent returns at `:200` and `:207`, so a failed read, an empty file and a mis-click all look the same. `messageTester.fileReadError` already exists in all 13 locales (`en.json:624`), so this is three lines and no new translation. It is in the snippet above. In fairness, three of the dashboard's four FileReader sites also lack one, so you did not break a convention here.
- No `file.size` guard before `readAsText` at `:218`. `accept` is only a picker hint. The media path rejects oversized files at `:242` using `MEDIA_UPLOAD_MAX_BYTES` (`:84`). Optional, and not free, since reusing `messageTester.fileTooLarge` would show "max 18 MB" next to a much smaller cap.
- Nothing in the upload path consults `BULK_MAX_RECIPIENTS` (imported at `:15`, enforced only at `:292`), so a 500-row file leaves Send greyed out. The hint at `:877-880` does show "(max 100) · Recipients: 500", and pasting 500 lines already reaches the same state, so this is pre-existing. Truncating at the cap would be a good follow-up.
- `he.json:693` uses the imperative where the catalogue uses the verbal noun (`he.json:630`, `he.json:862`). "העלאת CSV/TXT" would match.
- The new key sits beside its siblings in `en.json:653` but at the tail of `messageTester`, after the nested `batch` block, in all 12 translated files (`de.json:684`, `fr.json:684`, `ar.json:699`, `he.json:693`). Nothing enforces order, but moving it next to `bulkRecipients` keeps the files readable side by side.
- `marginBottom: '8px'` at `:842` replaces a `0.5rem` (`MessageTester.css:51`). Identical at the default root size, but it stops tracking the form if someone raises their browser font size. `'0.5rem'` fixes it.
- The description says the handler "extracts phone numbers". It extracts every delimited token, which is how column names end up in the textarea. Worth updating once the parsing changes.

Things I checked and would not change: BOM, CRLF, trailing newlines, spaces, `+` prefixes and Excel's leading apostrophe all come through correctly. The `e.target.value = ''` at `:194`, placed before the early return, is exactly right and matches `Infrastructure.tsx:361-365`. The functional `setBulkRecipients(prev => ...)` at `:209` is the correct choice inside an async callback and is why this handler does not need the staleness token the media reader uses. Merging rather than replacing is the right call. The commit is clean and its subject works as-is for the CHANGELOG line.

The blocker is the comma split; everything else is polish. Ping me when the rewrite is up.

harry0x:
I have applied all the requested changes. The UI now uses the .browse-btn pattern, the translations are fixed, and the comma split has been removed. The rewrite is up and ready for review! Please review it.

PR #1514, opened 2026-09-02
feat(api): page messages with an `after` keyset cursor

### LINKED ISSUE #1479: [Bug]: Offset paging over messages can skip or repeat rows when createdAt ties
`MessageService.findMessages` orders by `createdAt` alone and pages with `skip`/`take`:

```ts
// src/modules/message/message.service.ts:238
.orderBy('message.createdAt', 'DESC')
.skip(offset)
.take(limit)
```

`createdAt` is a `@CreateDateColumn()` (`src/modules/message/entities/message.entity.ts:121`), which on SQLite resolves to `datetime('now')` and stores whole seconds. A history sync writes many rows inside a single second, so ties are routine rather than rare.

Rows sharing a sort key have no guaranteed order between them, and nothing pins that order across two separate queries. A row can therefore move between the page already fetched and the page still to come, which means a message is silently absent from every page, or returned in two of them. Neither side raises an error.

Nothing paged this endpoint until now, so the defect has not surfaced in practice. #1464 is the first client to walk it with `offset=100`, `offset=200` and so on.

## The obvious fix makes it worse

Adding `.addOrderBy('message.id', 'DESC')` is the standard remedy and is wrong here: `id` is `@PrimaryGeneratedColumn('uuid')`, a random v4 with no relationship to insertion order. A random tiebreaker gives a newly inserted row a random rank inside its tie group, so it can sort ahead of rows an earlier page already returned and shift the window further than no tiebreaker at all. Whatever lands here needs measuring on both SQLite and PostgreSQL first, not just reasoning.

## Direction

A keyset cursor:

```sql
WHERE (createdAt, id) < (:lastCreatedAt, :lastId)
ORDER BY createdAt DESC, id DESC
LIMIT :limit
```

Position comes from the last row returned rather than from a row count, so a concurrent insert cannot shift it, and the database stops counting and discarding the skipped prefix on deep pages.

Scope beyond the query itself:

- a cursor parameter on `GET /sessions/{sessionId}/messages`, so a DTO and an OpenAPI snapshot change
- `openapi:check`, `check:sdk-*` and `check:contract-shapes` then require the JavaScript, Python, Go and Java clients plus the dashboard wire types to follow
- `offset` keeps working, so both modes coexist for at least one minor

Widening `createdAt` to millisecond precision is cheaper but only makes ties less likely, so it delays the problem rather than fixing it.

## Blocked until #1464 lands

That PR is mid-revision against the current `limit`/`offset` contract. Changing the contract underneath it would force a rewrite of work already in progress, and the defect has been latent long enough that waiting costs nothing.

rmyndharis:
Stacked on #1512, which gives these lists a total order. Review that one first; this diff shows only the cursor.

A total order stops two pages disagreeing about a tie group. It does not stop the **window** moving. `offset` addresses a position by count, so a message arriving mid-walk pushes every older row down one and the next page re-serves a row the previous one already returned. Measured on postgres:16, walking 500 rows in pages of 100 with one arrival between each page:

| walk | result |
| --- | --- |
| `offset=0,100,200,…` | a duplicate on every page boundary |
| `after=<last id>` | all 500 rows, exactly once |

## The cursor carries an id and nothing else

The anchor's sort key is resolved inside the statement:

```sql
(message.createdAt, message.id) <
  (SELECT anchor."createdAt", anchor."id" FROM messages anchor
   WHERE anchor."id" = :after AND anchor."sessionId" = :sessionId)
```

This is the part worth reviewing. A cursor that carries the timestamp instead round-trips wrong on **both** dialects, silently:

- SQLite holds two text shapes for the same instant and compares them as text. The live path takes TypeORM's inlined `datetime('now')` and stores 19 characters; the history projector stamps a JS `Date` and stores 23. A cursor built from a hydrated row re-selects the row it came from.
- node-postgres parses `timestamp` into a millisecond `Date`, but `NOW()` stores microseconds. Binding that back is neither `<` nor `=` the stored value, so the rest of the tie group is skipped.

Both reproduce the exact defect this is meant to remove. Keeping the value inside SQL makes the comparison consistent with the `ORDER BY` by construction.

## Behaviour

- `offset` is untouched and still works. `after` takes precedence when both arrive, so an SDK that always sends `offset=0` keeps working rather than starting to 400.
- `total` still counts rows matching the filters, not the post-cursor remainder, so it stays stable across a walk.
- An anchor naming no row in the session makes the row comparison `NULL` and returns zero rows, which reads exactly like the end of the history. The existence lookup that turns this into a `400` runs only on an empty page, so the happy path pays nothing.
- The response shape is unchanged. No `nextCursor` field, so `check:contract-shapes` stays green and no client is forced to follow.

## Cost

No schema change, no migration, no new index. SQLite pushes the row value into the existing `(sessionId, createdAt)` range: `SEARCH message USING INDEX IDX_399833392126349ef0b04b9bed (sessionId=? AND createdAt<?)`.

`openapi.json` grows by the one query parameter (9 lines). `after` added to the JavaScript, Python, Go and Java query types; PHP takes an untyped array and needs no change.

## Verification

`npm test`, `test:docs`, `test:e2e`, `test:cov`, `tsc --noEmit`, `lint`, `format:check`, `build`, `openapi:check`, `check:versions`, `check:dockerignore`, all four `check:sdk-*`, `check:contract-shapes`.

Three unit tests cover the predicate, the ignored `offset`, and the unknown-cursor `400`. The postgres-gated spec gained three more: that an offset walk duplicates a row under a concurrent arrival, that an `after` walk does not, and that a cursor naming a row in another session is rejected.

Not addressed here: the dashboard still pages by offset. #1464 owns that surface and can move to `after` on its own schedule, at which point `dashboard/src/utils/messagePages.ts` offset-drift compensation can go.

Closes #1479.


rmyndharis:
Superseded by #1517. This one targeted `fix/paged-list-total-order` and GitHub closed it when that branch was deleted on merge; a closed pull request cannot be retargeted. Same commit, rebased onto main.

PR #1518, opened 2026-09-02
fix(db): quote mediaPath in the partial index predicate

### LINKED ISSUE #1515: [Bug]: Message entity's mediaPath partial index breaks synchronize on PostgreSQL
`Message.mediaPath` carries a partial index whose predicate is unquoted:

```ts
// src/modules/message/entities/message.entity.ts:101
@Index('IDX_messages_mediaPath', { where: 'mediaPath IS NOT NULL' })
```

PostgreSQL folds an unquoted identifier to lower case, and the column is `mediaPath`, so `DataSource.synchronize()` fails while creating it:

```
QueryFailedError: column "mediapath" does not exist
    at RdbmsSchemaBuilder.createNewIndices
    at DataSource.synchronize
```

The migration that creates the same index quotes it correctly, so the two schema paths do not agree:

```sql
-- src/database/migrations/1786100000000-AddMessageMediaPathIndex.ts:28
CREATE INDEX IF NOT EXISTS "IDX_messages_mediaPath" ON "messages" ("mediaPath") WHERE "mediaPath" IS NOT NULL
```

The entity comment even states the intent: "The explicit name matches the migration that creates it on synchronize-disabled deployments, so both schema paths converge on one index."

## Impact

No deployment is broken today. The data connection runs its migration chain rather than `synchronize`, so the failing path is never taken in production. It bites anyone who points a TypeORM `DataSource` carrying the `Message` entity at PostgreSQL with `synchronize: true`, which is the natural way to write a Postgres-backed test harness against the real entity. Found while building one.

SQLite is unaffected: it does not case-fold the bare identifier.

## Fix

Quote the column in the entity predicate so it matches the migration:

```ts
@Index('IDX_messages_mediaPath', { where: '"mediaPath" IS NOT NULL' })
```

Worth a check for the same shape on other partial indexes at the same time.

### LINKED ISSUE #1516: [Bug]: Message-list inline media budget is per response, so a paged walk multiplies it
`MESSAGE_LIST_INLINE_MEDIA_BUDGET_BYTES` bounds inline base64 per **response**, not per walk. Its own docblock states the assumption it was sized under:

```
// src/modules/message/message.service.ts
 * Aggregate budget for the inline base64 media ONE message-list response may carry [...]
 * the dashboard requests the maximum page size with no way to ask for less
```

That last clause no longer holds. A client that pages the endpoint pulls up to `8 MiB x pages`, and nothing bounds the number of pages. The per-page allowance compounds it: `spendInlineMediaBudget` deliberately lets the newest payload through whatever its size, so each page can exceed the budget by one payload, itself bounded only by `MEDIA_DOWNLOAD_MAX_BYTES` (50 MiB, about 68 MiB once base64-encoded).

Walking a media-heavy chat 20 pages deep is therefore ~160 MiB of base64 on the wire, in a shape no single response would ever be allowed to return.

## Why now

Paging this endpoint was theoretical when the budget was written. It is not any more: #1464 walks it from the dashboard, #1514 adds a cursor for walking it safely, and the MCP `MessageList` tool shares the same path and budget.

## Options, none obviously right

- Leave it and document that a paged walk is not budgeted, so a caller wanting media pages with `metadata.media.data` omitted has a way to ask.
- An explicit `inlineMedia=false` query parameter, letting a walking client opt out and fetch bytes from `GET /messages/:chatId/:messageId/media` on demand. This is probably what a bulk exporter wants anyway.
- Shrink the default budget now that a caller can page rather than needing everything in one shot.

Filing rather than fixing because the right answer depends on what walking clients should get by default, which is a product call.

rmyndharis:
Two independent follow-ups from #1479, one commit each.

## Quote `mediaPath` in the partial index predicate (#1515)

PostgreSQL folds a bare identifier to lower case, so the entity's `where: 'mediaPath IS NOT NULL'` made `DataSource.synchronize()` fail:

```
QueryFailedError: column "mediapath" does not exist
    at RdbmsSchemaBuilder.createNewIndices
```

The migration creating the same index already quoted it, so the two schema paths disagreed despite the entity comment stating they should converge.

No deployment was broken. The data connection runs its migration chain rather than `synchronize`, so the failing path was never taken in production. It blocked anyone pointing a TypeORM `DataSource` carrying the `Message` entity at PostgreSQL with `synchronize: true`, which is the natural shape for a Postgres test harness built on the real entity.

Verified on postgres:16 and better-sqlite3, before and after:

| | before | after |
| --- | --- | --- |
| postgres | `column "mediapath" does not exist` | `WHERE ("mediaPath" IS NOT NULL)`, matching the migration |
| better-sqlite3 | succeeded (no case folding) | succeeded |

No migration needed: booting an existing SQLite deployment three times against the new entity leaves the old index in place rather than recreating it, and the two predicates are semantically identical there.

It is the only partial index in the repo, so there is nothing else to sweep.

## `inlineMedia=false` on the message list (#1516)

`MESSAGE_LIST_INLINE_MEDIA_BUDGET_BYTES` bounds inline base64 per **response**, and its docblock names the assumption it was sized under: "the dashboard requests the maximum page size with no way to ask for less". #1517 ended that. A walk pulls the full budget on every page, so twenty pages of a media-heavy chat is roughly 160 MiB of base64, in a shape no single response would ever be allowed to return. The per-page allowance compounds it: `spendInlineMediaBudget` lets the newest payload through whatever its size, bounded only by `MEDIA_DOWNLOAD_MAX_BYTES`.

`inlineMedia=false` lets a walking caller ask for the rows without the bytes:

```
GET /sessions/s1/messages?chatId=x&limit=100&after=<id>&inlineMedia=false

{ "messages": [ { "id": "...", "metadata": { "media": {
      "mimetype": "image/jpeg", "omitted": true, "sizeBytes": 184320 } } } ],
  "total": 4211 }
```

That `{ omitted, sizeBytes }` marker is the one the engine already writes for inbound media over the download cap, so it is not a new shape to learn, and the bytes stay available from `GET /messages/:chatId/:messageId/media`.

The flag picks the budget rather than adding a code path: `0` already means "never inline" and grants no single-payload allowance, which is exactly what an opted-out caller asks for.

**It is an opt-out**, unlike every other boolean on this controller, so the parse reads the same string pair the other way round: only `false` and `0` disable, and an absent or unparseable value keeps today's behaviour. Inverting that would quietly strip media from every default read and no service-level test could tell, since the service receives a boolean and cannot see who set it. A controller test pins both directions.

Purely additive: the response shape is unchanged, so `check:contract-shapes` stays green and no client is forced to follow. `inlineMedia` added to the JavaScript, Python, Go and Java query types; PHP takes an untyped array.

Not addressed: the MCP `MessageList` tool shares this path and still inlines. Whether an agent should receive base64 in its context window is a different question from bounding a walk, and is better decided on its own.

## Verification

`npm test`, `test:docs`, `test:e2e`, `test:cov`, `tsc --noEmit`, `lint`, `format:check`, `build`, `openapi:check`, `check:versions`, `check:dockerignore`, all four `check:sdk-*`, `check:contract-shapes`, the postgres lane, and the JavaScript SDK build.

Closes #1515. Closes #1516.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `GET /sessions/{sessionId}/messages` accepts `inlineMedia=false`, which omits every inline media
+  payload and leaves each row's `{ omitted, sizeBytes }` marker plus the media endpoint. The budget
+  bounds one response, so a paged walk pulls up to 8 MiB of base64 per page; a client reading many
+  pages can now ask for the rows without the bytes
+  ([#1516](https://github.com/rmyndharis/OpenWA/issues/1516)).
 - `GET /sessions/{sessionId}/messages` accepts `after`, a keyset cursor holding the `id` of the last
   message of the previous page. It anchors the window to a row instead of a count, so a message
   arriving mid-walk can no longer shift it and make a page repeat or skip rows. `offset` keeps
--- a/docs/06-api-specification.md
+++ b/docs/06-api-specification.md
@@ -1207,13 +1207,14 @@ Get persisted message history for a session from the local DB (paginated, filter
 
 **Query parameters**
 
-| Name   | Type    | Required | Default | Description                                                                                                                                                                                                                             |
-| ------ | ------- | -------- | ------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| chatId | string  | No       | —       | Filter by chat ID. Matched across `@c.us` / `@s.whatsapp.net` dialects via the lid-mapping table.                                                                                                                                       |
-| from   | string  | No       | —       | Filter by sender. A phone also matches any lid that resolves to it.                                                                                                                                                                     |
-| limit  | integer | No       | 50      | Clamped to `[1,100]`; a non-finite value falls back to 50.                                                                                                                                                                              |
-| offset | integer | No       | 0       | Clamped to `>=0`; a non-finite value falls back to 0.                                                                                                                                                                                   |
-| after  | string  | No       | —       | Keyset cursor: the `id` of the last message of the previous page. Anchors the window to a row rather than a count, so a message arriving mid-walk cannot shift it. Takes precedence over `offset`. Unknown in this session gives `400`. |
+| Name        | Type    | Required | Default | Description                                                                                                                                                                                                                             |
+| ----------- | ------- | -------- | ------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| chatId      | string  | No       | —       | Filter by chat ID. Matched across `@c.us` / `@s.whatsapp.net` dialects via the lid-mapping table.                                                                                                                                       |
+| from        | string  | No       | —       | Filter by sender. A phone also matches any lid that resolves to it.                                                                                                                                                                     |
+| limit       | integer | No       | 50      | Clamped to `[1,100]`; a non-finite value falls back to 50.                                                                                                                                                                              |
+| offset      | integer | No       | 0       | Clamped to `>=0`; a non-finite value falls back to 0.                                                                                                                                                                                   |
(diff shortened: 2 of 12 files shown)


rmyndharis:
Closing this: it was opened without authorization and bundles two unrelated changes. The mediaPath index fix will be reopened on its own branch with a regression test that actually fails without it. The inline-media opt-out (#1516) will be handled separately.

PR #1547, opened 2026-09-05
feat(commerce): surface inbound order and product messages

m7fz7:
## What

Inbound WhatsApp Business commerce messages currently arrive as a bodyless `unknown`. A cart placed from the catalog and a shared product card each carry an id that identifies them, and the generic path drops it.

This maps `orderMessage` -> a new `order` type and `productMessage` -> a new `product` type on both engines, and carries the ids through on `IncomingMessage`:

- `order { orderId, token? }`
- `product { productId, title?, description?, businessOwnerJid? }`

## What this actually ships

A typed message, a non-empty `body`, and a correlation id the caller redeems elsewhere. Concretely:

- **The order ids are not a handle this project can spend.** `getOrderDetails` is not exposed on either engine, so `orderId` + `token` are what you hand to whatever does resolve line items. The token is scoped to that one order.
- **`productId` resolves only against the session's own catalog.** Both catalog routes that take a `productId` query the session's catalog, so a product shared from another business will not resolve through them. `businessOwnerJid` says whose catalog it came from.
- **The ids are genuinely unreachable today.** There is no raw passthrough on `IncomingMessage`, so without this the message is a bodyless `unknown` and the ids are gone.

The payloads are not persisted — `buildMessageMetadata` writes only `media`, `quotedMessage` and `call` — so, like `location` today, an order token exists on the live dispatch (`message.received`, websocket) and the live chat-history read, never on `GET /messages`.

## Field selection

Only fields **both** engines populate are carried. Baileys additionally reports `itemCount`, `status`, `surface`, `sellerJid`, `total`, `currency`, `price`, `retailerId` and `url`; those are omitted. They have no consumer, whatsapp-web.js cannot fill them, and carrying them would commit five SDK surfaces plus docs/29 to keeping them true. `surface` in particular carries no information — `OrderSurface` has exactly one member.

## The catalog case

There is no `catalogMessage` content type. Sharing a whole catalog sends a `productMessage` carrying the `catalog` arm (`ICatalogSnapshot`) instead of `product`. It has no product id, so it would otherwise land as a `product` message with no `product` object and an empty body — exactly what the missing-id guard exists to prevent. It is detected (`isBaileysCatalogShare`) and stays `unknown`, with the catalog title flattened into `body`.

## Also

- The Baileys `body` falls back to the customer's order note, else the order title, else the product's accompanying text, else the product or catalog title. **The fallback is Baileys-only** — the whatsapp-web.js mapper assigns `msg.body` as-is, which is already populated there.
- `businessOwnerJid` goes through the same `normalizeJid` as every other JID on the payload, so one object never mixes `@s.whatsapp.net` or `@lid` with `@c.us` fields.
- Both types are accepted by the webhook and automation-rule message-type filters, and offered in the dashboard filter builder.
- Contract surfaces updated together: `ChatHistoryMessageDto` gains `ChatHistoryOrderDto` / `ChatHistoryProductDto` (alongside the existing `call` and `location` sub-DTOs), `openapi.json` regenerated, and `order` / `product` added to the Go, Java, TypeScript and dashboard `MessageType` enums and `ChatHistoryMessage` shapes.
- No response or status-code changes: `order` and `product` are additions to the `type` union, and both payload objects are optional.

## Testing

`npm run build`, `npm test`, `npm run lint`, `npm run check:contract-shapes` (332 pairs conform), `npm run openapi:check`, `check:sdk-routes` / `-coverage` / `-events` / `-docs`, `npm run test:docs` and `npm --prefix dashboard run build` all pass.

New unit cases in `baileys-message-mapper.spec.ts` and `message-mapper.spec.ts` cover the type mapping on both engines, the order with and without a token, the product's descriptive fields, the missing-id case, a non-commerce message that must not produce a commerce payload, the catalog-share detection and its body fallback, and the `businessOwnerJid` normalization.

Manually verified against a live Baileys session by placing a catalog order and sharing a product card from a phone.

## Not in scope

Resolving an order's line items is a new REST surface and belongs in its own PR. This one is the inbound capture that makes it possible.


rmyndharis (changes requested):
The extraction work here is accurate and I checked it against the pinned library rather than the diff. Every proto field name matches `@whiskeysockets/baileys` 7.0.0-rc14, both enum tables are real (`INQUIRY=1`, `ACCEPTED=2`, `DECLINED=3`, `CATALOG=1`), the thousandths conversion and the `Long` arm are right, and the unpriced-currency case is a genuine trap that you handled correctly. On whatsapp-web.js, `MessageTypes.ORDER` and `MessageTypes.PRODUCT` both exist and every producer hands the raw `Message` to `buildIncomingMessageBase`, so both engines really do populate the ids. Two `MessageType` members plus the body fallback is also the right shape and matches how `poll` and `masked` were introduced.

Six things need to change before this can land.

**1. CI, and what is waiting behind it**

Lint is red on `Check OpenAPI snapshot is up to date`, which skipped the five gates after it. Running them locally against your head: regenerating the snapshot clears that step and `check:contract-shapes` then fails on two clients.

```
sdk/go (types_*.go):   ChatHistoryMessage <-> ChatHistoryMessageDto
  "type": hand enum(14 members), contract enum(16 members, adds order, product)
sdk/java (model/*.java): same pair, same drift
```

* `npm run openapi:export` and commit `openapi.json` (two lines)
* `sdk/go/types_message.go:440` add `MsgOrder` and `MsgProduct`
* `sdk/java/src/main/java/com/rmyndharis/openwa/model/MessageType.java:18` add `@SerializedName("order") ORDER` and `@SerializedName("product") PRODUCT`
* `sdk/javascript/src/types.ts:537` add both members. No gate covers the TypeScript union, but without it `msg.type === 'order'` is a TS2367 error for every typed consumer.

With those four, `check:contract-shapes` reports 332 pairs conform. `check:sdk-routes`, `check:sdk-coverage`, `check:sdk-events` and `check:sdk-docs` are unaffected.

**2. `catalogMessage` does not exist**

`baileys-message-mapper.ts:52` documents it as deliberately unmapped and `baileys-message-mapper.spec.ts:46` asserts it maps to `unknown`, but there is no such content type in the pinned proto. A catalog share is a `productMessage` carrying `catalog` (`ICatalogSnapshot`) instead of `product` (`WAProto/index.d.ts:8470,8472). It reaches `case 'productMessage'`, gets `type: 'product'`, and `extractBaileysCommerce` returns `{}` because `snapshot?.productId` is undefined. The result is a `product` message with no `product` object and an empty body, which is the outcome the missing-id guard exists to prevent. Handle the `catalog` arm, correct the comment, and drop the spec row.

**3. The DTO does not match the docs**

`docs/06-api-specification.md:1304` now promises `order` and `product` on the chat-history response, but `ChatHistoryMessageDto` (`src/modules/message/dto/message-responses.dto.ts:224-300`) declares neither, unlike `call` and `location`, which each have their own sub-DTO. Either add the two `@ApiPropertyOptional` sub-DTOs, or narrow the doc sentence to the `message.received` payload, which is where the fields actually reach a consumer.

**4. Trim the Baileys-only block**

`itemCount`, `status`, `surface`, `sellerJid`, `total`, `currency`, `price`, `retailerId` and `url` have no consumer, appear in no client type, and commit the project to keeping them true across five SDK surfaces and `docs/29`. `surface` carries no information at all: `OrderSurface` has exactly one member. `order { orderId, token }` and `product { productId, title, description, businessOwnerJid }` are what both engines populate, and dropping the rest removes the four helpers and two enum tables with them.

If the priced fields stay, align the product shape with `Product` at `src/engine/interfaces/whatsapp-engine.interface.ts:494`, which already models a catalog product as `id` / `name` / `description` / `price` / `currency` / `retailerId` / `url`. Two vocabularies for the same object in one interface file is worse than either one on its own.

**5. Two smaller items**

* `sellerJid` (`baileys-message-mapper.ts:255`) and `businessOwnerJid` (`:275`) are passed through verbatim, while every other JID on the same payload goes through `normalizeJid`. An order can currently emit `@s.whatsapp.net` or `@lid` next to `@c.us` fields in one object.
* The CHANGELOG sentence about the body fallback is unqualified, but the fallback lives only in `extractBaileysBody`. The whatsapp-web.js mapper still assigns `msg.body` as-is.

**6. The framing overstates what ships**

The description and `whatsapp-engine.interface.ts:118` present the ids as the handle for `getOrderDetails`, but that is not exposed on either engine, and both routes that take a `productId` resolve it against the session's own catalog, so a product shared from another business will not resolve through them. The capture is still worth having, since the ids are genuinely unreachable today and there is no raw passthrough on `IncomingMessage`. Describe what actually lands: a typed message, a non-empty body, and a correlation id the caller redeems elsewhere.

One note that is not a request: the payloads are not persisted, because `buildMessageMetadata` writes only `media`, `quotedMessage` and `call`. That matches how `location` already behaves, so it is consistent, but it does mean the order token exists only on the live dispatch and never on `GET /messages`.

m7fz7:
Thanks — all six addressed, rebased onto main (the CHANGELOG conflict was the only real one).

**1. CI.** `openapi.json` regenerated, and `order`/`product` added to `sdk/go/types_message.go`, `MessageType.java`, `sdk/javascript/src/types.ts` and the dashboard's `MESSAGE_TYPES`. Adding the two sub-DTOs in (3) then pulled `ChatHistoryMessage` into it as well, so Go gained `MessageOrder` / `MessageProduct`, Java gained the two nested records, and the TS and dashboard interfaces gained both fields. `check:contract-shapes` reports 332 pairs conform. Two things the root suite doesn't cover turned up after the first push: the JS SDK's own `wire-contract.test-d.ts` needed the same members and fields, and `format:check` wanted a CHANGELOG reflow. Both fixed; the board is green.

**2. `catalogMessage`.** You're right, it doesn't exist — I took it from a community mapping table and never checked it against the proto. A catalog share is a `productMessage` carrying `catalog`, so it reached `case 'productMessage'` and produced exactly the bodyless `product` the id guard is supposed to prevent. Now detected (`isBaileysCatalogShare`) and typed `unknown`, with the catalog title flattened into `body` so it isn't empty either. Comment corrected, spec row replaced with one that asserts the real shape.

**3. DTO.** Added `ChatHistoryOrderDto` / `ChatHistoryProductDto` alongside the existing `call` and `location` sub-DTOs rather than narrowing the doc — both engines populate these on the history path too, so the doc sentence was right and the DTO was the thing that was wrong. Also added the pair to the `message.received` payload sketch, since that's where they actually reach a consumer.

**4. Trimmed.** Dropped to `order { orderId, token }` and `product { productId, title, description, businessOwnerJid }`. That removed the four helpers, both enum tables and the `ProtoAmount` type. The unpriced-currency handling went with them, which is a little sad, but you're right that it was defending a field nothing consumes — and `surface` having one member is a fair hit. Keeping one vocabulary this way also sidesteps the `Product` clash.

**5.** `businessOwnerJid` now goes through `normalizeJid` in `buildIncomingMessageFromBaileys` (`sellerJid` left with the trim), with a test that a `@s.whatsapp.net` owner comes back `@c.us`. CHANGELOG sentence now says the fallback is Baileys-only.

**6. Framing.** Rewrote the description, the two interface doc comments and the §6 prose around what actually ships: a typed message, a non-empty body, and a correlation id redeemed elsewhere. I also called out that `productId` only resolves against the session's own catalog, and added your note about the payloads not being persisted — it's worth a reader knowing the token is live-dispatch only, even though matching `location` makes it consistent.

rmyndharis (approved):
All six verified against the head, not the reply: the catalog-share guard, the history sub-DTOs, the trim to the two id pairs, the businessOwnerJid normalization with its test, and the reframed description and docs all check out, and the board is green.

The wire-contract and CHANGELOG reflow catches beyond the requested list were the right calls, and routing both filter surfaces through the one MESSAGE_TYPE_FLAGS change keeps it a two-line diff. Merging.

Issue #1546, opened 2026-09-05
[Bug]: WhatsApp connection gets disconnected after a while


OdaiAhmed99:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm on the latest released version (or noted my version below)

### OpenWA version

v0.23.3 · 20260904

### Deployment

Docker Compose

### Database

SQLite (default)

### WhatsApp engine

baileys

## Bug Description

The WhatsApp connection gets disconnected automatically after running for a while.

### Environment
- WhatsApp engine: `ENGINE_TYPE=baileys`
- Deployment: Docker Compose

### Expected Behavior
The WhatsApp connection should remain connected and stable without disconnecting.

### Actual Behavior
The connection works normally at first, but after some time it becomes disconnected.

### Additional Information
This happens with the Baileys engine:
`ENGINE_TYPE=baileys`
<img width="1413" height="701" alt="Image" src="https://github.com/user-attachments/assets/16c7c0b0-d0c1-4f4b-bf2d-ef4968ee2253" />

### Steps to reproduce

1. Set `ENGINE_TYPE=baileys`.
2. Start OpenWA and connect a WhatsApp session.
3. Confirm that the session is connected and working normally.
4. Leave the session running for some time without manually disconnecting it.
5. The session eventually changes to `disconnected`.

### Expected behavior

The WhatsApp connection should remain connected and stable without disconnecting.

### Relevant logs

```shell

```

rmyndharis:
Thanks for the report, and for the screenshot: it carries more than it looks like it does, and it points somewhere different from a plain disconnect.

The badge reading "Starting..." on that card is session status `initializing`, not `disconnected`. On the Baileys engine a session that is still linked sits at `initializing` for the entire reconnect backoff after a dropped socket, and the card renders the QR placeholder in that state even when no pairing is pending. So the picture is consistent with a session cycling through reconnect attempts, and equally consistent with the dashboard's event feed having stopped updating the card while the backend is fine. Those have different fixes and I cannot tell them apart from here.

It also explains the empty logs box. An engine-level reconnect writes one line and no webhook, so there is genuinely little to copy out.

Four things would settle it. Any subset helps, and the second is the most valuable.

1. Container logs covering a window where the symptom happened:

```
docker compose logs --since 24h openwa-api | grep -iE "connection dropped|reconnect attempt failed|Cleared Baileys auth state|startup_reset"
```

[...]

PR #1567, opened 2026-09-08
perf: core caching, typing bypass, dashboard vendor chunking, and slim docker target

AnandkumarMall:
## Description

Comprehensive performance optimizations, cache efficiency improvements, and production deployment slimming across the OpenWA gateway:

1. **Backend Cache & Latency Reductions**:
   - **Redis Wire Ping Elimination** (`CacheService`): Synchronously validates client readiness (`client.status === 'ready'`) instead of executing `await this.ping()` before every read/write. `ping()` is preserved for health and readiness probes.
   - **Immediate API Message Dispatch** (`MessageSendService`): Changed `SIMULATE_TYPING` to opt-in (`env.SIMULATE_TYPING === 'true'`), stripping the default 500ms–5000ms delay per outgoing API message.
   - **In-Memory API Key LRU Cache** (`AuthService`): Implemented a 60s TTL LRU cache (1,000 max entries) with entity cloning (`cloneApiKey`) to eliminate redundant DB lookups on every authenticated API request while isolating entity state. Includes auto-invalidation on key update/revocation/deletion.
   - **In-Memory Webhook Cache** (`WebhookDeliveryService`): Active session webhooks are cached in memory (60s TTL) with invalidation hooks across webhook CRUD operations, eliminating per-event DB queries on high-frequency dispatches.
   - **Session Phone Caching** (`MessageSendService`): Caches session phone resolution in memory to prevent repeated DB queries during outgoing message persistence.
   - **Media Inline Budget Spend** (`MessageService`): Pure-JS ASCII byte estimation calculation in `spendInlineMediaBudget` to reduce V8-to-C++ context switching.

2. **Dashboard Bundle Optimization**:
   - Configured `build.rollupOptions.output.manualChunks` in `dashboard/vite.config.ts` to separate `vendor-react`, `vendor-charts`, and `vendor-ui`.
   - Initial entry bundle payload (`index.js`) reduced from **319.6 kB to 132.2 kB (42.3 kB gzipped)** — a **58.6% reduction in initial load weight**.

3. **Multi-Stage Slim Docker Target & Env Docs**:
   - Added `production-slim` build target in `Dockerfile` (~180 MB vs ~1.2 GB) for headless Baileys and API-only deployments, omitting Chromium for Testing, X11/GL, and ffmpeg while preserving the default `production` target for `whatsapp-web.js` users.
   - Documented `# BAILEYS_AUTH_DIR=./data/baileys` in `.env.example` under engine settings.

## Type of Change

- [x] Bug fix / Performance improvement
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist

- [x] Tests added/updated: All core backend suites (`cache.service.spec.ts`, `feature-flags.spec.ts`, `env-precedence.spec.ts`, `message-send.service.spec.ts`, `auth.service.spec.ts`, `webhook-delivery.service.spec.ts`, `engine.factory.spec.ts`) pass.
- [x] Dashboard tests: All 393 unit tests pass (`npm --prefix dashboard test`).
- [x] Dockerfile patchers verified (`scripts/dockerfile-patchers.spec.js` and `npm run check:dockerignore` pass).
- [x] Backend build: `npm run build` compiles with zero TypeScript errors.
- [x] Dashboard build: `npm run dashboard:build` compiles cleanly with vendor chunks.
- [x] Self-reviewed.

## Related Issues
Partially addresses #1546 (Baileys credential persistence & slim container sizing without OOM)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -118,6 +118,7 @@ STATUS_SEED_ON_READY=false
 
 # Engine-specific settings
 # SESSION_DATA_PATH=./data/sessions
+# BAILEYS_AUTH_DIR=./data/baileys
 # PUPPETEER_HEADLESS=true
 # Browser flags, comma- or space-separated. NOTE: this REPLACES the default list rather than adding to
 # it, so repeat the flags you still want — dropping --no-sandbox stops Chromium launching in a
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,7 +57,82 @@ COPY . .
 # ship dead compiler metadata in every image.
 RUN npm run build && npm run dashboard:ci -- --include=dev && npm run dashboard:build && rm -f dist/*.tsbuildinfo
 
-# ===== Stage 2: Production =====
+# ===== Stage 2: Production Slim (Baileys / Headless API) =====
+# A lightweight production image (~180 MB vs ~1.2 GB) for operators deploying Baileys or
+# headless API workloads that do not require Chromium, X11, ffmpeg, or Puppeteer browser binaries.
+# Build explicitly via `docker build --target production-slim -t openwa:slim .`
+FROM docker.io/node:22-slim@sha256:d649c27dae7ba0137b3cef5dd75baa422c08dc3d9e3fc0c23dfb172dc3cc6436 AS production-slim
+
+ENV NODE_ENV=production
+
+# Minimal runtime packages: dumb-init (PID 1 signal forwarding), gosu (root->openwa drop),
+# curl (health probes), ca-certificates (outbound TLS), sqlite3 (online database snapshots).
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    ca-certificates \
+    curl \
+    dumb-init \
+    gosu \
+    sqlite3 \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+# Non-root system user (deterministic UID 10001 matching production stage)
+RUN useradd -r -s /bin/false -u 10001 openwa
+
+# Copy package manifests and patches for clean production install
+COPY package*.json ./
+COPY scripts/postinstall.js ./scripts/
+COPY scripts/patch-wwebjs-newsletter-preview.js \
+     scripts/patch-wwebjs-status.js \
+     scripts/patch-wwebjs-ready-sync.js \
+     scripts/patch-wwebjs-participant-arity.js \
+     scripts/patch-wwebjs-block.js \
+     scripts/patch-wwebjs-group-description.js \
+     scripts/patch-baileys-appstate.js \
+     scripts/patch-baileys-newsletter-create.js \
+     ./scripts/
+
+RUN PUPPETEER_SKIP_DOWNLOAD=true npm ci --omit=dev \
+    && node scripts/patch-wwebjs-newsletter-preview.js \
+    && node scripts/patch-wwebjs-status.js \
+    && node scripts/patch-wwebjs-ready-sync.js \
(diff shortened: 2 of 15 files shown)


AnandkumarMall:
Friendly ping! Let me know if there's anything else needed from my side or if this is ready for review. Thanks!

rmyndharis:
Thank you for the time you put into this. I have closed it, and the reasons are specific, so here they are in full.

The PR combines eight independent changes. Several change behaviour the description does not mention, none of the performance claims comes with a measurement, and none can be merged as it stands:

- **Typing simulation default.** `SIMULATE_TYPING` is on by default as an anti-ban measure (`docs/16-risk-management.md`, `.env.example`), and deployments that want immediate sends can already set `SIMULATE_TYPING=false`. The change turns it off for every deployment that never set the variable, while the docs still say it is on. The pause also applies only to single text sends.
- **API key cache.** `ApiKeyUsageTracker.record()` decides whether a usage write is due from the entity's freshly loaded `lastUsedAt`. A cache hit hands it a snapshot from fill time, so for a key last used more than 60 s earlier every request in the window issues an UPDATE and writes `snapshot + 1`: write coalescing is lost and `usageCount` stops advancing. Revocation, role, IP and session-scope changes also become local to one process for up to 60 s, and a lookup that read the row before a revoke committed can fill the cache after the revoke invalidated it. The query being cached is a unique-index read on the local SQLite database.
- **Session phone cache.** Nothing invalidates it on logout or relink, which is exactly when the session lifecycle clears and rewrites `phone`, and its bare `catch` turns a transient database error into a saved row instead of a failed call.
- **Webhook cache.** This one has a plausible hot-path benefit, but `POST /api/infra/import-data` replaces the `webhooks` table with raw SQL outside `WebhookService`, so after a restore, removed webhooks keep receiving events for up to 60 s and restored ones receive none.
- **Redis ping.** The only production caller of these `CacheService` methods is the overview stats write, already memoized for 30 s, so the saving is at most one ping every 30 s. The `client.status === 'ready'` check in the description is not in the code.
- **Dashboard chunks.** Every route and `DashboardCharts` are already lazy. With the new `manualChunks`, `index.html` modulepreloads `vendor-charts` (React core plus recharts), so startup JavaScript grows from about 345 kB to about 806 kB (110 kB to 243 kB gzip) in a build of this branch.
- **Slim image.** `docker build --target production-slim` fails: `npm ci --omit=dev` runs without `--ignore-scripts`, better-sqlite3 triggers `node-gyp rebuild`, and the stage has no Python. The stage also skips `patch-wwebjs-201832.js`, keeps the default `ENGINE_TYPE=whatsapp-web.js` with no browser, and drops ffmpeg and the PostgreSQL client that media conversion and `backup.sh` use; the production stage does not pin UID 10001 either. No workflow builds this target.
- **Checks.** On this branch merged with `main`, `npm run lint` reports seven errors in `cloneApiKey` and `validateApiKey`, and the `./src/common/cache/` statements coverage drops to 41.98% against its 42% floor.

Smaller points: the `.env.example` addition duplicates the existing `BAILEYS_AUTH_DIR` entry, #1546 is a Baileys connection attempt that never retried and is unrelated to these changes, and the Windows test tweaks are unrelated to performance.

Changes along these lines are welcome as separate pull requests, one change each, with before and after numbers for any performance claim. A webhook cache needs invalidation on the import path, with a test for it. A slim image needs the target built and smoke-tested in CI, `ENGINE_TYPE=baileys` set in the image, and install and patch steps that match the production stage.

PR #1623, opened 2026-09-15
 Improve dashboard UX, session reliability, authentication integration, and template workflows

AbdoHussein1998:
## Summary

This PR consolidates the `FinalV1` improvements while keeping the existing OpenWA architecture and API behavior intact.

* Improves dashboard authentication, role awareness, login flow, and session-scoped access handling.
* Extends frontend template management/composer functionality and its backend API integration.
* Improves WhatsApp QR/session readiness handling for devices that require significantly longer to initialize.
* Adds safer real-time session-feed subscription/fallback behavior and more reliable connection-state synchronization.
* Keeps the dashboard version aligned with the root application release version and improves the login/dashboard UX.
* Adds and updates tests around session readiness, subscriptions, templates, authentication, and related edge cases.

### Main Roles Added

* **Admin:** Unchanged.
* **Operator:** Similar to Admin, but does not have access to Infrastructure and cannot delete Admins.
* **Viewer:** Can view everything but has no ability to make changes or interact with resources.
* **Team Leader:** Can create and fully manage Sessions and Agents. Team Leaders can be assigned by the Admin.
* **Agent:** Can only control their assigned Sessions. They cannot delete Sessions, and Sessions can be assigned to them by higher-level roles.

### Follow-up Work

* Prevent Agents from being assigned multiple Sessions.
* **Localization:** Frontend changes have not been fully translated. All screens and locales should be reviewed to ensure translation keys and displayed text are consistent.
* **Database Relationships:** Proper **FK constraints and relational integrity still need to be established**. Several tables are currently loosely connected and should be explicitly tied together.

I just wish I could’ve done something useful to help, or at least felt like I tried.

With my best regards, 😊


rmyndharis:
Thanks for putting this together, and for the heads-up on #1484 about what you were building.

First, what is actually in the branch, because the title undersells it. Beyond the dashboard and template work it adds two values to the `ApiKeyRole` enum, two new modules (`src/modules/teamleader/`, `src/modules/access-control/`), five migrations across both database connections, a session ownership model, and a swap of the runtime browser from Chrome for Testing to Brave with a per-session profile manager. That is five independent product decisions on one branch, 221 files and +75,544/-7,810, and several of them change behaviour for every existing deployment.

I am closing it, and I want to be precise about why, because most of the individual pieces are not the problem.

**Authorization widens for keys that already exist.** Every `/auth/api-keys` route moves from `@RequireRole(ApiKeyRole.ADMIN)` (`src/modules/auth/auth.controller.ts:37,71,96,121,164,179` on main) to `@RequireCapability(ApiCapability.API_KEY_MANAGE)` (`:134,194,222,248,347,439,512` here), and `ROLE_CAPABILITIES[OPERATOR]` holds that capability (`src/modules/auth/auth.service.ts:219`). `createApiKey` takes `dto.role` verbatim, and the only role check on the way in rejects the two new management roles (`:663-675`), so an operator key can mint an admin key for itself. The plugin surface moves the same way (14 admin-guarded handlers become one class-level `PLUGIN_MANAGE` at `src/modules/plugins/plugins.controller.ts:41`), as do the audit log and the stats routes. Operator is the role this project tells people to use for MCP agents and integrations, so those are exactly the keys that would widen, and it happens on upgrade with no opt-in.

**Every whatsapp-web.js session loses its login.** `LocalAuth` becomes `NoAuth` with an explicit `userDataDir` (`src/engine/adapters/wwebjs-lifecycle.ts:16,419,431`), with no migration from the existing profile and no way back, since the new profile is not readable by the released code. The profile is also keyed by session name: `src/modules/session/session-engine-lifecycle.service.ts:540` passes `sessionId: session.name`, where main passes the id with the reason written above the call. Keying auth state on the name is the collision 0.23.5 closed, and it is worse here than it was there, because this profile is the only copy of the credentials rather than a cache: two sessions whose names differ only in case would share one WhatsApp login.

**The published image loses arm64.** `ARG TARGETARCH` and every per-architecture branch are gone, and `brave-browser` is installed from a third-party apt repository whose signing key is fetched during the build. The release pipeline builds, scans and boots both `linux/amd64` and `linux/arm64` and promotes only when both pass; the arm64 branch exists because Chrome for Testing has no Linux arm64 build. An unpinned browser package also means two builds of the same tag ship different binaries with nothing in the repo recording which.

**The branch has not been built or run.** `isValidEmail` is called at `dashboard/src/pages/AdminTeamLeaders.tsx:381,1360` and is neither defined nor imported there, so the dashboard does not compile and that page cannot render. Fork runs are approval gated here so CI has not executed, but this one needs no CI to see.

**The contract is not updated.** The branch publishes more than twenty new routes and two role values and touches no `openapi.json`, nothing under `sdk/`, nothing under `docs/`, and no `CHANGELOG.md` entry. `openapi:check`, the four `check:sdk-*` gates and `check:contract-shapes` gate the Lint job on exactly that.

**The diff cannot be read as it stands.** Files that are not otherwise changed are reformatted at a much narrower width than the repo's `printWidth: 120`, which rewrites or deletes the comments that record why earlier fixes look the way they do. `src/common/storage/storage-local-files.ts` is +398/-59 with no behaviour change at all. The real changes are in there, but they cannot be separated from the rewrapping.

Separately, the branch is 358 commits behind `main` and conflicts. 0.23.4 and 0.23.5 shipped in that window and touched the same files, including the auth-directory keying above, several whatsapp-web.js lifecycle fixes, and session-scoped API keys in the dashboard. Resolving those conflicts in this branch's favour would revert them.

None of that means the work is lost. These are the pieces that would be welcome, each as its own pull request against current `main`, roughly in this order:

1. **Template picker in the chat composer.** There is no template access from the Chats page on `main` today, which is the gap #1484 was about, and this version is the better one: it sends the stored template through the send route instead of concatenating the parts into the text box. Keep the existing `<input>` rather than switching to a `<textarea>` (Enter is the only send gesture the composer has), give the `<select>` an accessible name, and add the new strings to all thirteen locale files, `he.json` included. Roughly eighty lines plus the locale entries.
2. **`POST /auth/api-keys/{id}/reissue`.** There is no key rotation on `main` at all: rotating a leaked key today means delete and recreate, which changes the key id and drops its scoping. Guard it with `@RequireRole(ApiKeyRole.ADMIN)` and give it its own audit action rather than reusing `API_KEY_UPDATED`. It also needs `npm run openapi:export`, a method in each of the five clients, a `docs/06` row and a changelog entry.
3. **Persisting a live chat-history read into the message store.** `persistHistoryMessages` already exists on `main` and is only wired to the history-sync event, so messages fetched by `getChatHistory` never reach the local store. About twenty-five lines, plus the spec you already wrote for it.
4. **The whatsapp-web.js post-QR readiness reconciliation**, on its own and without the browser swap: the reinjection grace window, the direct page-authentication probe, and a configurable readiness deadline. That closes a real failure `main` has not closed, where a QR scan succeeds on the phone while the library misses its handoff. It has to come out of the reformatting first, and the new environment variables need `.env.example` and `docs/12` entries.
5. **The Vite dev-proxy fix.** `dashboard/vite.config.ts:30` on `main` keys the proxy on the `/api` prefix, so the frontend route `/api-keys` is proxied to the backend and a hard refresh 404s. The regex form here fixes it. Ten lines.
6. **Windows process enumeration for the orphaned-browser sweep.** `src/engine/adapters/chromium-profile-hygiene.ts:32` returns early on anything that is not darwin or linux, so a Windows install leaks an orphaned browser after a hard kill. Keep main's best-effort contract there: log and move on, never throw, so it can never block an engine start.

Two notes on the parts I am not taking. The five-role model is a product decision rather than a patch: roles here are an ordered hierarchy (`docs/06-api-specification.md:42-46`), per-session confinement already exists as `allowedSessions` on any role, and tenant ownership stored on the always-SQLite `main` connection cannot be the authority for sessions that live on the data connection once more than one node is running. And the browser swap is its own change with its own risk, not a detail of a dashboard PR. If you want either of them, open a discussion describing the deployment you need it for before writing code.

#1484 is contained in this branch, so I am closing it alongside this one. Item 1 above is that work, and it is the one I would start with.

AbdoHussein1998:
I’m really grateful for your feedback and suggestions, and I genuinely wanted to address them properly.

At the moment, I may not be able to split these changes into several separate PRs, mainly because I developed the features iteratively and ended up committing multiple related changes together under commit names that were not clearly structured.

Since this was one of my first contributions to a public open-source project, I was still getting familiar with the team’s workflow and conventions. I’ll take this feedback into account going forward and make sure future changes are structured into smaller, more focused PRs with clearer commits.

I’ll also see if I can separate the current changes where it’s practical to do so.

And finally, I genuinely want to say **thank you from the bottom of my heart**. The core OpenWA project has been incredibly helpful in my work, and I’ve learned a lot from it. I truly appreciate the work you and the contributors have put into the project, and I’m grateful for the opportunity to contribute back to it.

PR #1631, opened 2026-09-17
feature(message): Added support for receiving button-type messages.

gabrielmmoraes1999:
## Description

Dashboard: render inbound prompt buttons and wire click-button
Inbound WhatsApp Business button/list prompts (Baileys) can now be shown and acted on in the dashboard chat UI.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] Tests added/updated
- [x] Documentation updated
- [x] Lint passes
- [x] Self-reviewed

image

## Related Issues

Closes #


rmyndharis (changes requested):
Thanks for this. The receive side is well built: I could not crash the mapper or the projector with malformed payloads (null array entries, non-object `paramsJson`, empty ids), the route's guards are right (session scope, OPERATOR, `requireStored` plus `assertStoredInChat`), and the dashboard escapes button text properly. Every gate passes locally once the branch is rebuilt on current `main`.

What blocks a merge is a disagreement between the choices the PR publishes and the choices it will accept back, plus three correctness defects. Details below, roughly in severity order.

## Blocking

**1. `buttons[]` advertises choices that the click route always refuses.**

`extractBaileysButtons` includes URL and call CTAs, using the URL or the phone number as the id (`src/engine/adapters/baileys-message-mapper.ts:414-423`), while `extractBaileysClickableButtons` deliberately drops them (`:526-539`). So a `templateMessage` whose only hydrated buttons are a `urlButton` and a `callButton` arrives on `message.received` as `buttons: [{ id: "https://...", text: "Pay now" }, { id: "+1555...", text: "Call us" }]`, is persisted into `metadata.buttons`, and is rendered as a tappable button by `ChatThread.tsx` (which has no clickability discriminator). Tapping it answers `400 not_a_prompt`, and because the outgoing row is saved before the engine call, every tap leaves a permanent failed message in the thread.

`ClickButtonDto.buttonId` already documents that CTAs cannot be clicked, which is the right call; the fix is to stop publishing them in the same array the docs point `buttonId` at. Simplest form: have the `templateMessage` arm of `extractBaileysButtons` drop `urlButton`/`callButton` and give its `interactiveMessage` arm the same `name` allowlist the clickable extractor uses, so the two functions agree. If the CTAs are wanted in the UI, they belong in a separate field carrying the URL explicitly, not in `buttons[]` where a URL masquerades as a button id and reaches every webhook consumer under that name.

**2. `selectedIndex` is the position in the filtered list, not the button's own index.**

`resolveBaileysButtonClick` uses `matchIndex`, the offset within the already filtered clickable choices (`baileys-message-mapper.ts:486-497`). A hydrated template of `[{index: 1, urlButton}, {index: 2, quickReplyButton "Sim"}]` therefore answers `selectedIndex: 0` for a button that is really index 2. `proto.IHydratedTemplateButton` carries its own `index`, but the local shape at `:346-358` does not declare it, so it cannot be read today. A bot branching on `selectedIndex` takes the wrong branch. Carrying `index` through the choice objects and using `entry.index ?? position` fixes it.

**3. The click reply ignores the chat's disappearing-messages timer.**

`clickButton` calls `generateWAMessageFromContent` without the ephemeral option (`baileys-messaging.ts:503-513`), while every other send in that class goes through `withEphemeral` (`:192`, `:704`). In a chat with a 7 day timer the prompt disappears on schedule and the answer stays forever. The new spec stubs `getEphemeralExpiration: () => undefined`, so nothing catches it.

**4. `metadata.buttons` has no bound.**

`collectButtons` filters nulls and nothing else (`baileys-message-mapper.ts:443-446`), and the projector stores the array whole (`message-row.mapper.ts:47-50`). A `listMessage` with a few thousand long rows is then persisted, dispatched to every webhook and WS client, and served back on every page of `GET /sessions/:id/messages`, which has a deliberate per-response budget for inline media precisely because unbounded per-row payloads break that response. A cap in `collectButtons` (WhatsApp's own ceiling is 3 buttons or 10 list rows, so something like 20 entries plus a length cap on `text`) covers all four arms at once.

## Please also address

**5. The native-flow reply shape is unverified, and the capability matrix overstates it.**

`buildBaileysButtonClickMessage` answers an `interactiveMessage` with `templateButtonReplyMessage` (`:575-584`). The comment justifies it with what quick-reply taps commonly *arrive* as, which does not establish what the server accepts when *sending*. Baileys has `proto.Message.InteractiveResponseMessage` with `nativeFlowResponseMessage`, and its outbound classifier has a dedicated `interactiveResponseMessage` to `native_flow_response` branch, so our reply currently goes out tagged as plain text. I could not settle from the source whether the server accepts the template shape here; it needs a live test against a real business native-flow prompt. Until it is tested, please do not advertise it as fully supported. Separately, the matrix entry at `engine-capability-matrix.ts:424-425` claims the adapter relays an `interactiveResponseMessage`, which the builder never constructs.

**6. Baileys does have button-reply helpers.**

The comment at `baileys-messaging.ts:480-481` says official Baileys has no `buttonReply` helper. 7.0.0-rc14 exposes `buttonReply` (with `type: 'template' | 'plain'`) and `listReply` in `Types/Message.d.ts`, implemented in `Utils/messages.js`, producing the same three shapes the PR hand-builds. Routing through `sendContent` would be a smaller diff and would inherit `withEphemeral`, the store put and the own-send echo for free. Only the `interactiveMessage` arm genuinely lacks a helper.

**7. `ChatHistoryMessageDto.button` / `.buttons` can never be populated.**

The description says Baileys only, and that route is whatsapp-web.js only. Both are true: `baileys.adapter.ts` answers `501` for `getChatHistory`, and the whatsapp-web.js adapter was not changed to fill either field. As it stands these are dead contract surface in the OpenAPI schema and in four SDKs. Please drop them and keep the fields on `message.received` and on persisted `metadata`, where they do work.

**8. Smaller items.**

- `ClickButtonDto.text` has no `@MaxLength`, unlike every other free-text field in that file; it reaches the wire as `selectedDisplayText` and the DB as `body`. When `text` is omitted, the persisted `body` is the raw `buttonId` while the wire carries the resolved label, so the row disagrees with what was sent.
- Duplicate ids are possible, since `id` falls back to the label when the wire id is absent. `ChatThread.tsx` uses `key={btn.id}`, and the click path resolves to the first match, so a list reusing a `rowId` across sections answers with the wrong row's title.
- `@ApiResponse({ status: 400 })` is declared twice on the route; the second silently wins in the generated schema.
- `mapBaileysMessageType` newly returns `text` for `listMessage`, `buttonsResponseMessage`, `templateButtonReplyMessage` and `listResponseMessage`, which previously fell through to `unknown`. That is a behavior change for any consumer filtering on `type`, so it belongs under `### Changed` rather than `### Added`.
- The CHANGELOG says the structured reply arrives on `message.received` and chat history; the REST chat history route answers `501` on Baileys, so only the first half holds. It also says persisted choices survive a reload, which is true of rendering but not of clicking: the engine store that backs the click has an eviction cap, so an old button renders and then 404s.

## Tests

Two of the new tests do not currently guard the behavior they name.

- `baileys-click-button.spec.ts:64-86` asserts `relayMessage` was called with the object its own `generateWAMessageFromContent` stub returns, so rewriting the `listMessage` or `templateMessage` arm to emit anything at all keeps it green. The same stubbing makes `assertStoredInChat` vacuous: deleting the cross-chat guard from `clickButton` breaks no test.
- `dashboard/src/utils/chatMessages.test.ts:75-82` covers `mapEngineHistoryMessage`, the whatsapp-web.js chat-history path, which no engine populates with buttons. The path that actually feeds the UI (the live WS payload and the DB `metadata`) has no dashboard test.

Untested branches worth covering with the fixes above: the `templateMessage` arm of both extractors, the `listMessage` and `templateMessage` arms of the reply builder, any `selectedIndex` other than 0, `assertStoredInChat` rejecting a click on a wrong-chat message, the ephemeral option on the click path, and `MessageSendService.clickButton` itself (no service or controller test exists, so the persisted row shape and the failure routing are unverified).

PR #1650, opened 2026-09-18
feat(dashboard): send a message to several groups from the message tester

C24212:
## Description

In Group mode the Message Tester can only target one group, picked from a dropdown. This replaces the dropdown with a group picker and sends the message to every selected group.

- A new `GroupPicker` component: search by name, a scrollable checkbox list, Select all (the groups currently shown) and Clear buttons, and a selected count.
- With more than one group selected, the message goes to each group in turn, 3 seconds apart, through the same single-send call the chosen message type already uses (text, media, location, contact, sticker, poll). A failing group does not stop the rest. The response panel shows how many groups were sent to and lists each failed group by name with its error. The send counts as failed only when every group failed. While sending, the button shows progress ("Sending 2/5...").
- With one group selected, nothing changes: same request, same response panel with the message ID.
- Forward keeps a single source chat, the first selected group, as with the previous single selection.
- The selection resets when the session or the recipient type changes, so a group from another session can't be sent to by accident. The old auto-selection of the first group is gone: with multiple selection, silently including a group the user did not pick is worse than starting empty.
- The personal number input is now bound to its label (`id="mt-13"`); before, the label only pointed at the group select.

Implementation notes:

- The per-type send `switch` moved into a `sendToChat(chatId)` helper inside `handleSend`, so the single and multi-group paths share it. The diff is easier to review with whitespace hidden.
- The pure logic lives in `utils/groupSelection.ts` and `utils/sendSequentially.ts`, with unit tests.
- New translation keys in all 13 locales.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] Tests added/updated (`groupSelection.test.ts`, `sendSequentially.test.ts`)
- [x] Documentation updated (`CHANGELOG.md` `[Unreleased]`)
- [x] Lint passes (`lint`, `format:check`, `typecheck`, `i18n:check`, `build`, `test:unit`)
- [x] Self-reviewed

## How it was tested

Dashboard dev server against a local mock of the API, where one group always fails:

- Search, Select all on a filtered list, Clear, and unticking single groups update the list and the count as expected.
- Sending to four groups, one of them failing: the requests went out in order, 3.0 s apart; the panel showed 3/4 sent and named the failing group with its error.
- Sending to one group returned the message ID exactly as before.
- Switching to Personal and back cleared the selection.

## Note on CI

While testing locally I found that the `App.test.ts` case "an admin keeps the Logs entry and the route" (added in b57fcf95) is sensitive to test-file scheduling: its 1 s `waitFor` on the lazy `/logs` route can time out when the runner is busy. On my machine, adding any test file under `src/utils/`, even a one-line placeholder on an unchanged `main`, made it fail consistently. This branch adds two test files there, and in three full runs on my machine it hit this case (or the similar lazy-route wait in `authLifecycle.test.ts`) twice and passed once. If it shows up in CI here, a re-run should clear it. I can send a separate PR that gives that wait more headroom if you want.

## Note

This is one of four small dashboard PRs opened together (template row delete, Escape in Chats, multi-group send, bulk attachments). Each adds a line under `### Added` in `CHANGELOG.md` `[Unreleased]`, and the two Message Tester ones also touch `MessageTester.tsx`, so they will conflict with each other as they land. I will rebase each remaining PR after the previous one is merged.


rmyndharis (changes requested):
Thanks for this, and for splitting the two pure helpers out with their own tests: `groupSelection` and `sendSequentially` are clean, and every dashboard gate passes on the branch with current `main` merged in.

What holds it back is that this is a client-driven bulk sender on a surface that talks to real WhatsApp groups, and it has none of the guards the server-side bulk path ten lines above it already has: it keeps sending after the gateway refuses, it cannot be stopped, and it has no recipient cap.

## Blocking

**1. The loop keeps firing at a gateway that already answered 429.**

`sendSequentially.ts:29-31` collapses every rejection to `err.message`. `api.ts` puts the two fields that matter on the Error itself (`err.status`, `err.code`), and both are thrown away, so nothing can tell a pacing refusal from a "group not found". With `SEND_PACING_ENABLED=true`, a session that hits its daily warm-up cap answers `429` with a `retryAfterSeconds` measured in hours; this loop records that as one more per-group failure and sends the remaining 47 requests, three seconds apart, each certain to be refused and each counted as a refusal server-side. The same applies to the global throttler's `429` and the inflight-body-budget's `503`.

Smallest fix: `stopOn?: (err: unknown) => boolean` on the options, checked in the catch, with the call site passing `err => [429, 503].includes((err as { status?: number }).status ?? 0)`. Report the untried groups as not attempted rather than dropping them from the total.

**2. A running send cannot be stopped, and leaving the page does not stop it either.**

There is no cancel control, no `AbortController` and no cancellation flag anywhere in the diff. `handleSend`'s async closure is not tied to the component, so navigating away mid-run keeps delivering: select 200 groups, spot a typo, go to Sessions, and the remaining 197 messages still go out, with the final `setResponse` landing on an unmounted component and vanishing. Nothing in the UI shows it is still running.

The bulk path next to it has `cancelBatch`, a `batchCancelling` state and a server-side batch that survives a reload. Minimum here: a ref checked between iterations, set on unmount, plus a Cancel button while `groupSendProgress` is non-null.

**3. No recipient cap, and the media payload is rebuilt for every group.**

`selectedGroups` is unbounded, while the sibling path caps at both ends (`BULK_MAX_RECIPIENTS = 100` in `bulkRecipients.ts`, `@ArrayMaxSize(100)` on the DTO). `sendToChat` rebuilds the payload per iteration, so an 18 MB local file across 200 groups uploads roughly 4.8 GB from the browser, each request contending for the inflight body budget that answers `503` when exhausted, which item 1 then ignores. Reuse `BULK_MAX_RECIPIENTS` and disable Select all at the cap.

**4. `groupLabel` throws on a group with no name, which blanks the page.**

`groupSelection.ts:6-8` is `group.name.trim() || group.id`. The wire type declares `name: string`, but nothing enforces it: the whatsapp-web.js adapter passes `g.name` and the Baileys adapter passes `metadata.subject` straight through, and `group.service.spec.ts` mocks a group as `{ id: 'g1' }` with no name at all. The code this replaces rendered an empty `<option>`; now it is an uncaught `TypeError` during render, so one nameless group takes down the whole Message Tester. `group.name?.trim() || group.id`, and widen the type to `string | undefined` so it stops claiming otherwise.

## Please also address

**5. Forward silently ignores every group but the first.** The picker stays live in Forward mode, the counter reads "Selected: 5", Send is enabled, and only `selectedGroups[0]` is used, as the source chat. Only the CHANGELOG says so, which is not where the user is looking. Either say it in the existing hint slot or make the picker single-select in that mode.

**6. A green "Success" header when almost everything failed.** `success: sent > 0` means 1 of 50 delivered still renders the green check above 49 red rows. `failures.length === 0` is the honest version; the per-group summary already carries the partial case.

**7. A selected group that leaves the list stays selected and is still sent to.** Nothing reconciles `selectedGroups` against `groups`, and the list does refetch under the user (`staleTime: 60_000` with `refetchOnWindowFocus`). The counter then disagrees with the visible rows, and the send addresses an id the picker cannot show, reported by raw JID. `SessionScopePicker` solves exactly this by keeping an unresolvable selected id visible so it can be unticked; mirroring that is the smallest fix.

**8. The picker stays editable while a send runs.** `GroupPicker` has no `disabled` prop, so the selection can be cleared mid-send while the captured array keeps going, leaving the counter at 0 and the result panel reporting groups the UI shows as unselected. `SessionScopePicker` threads `disabled` to every control.

**9. "Select all" adds only the filtered subset, in every locale.** The behaviour is right; the label is not. Typing a query, clicking it, then clearing the query leaves a long list with a few ticks and no way to reconcile the count. Rename the key or disable it while a query is active.

**10. Accessibility, against the sibling picker.** `SessionScopePicker` uses a real `ul`/`li` so a screen reader announces the list and the position, which matters more at 200 rows than at 12; this one is a bare `div role="group"` with labels. The send progress is rendered only in the disabled button's text, so it is announced nowhere, while the selection counter right above it already uses `role="status"` correctly.

**11. CHANGELOG.** Four sentences where the section runs one or two, no link, and the credit convention for outside contributions is missing. Something like: "The dashboard Message Tester sends to several groups at once: the Group dropdown is now a searchable checkbox list, and each selected group is messaged in turn ([#1650](https://github.com/rmyndharis/OpenWA/pull/1650)). Thanks @C24212."

## The test that is missing

The pacing itself is untested. All six `sendSequentially` tests inject `sleep`, while the call site injects nothing and runs against the default `wait`. Replace that default with `() => Promise.resolve()` and every test still passes, while the feature becomes an unpaced burst at WhatsApp, which is the ban risk the three-second gap exists to prevent. One test that omits `sleep` and asserts elapsed time for two targets at `delayMs: 20` closes it.

Also uncovered: the routing at the call site (the `> 1` branch, the forward bypass, `success: sent > 0`, the name fallback), and `groupLabel` with a missing name, which is item 4. Extracting the routing decision into `groupSelection.ts` would make it testable under the existing harness.

C24212:
Thanks for the detailed review. Reworked it and merged current main in.

Blocking:
1. `sendSequentially` keeps the `status` on each failure and takes a `stopOn` predicate. The page passes `isGatewayRefusal` (429/503), so the run stops at the first refusal and the remaining groups are reported as not sent ("2, stopped after HTTP 429") instead of dropping out of the total.
2. A running send can be cancelled: a Cancel button sits next to the progress while it runs, and leaving the page aborts it through the same `AbortController`. The wait between groups listens to the signal, so a cancel doesn't sit out the 3 seconds.
3. The selection is capped at `BULK_MAX_RECIPIENTS`. Select all fills up to the cap, unticked boxes are disabled once it's reached, and the counter shows `n/100`.
4. `groupLabel` is `group.name?.trim() || group.id`, and `name` is optional on the `getGroups` response type now.

Also:
5. In Forward mode the existing hint slot says only the first selected group is used, as the chat to forward from.
6. Success now means every group was sent to; a partial run shows its summary under a red header.
7. Selected ids that drop out of the list stay visible by their id so they can be unticked, same approach as `SessionScopePicker`.
8. The picker takes `disabled` and is locked during a run, along with the session select and the recipient-type toggle.
9. While a search is active the button reads "Select matches".
10. The list is a `ul`/`li`, and progress is announced through a `role="status"` line instead of the disabled button's text.
11. CHANGELOG entry replaced with your wording.

Tests: a real-timer pacing test with no injected `sleep`, plus abort (before the next target and during the wait) and `stopOn` cases for `sendSequentially`. The routing moved into `planGroupSend` in `groupSelection.ts`, tested for the single, forward and sequential paths, along with `groupLabel` with a missing name and the cap. Page tests cover a nameless group, Cancel, and a 429 stopping the run.

On merge order with #1652: `handleSend` is restructured so the diff no longer moves the bulk block. I checked that this branch and #1652 merge cleanly in either order, and that the merged result passes lint, typecheck, build and the tests, including #1652's request-body test.

C24212:
Thanks for merging main in and rerunning it!

rmyndharis:
I pushed two commits to this branch: current `main` merged in (e1ed076a), and f2c6570b:

- The composer is wrapped in a fieldset that is disabled while a multi-group run is in progress. The run sends what was on screen when it started, so an edit made during a run of several minutes looked applied and was ignored. Cancel sits outside the fieldset and stays usable. Ignoring whitespace the change is four lines; the rest is the re-indent.
- The cancel test now checks that the composer is locked during the run and that one of two groups sent is reported as failed, so reverting the success rule to "any group sent" turns it red.

rmyndharis:
#1652 is merged, and I merged `main` into this branch (f0990929). The composer fieldset from my previous commit re-indents the block #1652 edits, so `MessageTester.tsx` conflicted; the resolution is your version merged with #1652's changes, with the fieldset wrapped around the result. Both PRs' tests pass together on it.

PR #1652, opened 2026-09-18
feat(dashboard): attach a file or media URL to a bulk send in the message tester

C24212:
## Description

`POST /sessions/:id/messages/send-bulk` already accepts image, video, audio and document items, but the dashboard's Bulk mode only sends text. This adds an optional attachment to Bulk mode, with no backend change.

- Bulk mode shows the same Media URL and Upload File fields as the single media types, below the message.
- A picked file is sent inline (base64), and its kind is detected from its MIME type: image, video or audio, and anything else is sent as a document. For a URL, the kind is guessed from the file extension. A "Send as" toggle lets the user change it, for example to send a photo as a document.
- The message text becomes the caption, and it is optional when there is an attachment.
- Documents carry a filename: the picked file's name, or the last segment of the URL. Without it, Baileys names URL documents `file`.
- Inline media is repeated in every item, so the request grows with the recipient count. The form refuses a picked file when `base64 length × recipients` would exceed 24 MiB, the same margin under the default 25 MiB `BODY_SIZE_LIMIT` as the existing 18 MiB single-send check, and points to the URL option instead. With a URL, the gateway fetches the file for each recipient and the request stays small.
- Switching into or out of Bulk mode clears the media URL, so a URL typed for a single image send never rides along on a bulk text send unnoticed.

The pure logic lives in `utils/bulkMedia.ts`, with unit tests. Three new translation keys, in all 13 locales.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] Tests added/updated (`bulkMedia.test.ts`)
- [x] Documentation updated (`CHANGELOG.md` `[Unreleased]`)
- [x] Lint passes (`lint`, `format:check`, `typecheck`, `i18n:check`, `build`, `test:unit`)
- [x] Self-reviewed

## How it was tested

Dashboard dev server against a local mock of the API, checking the `send-bulk` body it received:

- A PNG to 2 recipients: `type: "image"`, `content.image` with `base64` and `mimetype`, and the message as `content.caption`.
- The same PNG with "Send as" set to Document: `type: "document"`, with `filename`.
- A 1 MB PDF to 100 recipients: the size warning appears and Send is disabled.
- A `.mp4` URL to 100 recipients: the kind switched to Video, and the request carried only the URL (about 15 KB for all 100 items).

## Note on CI

While testing locally I found that the `App.test.ts` case "an admin keeps the Logs entry and the route" (added in b57fcf95) is sensitive to test-file scheduling: its 1 s `waitFor` on the lazy `/logs` route can time out when the runner is busy. On my machine, adding any test file under `src/utils/`, even a one-line placeholder on an unchanged `main`, made it fail consistently, and this branch passes the full suite with its new test file removed. If it shows up in CI here, a re-run should clear it. I can send a separate PR that gives that wait more headroom if you want.

## Note

This is one of four small dashboard PRs opened together (template row delete, Escape in Chats, multi-group send, bulk attachments). Each adds a line under `### Added` in `CHANGELOG.md` `[Unreleased]`, and the two Message Tester ones also touch `MessageTester.tsx`, so they will conflict with each other as they land. I will rebase each remaining PR after the previous one is merged.


rmyndharis (changes requested):
Thanks for this. The shape is right: the file is read once at pick time rather than per recipient, the inline cap is anchored to the body limit, the mutual exclusion between the picker and the URL field is handled in both directions including across a mode switch, and the media-URL path lands on a route that is already SSRF-guarded, byte-capped and proxy-attributed, so it opens nothing new. Every dashboard gate passes on the branch with current `main` merged in, and the ten `bulkMedia` tests are well chosen for what they cover.

Two things need fixing before it goes in.

## Blocking

**1. A media URL with no scheme is delivered as garbage bytes to every recipient, and reported as sent.**

`bulkMedia.ts:57-70` accepts any non-blank string as a URL, and `MessageTester.tsx:307` lets a bare attachment make the form valid on its own. Server-side this is the one media path with no URL validation: `BulkMediaDto.url` is documented as "Media URL (http/https)" but carries only `@IsOptional()` and `@IsString()`, where the single-send DTO has `@IsUrl()`. Both engines then treat a non-`http(s)` string as base64 (`baileys-messaging.ts` tests `/^https?:\/\//i` before falling back to `Buffer.from(media.data, 'base64')`, and the whatsapp-web.js adapter passes the string straight to `MessageMedia`).

So `cdn.example.com/pricelist.pdf`, which the placeholder invites but the field does not reject, sends every recipient a `pricelist.pdf` of about twelve bytes of noise while the batch reports `failed: 0`. `file://` and `data:` behave the same way.

Smallest fix on this PR: reject it client-side, mirroring `sessionForm.ts`'s proxy-URL check, so `toBulkAttachment` returns `null` and Send stays disabled for anything that is not `http(s)`. Adding `@IsUrl()` to `BulkMediaDto.url` is the root-cause fix, but it changes behaviour on a public route for direct API callers too, so it belongs in its own PR rather than riding along here.

**2. Merge order against #1650, which moves the exact block this PR edits.**

This branch is based on plain `main` and does not assume #1650. But #1650's diff deletes the bulk branch of `handleSend` from where it is now and re-adds it further down, and this PR's only wiring line (`buildBulkMessages(...)` in that block) lives inside the deleted range. That is a delete-versus-modify conflict, and the natural resolution, taking #1650's relocated block, silently drops this feature's wiring: it still compiles, typechecks, lints and passes all unit tests, because nothing asserts the request body. The picker, the kind toggle, the cap and the hint all still render, and the send goes out as text.

Merging this one first and rebasing #1650 onto it avoids that, and the test in the last section makes the regression impossible to miss in either order.

## Please also address

**3. Attaching a file silently re-scopes the message box from 4096 to 1024 characters.** Once an attachment exists the text goes out as `caption`, which the DTO caps at 1024 while `text` is capped at 4096. The textarea has no `maxLength` and `formValid` does not check length, so a 2000-character broadcast that sends fine as text fails the whole batch with a validation error after a logo is attached, with no warning beforehand. A length check in the bulk arm of `formValid`, surfaced in the existing hint, covers it.

**4. The effective per-file ceiling is not stated anywhere.** The cap is `base64.length * recipients`, so the per-file budget collapses as the list grows: about 18 MiB at one recipient, 369 KiB at fifty, 184 KiB at the DTO's hundred. Neither hint names a number, so it is discovered by trial. Worth putting the effective budget in the too-large string, and a comment on `BULK_INLINE_MEDIA_MAX_BYTES` explaining that 24 MiB is the 25 mb body limit minus envelope, and that an operator who lowers `BODY_SIZE_LIMIT` is outside what the client can know.

**5. Smaller items.** Editing the URL re-derives the kind on every keystroke, so a deliberate "send as Document" choice snaps back to image when the URL is edited afterwards; guard it or only re-derive on blur. The too-large hint gains `role="alert"` at the same moment its text appears, which screen readers commonly miss; keeping `role="status"` on the span and only swapping the class is the usual fix. A staged attachment survives a send, matching the single-send path, but it now carries a payload, so a stray double-click re-uploads it; worth a thought, not a blocker.

**6. CHANGELOG.** Four sentences where the section runs one or two, and the credit convention for outside contributions is missing. Also "past the default 25 MB body limit" describes the cap loosely, since the refusal happens at 24 MiB of base64. Something like: "The dashboard Message Tester's Bulk mode can attach a file or a media URL, sent as image, video, audio or document with the message text as the caption. An inline file too large to repeat for every recipient is refused before sending. Thanks @C24212."

## Tests

The ten helper tests are good, but nothing covers the parts that can actually break. `MessageTester.test.ts` already has the harness for all of it: jsdom, a fetch stub, providers and a file-pick helper.

Two tests would close the gaps that matter: one asserting that a bulk send with a picked file POSTs `content.image` for every recipient, which is what makes item 2 dangerous, and one asserting that an oversized inline attachment disables Send, which is the feature's only safety mechanism and is currently unasserted. The `handleFileChange` bulk branch and the `reader.onerror` path are untested too.

C24212:
Thanks for the thorough review. Pushed the changes and merged current main in.

1. Only http(s) counts as a media URL now. `toBulkAttachment` returns null for anything else, and while the URL field holds something that isn't http(s), Send stays disabled with a hint under the field. I left `@IsUrl()` on `BulkMediaDto.url` for a separate PR, as you suggested.
2. Fine with this one going first. The new page test sends a bulk with a picked image and checks `content.image` on every item, so a rebase that loses the wiring fails it.
3. With an attachment the message is held to the 1024-character caption limit: Send is disabled and a hint under the message says why.
4. The too-large hint now states the per-file budget for the current recipient count (184 KB at 100, for example), and `BULK_INLINE_MEDIA_MAX_BYTES` has the comment. I moved it into `utils/bulkMedia.ts` so the tests can use it.
5. The kind follows the URL only until you pick one yourself. The hints keep `role="status"` and only swap the class. Send is already disabled while a send is in flight, so a double-click can't send twice; sending again after it finishes behaves like single sends.
6. CHANGELOG entry replaced with your wording.

Tests: page tests for the request body, the oversized-file guard, the non-http URL, the caption limit and a failed file read, plus unit tests for the URL check and the per-file budget.

C24212:
The Build failure is a network error during `npm ci`: better-sqlite3 fell back to `node-gyp rebuild`, and fetching the Node headers from nodejs.org failed with `ECONNRESET`. Nothing from this PR is involved (it doesn't touch any package files, and the same job passes on #1650), so a re-run should clear it.

rmyndharis:
I merged current `main` into this branch (89cdcf38). The only conflict was the CHANGELOG, where both entries are kept. The push also reruns CI, including the Build job that failed on the network error.

rmyndharis:
I pushed two commits to this branch: current `main` merged in (4ccc902d), and a431c150, which closes two gaps left in the send path:

- A media URL with a scheme typo, such as `https:/cdn.example.com/a.pdf`, passed `isHttpMediaUrl` because the URL parser repairs it, but the server treats only a string starting with `http://` or `https://` as a URL and decodes the rest as base64, so recipients got a file of garbage bytes and the batch reported success. The raw string now has to carry the prefix.
- Audio goes out without a caption, so text next to an audio attachment was dropped while the batch reported it sent. That combination is now refused with a hint (one new string in all 13 locales), and the caption limit applies only to the kinds that carry one. The caption is counted the way the server's `@MaxLength` counts it, so an emoji no longer counts twice.

The URL guard test now types message text first, so only the guard it covers can hold Send, and there is a page test for the audio case. The CHANGELOG line says audio carries no caption.

Issue #1634, opened 2026-09-17
[Feature]: Group isolation — per-chat scoping for webhooks and the read surface


bhavyachopra99:
### Pre-flight

- [x] I searched existing issues and this isn't already requested

### Problem / motivation

OpenWA links a full WhatsApp account. Any API key or dashboard session that can reach a session can read every chat on it — DMs included. For a personal number there is no way to let a downstream agent/tool see *only* a specific set of groups.

Two concrete gaps today:

1. **Webhook filters have no chat-id field.** In `src/modules/webhook/filters/filter-types.ts`, `FILTER_FIELDS.message` exposes `sender`, `recipient`, `body`, `type`, `isGroup`, `kind`, `fromMe`, `hasMedia`, `mentions`. In a group, `sender` resolves to the participant (`author`), so `kind: "group"` selects *all* groups and cannot be narrowed to chosen JIDs. There is no `chatId`.
2. **Filters only gate webhook dispatch.** The read surface is not scoped either: `GET /sessions/{id}/chats`, `/contacts`, and `/messages?chatId=` return everything, and `chatId` on `GET /sessions/{id}/messages` is **optional**, so a bare call lists messages across all chats.

Use case: link a personal number, let an agent read only a curated set of groups, and keep personal DMs and all other groups off the wire. Today the only option is an external filtering proxy in front of the API.

### Proposed solution

1. **Add a `chatId` field to the message filter registry** (kind `id`, operators `is`/`isNot`), resolving `data.chatId ?? data.from`. This looks like the documented extension point — one entry in `FILTER_FIELDS.message`, plus the dashboard `MESSAGE_FIELDS` list and one i18n key. Happy to open a focused PR for this part.
2. **Longer term: a group-level scope** alongside the existing session-scoped API keys (`allowedSessions`), enforced on the read surface (REST + MCP + dashboard). The repo already has route/query fence coverage machinery (`global-route-fence-coverage.spec.ts`, `query-session-scope-coverage.spec.ts`) that this could extend. Note the `chatId`-optional path on `GET /messages` would need handling (require `chatId` or filter the result set), and list endpoints (`/contacts`, `/groups`) can't 403 — they would need filtering rather than rejection.

### Alternatives considered

- **External relay**: receive the webhook and re-serve only allowlisted groups. Works, but every consumer must be prevented from reaching the OpenWA API directly, and it duplicates a scoping concept the server already implements for sessions.
[...]

PR #1674, opened 2026-09-20
feat(auth): allowedChats scope for API keys (PR 1: entity, guard fence, coverage spec)

bhavyachopra99:
Implements **PR 1 of the chat allowlist** accepted in #1634: the entity column, migration, guard fence, and the structural coverage spec — so the remaining REST read/send surfaces and the follow-up slices (events gateway, agent tools, dashboard form, SDK types) can be added against a fence that cannot be bypassed silently.

## What's here

**Storage & contract**
- `allowedChats: string[] | null` on `ApiKey` (`simple-array`, nullable), independent of `allowedSessions`, NULL/empty = unrestricted (same fail-open semantic).
- Main-connection migration `AddApiKeyAllowedChats` (`ALTER TABLE api_keys`, idempotent, with a regression spec).
- `CreateApiKeyDto` / `UpdateApiKeyDto` / `ApiKeyResponseDto` field + `IsChatIdConstraint` (accepts a group `<id>@g.us`, a contact `<phone>@c.us` / `<lid>@lid`, or a bare phone; rejects the `simple-array` footguns — empty, comma, surrounding whitespace).
- `allowedChats` wired into `apiKeyAuthorizationFingerprint` so a scope change evicts live WebSocket sockets like the session scope does.

**Enforcement**
- `ApiKeyGuard` rejects an out-of-scope chat with `403`, the way `allowedSessions` does:
  - **path** — `:chatId` / `:groupId` / `:contactId`;
  - **body** — `body.chatId`, which every send DTO names, so the whole send surface is fenced centrally instead of one handler at a time (and a handler cannot forget it).
- `ChatScopeService` (global AuthModule) compiles the allowlist through the shared lid mapping table: a contact allowlisted by phone also matches its resolved `@lid` form and vice versa; an **unmapped `@lid` stays outside the fence**.

**Spec**
- `chat-scope-coverage.spec.ts` — in the style of `query-session-scope-coverage.spec.ts`. It scans every controller and fails when a handler takes a chat id in the query or body without a fence or a recorded reason, and asserts the guard still references each path param plus the body rule so the guard cannot be narrowed silently.

## Deliberate scope / remaining work

The coverage spec carries an explicit `ALLOWLIST` (with reasons) for the three shapes not yet fenced, and is expected to shrink to empty:

- `GET /messages` without `chatId` — require `chatId` or filter the page;
- `sendBulk` — each `items[].chatId`;
- `forward` — the target travels as `to`.

List filtering for `/chats`, `/contacts`, `/groups`, `/search` is likewise still owed; `ChatScopeService.filter()` is the helper for it. If you'd rather this PR carry all of the REST read/send fences before it lands, say so and I'll fold them in — I scoped them out to keep the first PR reviewable and the mechanism verifiable on its own.

**Open question:** `@RequireUnscopedKey` currently only tests `allowedSessions`, so a key scoped *only* by `allowedChats` still passes it and can manage API keys. Extending it touches the last-admin invariant (`usableAdminCondition`), so I left it out of this slice rather than change the security model unilaterally — happy to take it as a follow-up or fold it in.

## Verification

`npm run build`, `npx tsc --noEmit`, `npm test` (6550 passing), `npm run test:docs`, `npm run lint`, `npm run format:check`, `npm run openapi:check`, `npm run check:contract-shapes`, `check:sdk-routes|coverage|docs`, and the dashboard `lint`/`format:check`/`typecheck` all pass. `openapi.json` regenerated; the dashboard `ApiKey` type gained the field so the contract-shapes gate stays green.

Refs #1634.


rmyndharis (changes requested):
Thank you for this. The shape is right and much of it is done carefully: the fence sits in `ApiKeyGuard` where a handler cannot forget it, the migration is idempotent, the validator rejects the `simple-array` footguns, and `allowedChats` is wired into both the authorization fingerprint and the pre-update snapshot, so socket eviction works.

On your two questions. Yes, the send fences belong in this PR, and so do the query ones. List filtering can come later under the amended default in #1634, because an unfiltered list route then refuses the key instead of leaking. And yes, a chat-scoped admin must not manage keys: under that default the key-management routes refuse it without a special case, which leaves the last-admin invariant. `usableAdminCondition` (`auth.service.ts:331-337`) and the update pre-check (`auth.service.ts:220-223`) must stop counting a key with `allowedChats` as a usable admin.

What has to change before this can merge:

1. CI fails. `npm test -- --coverage` (`ci.yml:160`) passes every test and then exits on `Coverage for functions (82.35%) does not meet "./src/modules/auth/" threshold (85%)`. No test sets a non-null `allowedChats` on the guard, so both 403 branches can be disabled with the suite still green. #1677 by @lasithadilshan adds ten guard tests on top of this branch; please cherry-pick 13baf77b (`git cherry-pick` keeps them as the commit author). Measured on that branch merged with current `main`, `src/modules/auth` reaches 86.15% and the gate passes. Its fence also uses the real field names, `fromChatId`/`toChatId` and `messages[].chatId`; the comments here name `to` and `items[]`, which do not exist.
2. A `@lid` entry admits the `@c.us` chat with the same digits. `buildChatScope` puts every contact user-part into one set (`chat-scope.ts:80`) and `chatScopeAllows` admits it under `c.us`, `s.whatsapp.net` and `lid` alike (`chat-scope.ts:104`), so `buildChatScope(['555000111@lid'])` admits `555000111@c.us`. Lid digits are not a phone number: keep the two in separate sets and cross them only through the mapping table. `MessageService.resolveJidCandidates` (`message.service.ts:355-367`) encodes exactly these rules and can be lifted into a shared helper. The lookup also has to be deterministic: `lidsForPhone` reads only the in-memory reverse map, which is evicted at `LID_MAPPING_CACHE_MAX` (`lid-mapping-store.service.ts:146-147`, `:244`), so the same key and chat can pass or 403 depending on cache residency.
3. A bare number is accepted in the allowlist but never matches: `chatScopeAllows` has no branch for a chat id without `@` and returns `false`. Normalize it to `<digits>@c.us` when the key is saved, or drop it from the grammar.
4. The guard never reads `request.query`, so `GET /search?chatId=` (`search.controller.ts:43`, `search-query.dto.ts:32`) and `GET /sessions/:sessionId/messages?chatId=` (`message.controller.ts:111`) read any chat. Also reject a chat field that is present but not a string rather than skipping it, since the global pipe converts it afterwards (`app-validation.ts:20`).
5. Invert the default as described in #1634: admit a key with `allowedChats` only on handlers marked safe for it, and answer 403 everywhere else. Webhooks, automation rules, status, group creation and channels are then closed without enumerating their fields, and so is every route added later. `/events` and the MCP mount authenticate outside the guard and should refuse such a key at their handshake until the second PR filters them.
6. `allowedChats` is never returned. The five response objects in `auth.controller.ts` are built field by field (`:60`, `:86`, `:111`, `:154`, `:200`) and none includes it, although `openapi.json` and the dashboard type now advertise it. `authzSnapshot` (`:133-136`) needs it too, or an update that widens a key's chat scope is audited with identical before and after.
7. The coverage spec should prove what the guard relies on. With the default inverted, it has to show that every marked handler is fenced by path, body or query, or filters, and it needs the non-vacuity check its model has (`query-session-scope-coverage.spec.ts:99`). As it stands, two of its three ALLOWLIST entries can be deleted without the spec failing, and the `to|recipient` classifier matches no DTO in the repository.
8. Ship `filterByChatScope` and `ChatScopeService.filter` with their first caller; without one they are dead code and part of the coverage shortfall. The CHANGELOG entry should describe what this PR enforces.

Run locally on this branch merged with current `main`: tsc, eslint, `format:check`, `test:docs`, the contract gates, `openapi:check`, the full unit suite and the dashboard lane pass; `npm test -- --coverage` fails on the threshold above.

bhavyachopra99:
All review points are addressed, and the branch is merged with current `main`. `npm test -- --coverage` now passes (the `./src/modules/auth/` threshold is met), along with `tsc`, `eslint`, `format:check`, `test:docs`, the contract gates, `openapi:check`, the dashboard lane.

**CI / coverage** — cherry-picked `13baf77` from #1677 (@lasithadilshan keeps authorship) for the forward/bulk fences and the ten guard tests, then extended them for the new default-deny, query and non-string branches.

**Default inverted** — a key with `allowedChats` is now admitted only on handlers marked `@ChatScoped` and refused `403` everywhere else, so webhooks, automation rules, status, key management, channels and any route added later are closed without enumerating fields. Marked handlers are the chat routes whose target the guard can see (path, query, or body), plus `GET .../chats` which filters through `ChatScopeService.filter` — that is the helper's first caller. A chat-scoped admin no longer counts as a usable admin (`usableAdminCondition` and the update pre-check), so key management closes without a special case.

**Query + non-string** — the guard now fences `?chatId=` (`GET /messages`, `GET /search`) as well as the path and body, and rejects a chat field that is present but not a string rather than skipping it, since the global pipe coerces with `enableImplicitConversion`.

**Identity** — phones and lids are separate address sets, crossed only through the mapping table, so `555000111@lid` no longer admits `555000111@c.us`. The rules are lifted into `engine/identity/jid-candidates`, shared with `MessageService`, and the authorization path reads the persisted table (`resolveLidPersisted` / `lidsForPhonePersisted`) instead of the LRU-evictable cache, so the same key and chat cannot pass or fail on cache residency. A bare number is normalized to `<digits>@c.us` on save.

**Surface** — `allowedChats` is returned by all five API-key responses and added to `authzSnapshot`. `/events` and the MCP/agent-tool invoker refuse a chat-restricted key at their handshake until the second PR filters them.

**Spec** — rewritten around the marked-handler contract: every `@ChatScoped` handler must be fenceable by the guard or filter through `ChatScopeService`, with a non-vacuity check (>30 marked handlers, DTO scan non-empty). The `to|recipient` classifier that matched no DTO is gone, and there is no ALLOWLIST left to delete entries from.

The CHANGELOG entry now describes what this PR enforces. Ready for another look — happy to merge on your go-ahead.

bhavyachopra99:
Checked the amendment against the branch. Every hazard it names is already closed, and I've now encoded the one category the coverage spec was missing.

**Closed by the default-deny (verified by an audit of the branch):**

| hazard | route | result |
|---|---|---|
| webhook `events: ['*']` | `webhook.controller` / `webhooks-list.controller` | unmarked → 403 |
| automation rule with no conditions | `automation-rule.controller` | unmarked → 403 |
| status `recipients[]` | `status.controller` — `sendText/Image/Video/VoiceStatus` | unmarked → 403 |
| group creation `participants[]` | `group.controller :: create` | unmarked → 403 |
| channels as `:channelId` | `channel.controller` — `findOne/getMessages/remove/mute/demoteAdmin/transferOwnership/unsubscribe` | unmarked → 403 |

**Consequences:** a list route that does not filter yet stays unmarked and refuses (`GET /messages`, `/contacts`, `/search`, group list); the only marked list is `GET .../chats`, which filters through `ChatScopeService.filter`. `/events` and the MCP/agent-tool invoker refuse a key with `allowedChats` at their own handshake. Key management stays unmarked so it needs no special case, and `usableAdminCondition` + the update pre-check stop counting a chat-scoped key as the surviving admin.

**What I just fixed:** the coverage spec only accepted the first two marking categories (fenced / filtered), so a *chat-agnostic* mark — the third category in the amendment — would have failed it. It now accepts a chat-agnostic mark and instead flags the actual hazard: a marked handler that reaches chats through a field the guard does not read (`recipients[]`, `participants[]`, `channelId`, …) while naming no guard-visible chat. A group route carrying `participants[]` is still cleared when its path names the `:groupId`, since the guard fences that. The three categories are now documented on `@ChatScoped`.

No route is marked chat-agnostic yet — that is a deliberate access grant, and the marker is the record of it, so I left it to you rather than guessing which non-chat routes a chat-scoped key should keep. If you want a set granted (session status/read, profile, label CRUD, catalog reads, template CRUD, …), name them and I'll mark them.

All gates pass: `npm test -- --coverage` (6654), `test:docs`, `tsc`, `eslint`, `format:check`, `openapi:check`, `check:contract-shapes`, the three SDK gates, dashboard lane.

rmyndharis (changes requested):
Thank you, this is a substantial round and most of it lands exactly as asked. Checked on the branch merged with current `main`: the cherry-pick keeps @lasithadilshan as the author of fe169482, and the default-deny, the query and non-string checks, the lid/phone separation on persisted lookups, the bare-number normalization, `usableAdminCondition` and the update pre-check, the five responses and `authzSnapshot`, and the `/events` and tool-invoker refusals all hold. `npm test -- --coverage` passes with the `src/modules/auth` floor met.

On your question: grant exactly one route as chat-agnostic, `GET /sessions/:sessionId`. It reports the session's own state and account, carries no chat data and changes nothing, and an integration needs it to know whether its session is connected. Leave everything else unmarked: label, template and profile writes change state every other key sees, and the remaining reads are not needed by an integration working inside a few chats.

What still has to change:

1. **Bull Board admits a chat-restricted admin key.** `/api/admin/queues` authenticates in `BullBoardAuthMiddleware`, which checks the ADMIN role and `allowedSessions` only (`src/common/security/bull-board-auth.middleware.ts:83-93`). Its webhook jobs carry every chat's event payloads, and the board can retry or clean them. It is a third surface outside `ApiKeyGuard` next to `/events` and MCP, and the #1634 amendment missed it too. Refuse a key with `allowedChats` beside the `allowedSessions` check, with a spec case.
2. **Participant routes reach DMs on whatsapp-web.js.** `POST :groupId/participants` is marked and fenced by `:groupId`, but `wwebjs-groups.ts:187` calls `chat.addParticipants(ids)` with the library's default `autoSendInviteV4: true`, which sends a group-invite message into the 1:1 chat of anyone whose privacy setting refuses the add. Leave the participant add, remove, promote and demote routes unmarked in this PR.
3. **The coverage spec proves less than it says.** Since 601b7216 it reports a marked handler only when it carries a field from `UNFENCED_CHAT_FIELDS` and names no guard-visible chat; everything else passes as chat-agnostic. As the branch stands, replacing `return this.chatScope.filter(...)` in `session.controller.ts:524` with `return chats`, or marking `GET /sessions/:sessionId/messages` (which would then return every chat's messages to a restricted key) or `GET :channelId/messages`, leaves the spec green. Make the three categories explicit, for example `@ChatScoped('fenced' | 'filtered' | 'agnostic')`: a fenced handler needs a path param or a required guard-read body field (an optional `@Query('chatId')` does not count), a filtered one returns through `chatScope.filter`, and agnostic grants sit on an explicit list the spec checks. Pair each decorator with a parsed handler the way `query-session-scope-coverage.spec.ts:99` does.
4. **Enforcement no spec pins.** Each of these can be reverted with every spec green: the `:groupId` and `:contactId` fence (the guard spec exercises only `:chatId`, and 25 of the 63 marked handlers depend on the other two), the `GET .../chats` filter, a phone entry not admitting the same-digits `@lid`, `lidsForPhonePersisted` reading the table rather than the cache, and the `allowedChats` clause in `usableAdminCondition` (the spec runs a JavaScript mirror of the SQL, so the string itself never executes). One case each.
5. **`GET .../chats` filters after paginating** (`session.service.ts:741`, `session.controller.ts:524`), so a restricted key gets short or empty pages and never sees an allowed chat past the window. Filter before paginating.
6. **One table query per allowlist entry on every request.** `buildChatScope` awaits the persisted lookup for each entry in turn (`chat-scope.ts:56-59`), twice on `GET .../chats`. For the guard, expand the requested chat id once with `resolveJidCandidates` and test the candidates against the allowlist's literal forms; batch the lookups for the filter.
7. **`quotedMessageId` is a chat reference the guard does not read.** A restricted key can quote a message from a chat outside its allowlist into an allowed one; Baileys resolves the quote from the global store (`baileys-messaging.ts:749-751`). Refuse a present `quotedMessageId` for a restricted key, or require the quoted message to belong to `body.chatId`, as `POST .../messages/reply` already does.
8. **Smaller:** mark `send-audio`, whose DTO extends the fenced media DTO. `send-bulk` can stay marked, but its batch status and cancel routes must stay unmarked until they check every batch `chatId` against the scope (batch rows have no key owner), so its `statusUrl` answers 403 for a restricted key and the docs should say so.
9. **CHANGELOG:** one bullet instead of two, `send-bulk` rather than `messages/bulk`, MCP refusing per tool call rather than at a handshake, Bull Board once it refuses, and which list routes a restricted key can use (only `GET .../chats`, filtered).

Reading a chat on Baileys stays out of reach for a restricted key after this (`/history` answers 501 there and `GET .../messages` refuses); that is fine for this PR and can follow as its own change.

Checked on the branch merged with current `main`: tsc, lint, `format:check`, `test:docs`, the contract gates, `openapi:check`, the dashboard lane and `npm test -- --coverage` all pass.

bhavyachopra99:
All points addressed, merged with current `main`; `npm test -- --coverage`, `tsc`, `eslint`, `format:check`, `test:docs`, the contract gates, `openapi:check` and the dashboard lane pass.

**Agnostic grant** — exactly one, `GET /sessions/:sessionId`, and `@ChatScoped` now carries its reason: `'fenced'`, `'filtered'` or `'agnostic'`. The spec checks each category against its rule (`'fenced'` needs a path param or a *required* guard-read body field — an optional `?chatId=` fails; `'filtered'` must return through `ChatScopeService.filter`; `'agnostic'` must be on `AGNOSTIC_GRANTS`), pairs every decorator with a parsed handler, and fails if a grant is declared but unused or used but undeclared. Replacing the `getChats` filter or marking `GET .../messages` / `GET :channelId/messages` now fails it.

**Bull Board** — `/api/admin/queues` refuses a key with `allowedChats` beside the `allowedSessions` check, with a spec case.

**Participant routes** — `add/remove/promote/demoteParticipants` are unmarked, so they refuse. `send-audio` is now marked `'fenced'` (its DTO inherits the media `chatId`; the spec resolves `extends`).

**Cost** — the guard expands the *requested* chat id once (≤2 lookups) and tests it against the allowlist's literal forms, instead of expanding every entry per request. The list filter expands the allowlist once, batched (two queries) through new store methods.

**Filter before paginating** — `SessionService.listChats` returns the unwindowed list and the controller filters then paginates; the spec proves it with a disallowed chat sorted newest.

**`quotedMessageId`** — refused outright for a restricted key.

**Specs pinning what was revertible** — one case each: the `:groupId`/`:contactId` fence, the `.../chats` filter, a phone entry not admitting the same-digits `@lid`, `lidsForPhonePersisted` reading the table for an LRU-evicted lid, and the `allowedChats` clause in `usableAdminCondition` (asserted on the AND-fragment the guard binds, since the JS mirror never runs the SQL).

**Smaller** — batch status/cancel stay unmarked and their `403` is documented on the routes (OpenAPI regenerated); CHANGELOG collapsed to one bullet using `send-bulk`, MCP refusing per tool call, Bull Board, and the list-route note.

Reading a chat on Baileys stays out of reach for a restricted key after this (`/history` is `501` there, `GET .../messages` refuses) — left as its own change, as you noted.

rmyndharis (changes requested):
Thank you, this round closes almost everything. Checked on the branch merged with current `main`: Bull Board refuses a chat-restricted key, the three `@ChatScoped` categories are enforced by the coverage spec (marking `GET .../messages`, dropping the chats filter, or an agnostic mark off `AGNOSTIC_GRANTS` each turns it red), the chats list filters before paginating, the guard expands only the requested id, and the pinning specs fail when the enforcement they cover is reverted. I could not get a restricted key past the fence on any marked route, and tsc, lint, `format:check`, `test:docs`, the contract gates, `openapi:check`, the dashboard lane and `npm test -- --coverage` all pass.

What is left:

1. **Participant remove, promote and demote are still marked.** Only `addParticipants` lost its mark; `removeParticipants`, `promoteParticipants` and `demoteParticipants` became `@ChatScoped('fenced')` (`group.controller.ts:214`, `:240`, `:267`), so a restricted key can still remove or promote numbers it was never scoped to. Drop the three marks, as the reply says was done.
2. **The reply route can never succeed.** `POST .../messages/reply` is marked, but the guard refuses any non-empty `quotedMessageId` (`api-key.guard.ts:193-201`) and `ReplyMessageDto` requires one, so a restricted key always gets 403. Both engines already bind the reply's quote to `chatId` (Baileys `assertStoredInChat`, `baileys-messaging.ts:477`; whatsapp-web.js fetches the quote from the named chat, `wwebjs-messaging.ts:543-549`), so exempt the reply route from the refusal, or unmark it.
3. **A bulk body is looked up entry by entry before the 100-item cap.** `chatIdsIn` collects every `messages[].chatId` and the guard awaits one lookup per entry (`api-key.guard.ts:170`, `:182-188`) before the pipe applies `@ArrayMaxSize(100)`, so one oversized request drives thousands of sequential table queries. Answer 400 in the guard when `messages` holds more than 100 entries, before any lookup.
4. **The guard and the chats filter disagree on spelling.** The guard's candidates keep a lid or group id as typed (`jid-candidates.ts:27`, `:30`), while the filter normalizes through `addressForms`, so `777000111:5@lid`, `555000111@LID` or an upper-case `@G.US` id is refused on a send but admitted in the list. It fails closed; start the lid candidates from `${userPart}@lid` and add the normalized group form, with those inputs in the equivalence cases.
5. **Cache-first forward lookups.** `resolveLidPersisted` and `phonesForLidsPersisted` return a cached value, including a cached negative, before reading the table (`lid-mapping-store.service.ts:157-161`, `:186-197`), while the reverse lookup merges cache and table. On a node holding a stale negative for a lid that another node has since mapped, the chats list shows that chat and the guard refuses it. Read the table for the forward direction too, or drop the CHANGELOG claim that the answer never depends on the cache.
6. **Two pins.** A phone entry not admitting the same-digits `@lid` is covered on the guard path only; add the same case for `buildExpandedChatScope` plus `filterByChatScope`. And the batched `IN` lists are unbounded, so a very large allowlist exceeds SQLite's 32766-variable limit; chunk them.
7. **CHANGELOG:** keep it to what an operator needs in one or two sentences (the entry is about 260 words), without the em dashes, and end it with `Thanks @bhavyachopra99 and @lasithadilshan.`

bhavyachopra99:
All five fixed, plus a full end-to-end pass. Merged with current `main`; `npm test -- --coverage` (6672), `tsc`, `eslint`, `format:check`, `test:docs`, `openapi:check`, `check:contract-shapes`, the three SDK gates and the dashboard lane all pass.

**Participant routes** — `removeParticipants`, `promoteParticipants` and `demoteParticipants` were still marked (my earlier edit only landed on `addParticipants`). All four are now unmarked, so a restricted key can neither add, remove, promote nor demote.

**Reply** — exempted rather than unmarked, since both engines bind the quote to the named chat (Baileys `assertStoredInChat`, whatsapp-web.js fetches from that chat). A new `@ChatQuotedAllowed` marker exempts the route from the `quotedMessageId` refusal; the coverage spec pins the marker to exactly that one route, so it cannot spread to `send-text` and friends.

**Bulk** — the guard now answers `400` when `messages` exceeds 100 entries, before any lid lookup. The cap is a shared `BULK_MESSAGES_MAX` used by both the DTO's `@ArrayMaxSize` and the guard, so the two cannot drift.

**Spelling** — `resolveJidCandidates` now starts a lid from `<lid>@lid` and returns the normalized `<id>@g.us` for a group, so `777000111:5@lid`, `555000111@LID` and an upper-case `@G.US` no longer get refused on a send while being admitted in the list. Equivalence cases added.

**Forward lookups** — `resolveLidPersisted` reads the table first (the cache is only a fallback when the table has no row) and `phonesForLidsPersisted` lets the table win, so a stale cached negative on one node cannot shadow a mapping another node persisted. The batched `IN` lists are chunked at 500, well inside SQLite's 32768-variable limit.

**Pins** — the phone-vs-same-digits-`@lid` rule is now covered on the expanded filter scope as well as the guard path (`buildExpandedChatScope` + `filterByChatScope`).

**CHANGELOG** — one 91-word bullet, no em dashes, ending with the credit line.

End-to-end re-check on the branch: an unrestricted key is unaffected (the fence is skipped entirely); a restricted key is refused on every unmarked route, including webhooks, automation, status, channels, key management, Bull Board, `/events`, MCP (per tool call) and the bulk batch status/cancel `statusUrl`; on marked routes the guard expands only the requested id, caps bulk before lookups, and refuses a quote except on reply; `GET .../chats` filters before paginating.

rmyndharis:
Thank you for round four: every point from the last review is in. `main` moved a lot since (#1693, #1692) and conflicted with the lid lookups and the message routes, so I merged it into the branch and pushed the remaining gaps as separate commits:

- The merge keeps both sides: `main`'s batch-route order and encoded `statusUrl` with your 403 docs, the `@ChatScoped` mark on `:chatId/history`, and the message filter on `main`'s table-reading lid lookups with the raw input kept.
- `findPhoneForLid` reads the lid table first, so the fence and every filter behind it agree when another node has re-mapped a lid; `findLidsForPhone` drops a cached lid the table maps elsewhere. Your `resolveLidPersisted`/`lidsForPhonePersisted` fold into these two.
- A reply or button click reads the quoted body from the target chat only. The pending row was saved before the engine refused a cross-chat quote, so a chat-restricted key could file another chat's body under an allowed one. Quoting sends keep cross-chat quotes, since the guard already refuses those keys a `quotedMessageId`.
- WebSocket subscribe refuses a chat-restricted key, as the handshake does; Baileys edit checks the chat before ownership, so it no longer tells an inbound message from an own one in another chat; bulk checks each distinct recipient once.
- The lid store fake repo honours its `WHERE` clause (including `In()`), so the batched lookups are tested for real; the coverage spec pins the participant and batch routes unmarked.
- `allowedChats` is documented in docs/04, docs/05 and docs/06, including the new 403 cases.

All CI jobs pass locally, including the PostgreSQL job. 0.23.6 is being cut from the current `main`, so this change is not part of it.

bhavyachopra99:
Thanks for carrying this the rest of the way. The follow-up commits all look right to me (the WebSocket subscribe refusal, the cross-chat quote on the send routes, the table-first lid lookups, the bulk de-dup), and the merge keeps everything intact. I'll be waiting for it to merge, and I'm happy to pick up anything else on this or a follow-up.

It's been a pleasure working on OpenWA. The depth of the reviews is rare, and the project looks genuinely promising. Happy to help and contribute again.

bhavyachopra99:
Added a README section for this — a **Chat-scoped operator & viewer tokens** subsection beside the session-scoped one, plus rows in the Why OpenWA and Advanced feature tables. It covers the default-deny fence, the two scopes being independent, the agent-on-a-shared-account story (scoped to a few groups/DMs, cannot see personal chats), lid-aware identity matching, and the surfaces that refuse such a key (`/events`, MCP, Bull Board) with the one list route that filters. It also says plainly that this limits what a key can reach, not the ban risk, so it does not contradict the existing guidance above.

It's docs-only, so it should not touch the merge. Happy to undo or reshape it if you'd rather word it differently or keep the README as it was.

rmyndharis:
Thanks for the README section, and for the kind words.

Two commits are pushed on top of 24db9944.

d81d186b keeps group join requests and invite links closed to a chat-scoped key. The list, approve and reject membership-request routes and the invite-code read and revoke routes were marked fenced on `:groupId`. A join request comes from a number outside the key's scope: approving it admits that number, the same membership change the participant add, remove, promote and demote routes stay unmarked to refuse, and the list names who is asking. An invite link does the same without asking: anyone holding it can join, it works outside OpenWA and keeps working after the key is revoked (the comment at `group.controller.ts:533` says so), and revoke returns a fresh one. The five routes are unmarked now, so the default deny answers `403`, and they are pinned in `MUST_STAY_UNMARKED` next to the participant routes.

75ee1a62 matches the chat-scoped key section to the code:

- The dashboard does not set or show `allowedChats`. Create and edit send only `allowedSessions` (`dashboard/src/pages/ApiKeys.tsx:104`, `:132`), and the server writes `allowedChats` only when the field is sent (`src/modules/auth/auth.service.ts:246`). The section now names `POST /auth/api-keys` and `PUT /auth/api-keys/{id}` and says a dashboard edit leaves the chats unchanged.
- "Only read and send" understated an operator key. Marked operator routes also delete a chat (`src/modules/session/session.controller.ts:800`), clear its messages (`:667`), rename a group (`src/modules/group/group.controller.ts:381`), leave it (`:433`) and block the contact (`src/modules/contact/contact.controller.ts:231`). The section also says what stays closed: participants, join requests and the invite link.
- Reading a chat goes through the history route (`src/modules/message/message.controller.ts:528`), which is whatsapp-web.js only: on Baileys `getChatHistory` answers `501` (`src/engine/adapters/baileys.adapter.ts:567-574`), and the key sees only the chat list's last-message preview.
- The key receives no pushed events. `/events` refuses it at the handshake and on subscribe (`src/modules/events/events.gateway.ts:350`, `:514`), and the webhook routes are unmarked.
- The default deny lives in `ApiKeyGuard` (`src/modules/auth/guards/api-key.guard.ts:118-127`) and covers authenticated REST routes, so "a surface we have not thought of yet cannot leak" held only there. `/events`, the MCP mount (`src/core/agent-tools/tool-invoker.ts:61`) and Bull Board (`src/common/security/bull-board-auth.middleware.ts:99`) each refuse the key with their own check, and a session outside `allowedSessions` still answers `401` before the chat check.
- A phone entry matches its `@lid` form only once the lid table maps the two, and an unmapped `@lid` is refused.
- An admin key can carry `allowedChats`, but no marked route requires `ADMIN`, and the last usable admin cannot be scoped: the update check counts `allowedChats` (`src/modules/auth/auth.service.ts:226`) and the usable-admin condition requires it empty (`:342`).
- "An account you also use yourself" contradicted the advice to use a dedicated number, so the section now says a shared account. That ban-risk guidance sits below the section, not above.
- The Chat Scoping row in the Advanced table now says the same thing as the section.

This PR stays open until 0.23.6 is released. After that, main is merged in and it goes in.

PR #1675, opened 2026-09-20
fix(wwebjs): resolve history message sender through getContact()

TanmayChachra:
## Description

Split out of the closed client-mapping PR, as suggested in review. Independent of the directory feature.

\`getChatHistory\` only took the synchronous \`notifyName\` from the raw payload, which is often absent on a history-fetched message. A group participant outside the account's contacts therefore had no sender label in history, while the live \`message\` handler resolves one via \`getContact()\` for the same message. This mirrors that enrichment in history (same \`WEBHOOK_CONTACT_DETAILS\` handling via \`mapContactFields\`). A \`getContact()\` failure is logged and does not fail the history fetch.

## Type of Change

- [x] Bug fix

## Checklist

- [x] Tests added/updated (two specs in \`whatsapp-web-js.adapter.spec.ts\`: name resolved, and failure tolerated)
- [x] Documentation updated (CHANGELOG \`[Unreleased]\`)
- [x] Lint passes
- [x] Self-reviewed

Note: 3 "orphaned Chromium sweep" specs in that file fail on Windows on unmodified \`main\` as well; unrelated to this change.


rmyndharis:
I pushed two commits to this branch:

- `main` merged in, resolving the CHANGELOG conflict. The entry keeps your wording and ends with the credit line.
- 62ebcdd9: the sender lookup is cached per sender, keyed on `author || from` as `getContact()` keys it. Each call is a round trip into the page and a history page repeats the same few senders, so resolving per message cost one lookup per message, up to 2000 with `deep=true`; a failed lookup is cached as well. The failure spec now asserts that the lookup ran (it passed with the lookup removed), and a new spec covers the cache. The client type comments, the dashboard type, `.env.example` and the status seed comment said history carries `pushName` only or comes back nameless, which this change makes untrue, so they are updated too.

Thanks for splitting this out.

PR #1676, opened 2026-09-20
fix(dashboard): render @<digits> mentions as @FirstName

TanmayChachra:
## Description

Split out of the closed client-mapping PR, as suggested in review. Independent of the directory feature.

An @mention reaches the body as raw digits (\`@166868170059932\`). \`buildMentionNameMap\` maps each thread author's id (local part of \`author\`, \`:device\` stripped) to their resolved name, and \`resolveMentions\` substitutes \`@<digits>\` with \`@FirstName\` in message and quoted bodies. Frontend only, no schema or API change. An unmatched token, or one shorter than 7 digits, is left exactly as WhatsApp sent it.

## Type of Change

- [x] Bug fix

## Checklist

- [x] Tests added/updated (6 tests in \`chatMessages.test.ts\`)
- [x] Documentation updated (CHANGELOG \`[Unreleased]\`)
- [x] Lint passes
- [x] Self-reviewed


rmyndharis (changes requested):
Thank you for splitting this out. The two helpers are the right size, the name map is memoized on the message list, and the tests live in a `*.test.ts` file, so they actually run.

What has to change before this can merge:

1. CI fails: `npm run format:check` in `dashboard/` flags `src/components/chats/ChatThread.tsx` and `src/utils/chatMessages.test.ts`. `npm run format` fixes both.
2. The substituted name can become a link. `chatName` is the sender's own push name (`src/modules/session/message-projector.service.ts:250`), which any WhatsApp user sets freely, and `resolveMentions` splices it into the body before `MessageBody` runs `Linkify` over the text (`MessageBody.tsx:44-46`). A participant whose push name is `bit.ly/free` turns every mention of them, in anyone's message, into a clickable `https://bit.ly/free`. Render the resolved mention as its own element outside the linkified text, or reduce the name to characters that cannot form a link before inserting it.
3. The match needs a left boundary. `/@(\d{7,})/g` also fires inside `admin@12345678.com` when `12345678` is a participant id, and because the substitution runs before parsing it rewrites inside code spans and URLs as well. Only replace an `@` that starts the text or follows whitespace or punctuation.
4. A push name that is blank after trimming destroys the mention: `'   '` passes the truthiness check in `buildMentionNameMap`, and `'   '.split(' ')[0]` is `''`, so `@6281112345` renders as a bare `@`. Trim the name before storing it and skip an empty result.
5. On Baileys the lid case from the description does not resolve. The adapter normalizes `author` from `<lid>@lid` to `<phone>@c.us` whenever the mapping is known (`src/engine/adapters/baileys-message-mapper.ts:908-911`, `toNeutralJid` in `src/engine/identity/wa-id.ts`), so the map is keyed by phone digits while a body like `@166868170059932` carries lid digits. Either resolve both forms or state in the changelog entry which case is covered.
6. The `:device` test uses `628111@c.us:7`, which is not a JID shape; the device suffix sits before the `@`, as in `628:12@s.whatsapp.net` (`userPart` in `src/engine/identity/wa-id.ts:64`). As written, the test still passes with `.split(':')[0]` removed from the helper.

rmyndharis (changes requested):
Thank you for the quick turnaround. The format check, the left boundary for the email case, the blank-name check, the Baileys note in the CHANGELOG and the real `:device` fixture are all in, and the new tests fail when the lines they cover are reverted.

Two things still need a change:

1. **A push name can still become a link.** Stripping characters stops `bit.ly/free`, but linkifyjs treats the bare word `localhost` as a URL, so a participant named `localhost` turns every mention of them into `https://localhost`; and when the text continues right after the mention (`@6281112345.So`), the name becomes a domain label. Stripping cannot close this, because the name keeps passing through `Linkify`. Render the resolved mention as its own element that Linkify skips instead: for example wrap it in `<bdi>` (which also isolates a right-to-left name) and add `'bdi'` to `ignoreTags` in `MessageBody.tsx`. Checked with linkify-react 4.3.3: `hi <bdi>@localhost</bdi> there` renders with no anchor. With that in place, the character stripping can go.
2. **The left boundary still fires after `/` and a backtick**, and `resolveMentions` runs on the raw body before it is parsed, so a URL with an `@<digits>` path segment and inline code are rewritten. Restrict the boundary to the start of the text, whitespace or opening punctuation, for example `(?<=^|[\s(\[{"'])`, with one test each for a URL and a code span.

Smaller: a push name made only of the Hangul filler (U+3164) passes the blank check and still renders a bare `@`; treat it like whitespace. The CHANGELOG line should end with `Thanks @TanmayChachra.`

I merged current `main` into the branch, since it conflicted in the CHANGELOG; please pull before pushing again.

rmyndharis:
Thank you, this round closes both items: a push name can no longer become a link (checked again against linkify-react 4.3.3 for `localhost`, `bit.ly/free`, a `.So` continuation and a URL-shaped name, no anchor in any of them), and the boundary no longer fires after `/` or a backtick.

I pushed two commits on top of your branch rather than opening a fourth round:

- A mention inside `*bold*`, `_italic_`, `~strike~` or a fenced block lost its formatting: the span was split out before the format parser ran, so `*Reminder @6281112345 at 10*` rendered with literal asterisks once that participant had posted, and `` `echo @6281112345` `` lost its code span. The split now happens at the text leaf after code and format parsing, and the format scanner steps over a span whole, so a `*` in a push name cannot open a span either. `resolveMentions` skips code segments, and a format opener counts as a left boundary, so `*@digits*` becomes a bold mention.
- A push name carrying the closing delimiter could end its own span early and hand the rest of the name back to Linkify; the delimiters are now stripped before the name is stored. The blank-name check also covers the other invisible code points (zero-width space, soft hyphen, word joiner, bidi marks, combining marks, the other Hangul fillers, the blank braille pattern), not only U+3164.
- The newest push name in the thread now wins, so a participant who renamed themselves resolves to their current name.
- The changelog entry had landed under `[0.23.6]` when the branch merged the release commit; it is back under `[Unreleased]`.

Tests for each case are in `messageFormatter.test.ts`, `chatMessages.test.ts` and `MessageBody.test.ts`. Please pull before pushing again.

